PR #28, opened 2021-01-15
[WIP] Build once

tgraf:


tgraf:
This idea doesn't really work because `workflow_run` only functions on default branches

PR #41, opened 2021-01-22
[RFC] status: Bail out early on image pull failures

christarazi:
See commit msgs.

I marked this PR as RFC because I don't think this quite done. For one, the
status printed out at the end shows the Operator, Hubble, ClusterMesh are OK,
which is not the case. Another area to look into is how sensitive this
detection and bail out is. We don't want to bail out too early, for example, if
an image pull transiently fails.

I submitted the minimally working version of this to gather feedback first on
if we think this is even worth optimizing. If so, then we can iterate on the
 outlined areas above.

### DIFF
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -16,6 +16,7 @@ package status
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


christarazi:
Closed because GitHub was not updating to my new commits. Probably a bug from when this repo was moved from `isovalent` org to `cilium`.

PR #64, opened 2021-01-27
add make targets to build release binaries for Linux and MacOS

rolinh:
This allows building release binaries for the following platforms:

  - darwin/amd64
  - linux/386
  - linux/amd64
  - linux/arm
  - linux/arm64

Unfortunately, Windows is not supported yet due to syslog usage by the cilium logger.

This PR is based on how it's done for the Hubble CLI.

NOTE: ~This PR depends on #62~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ GIT_BRANCH = $(shell which git >/dev/null 2>&1 && git rev-parse --abbrev-ref HEA
 GIT_HASH = $(shell which git >/dev/null 2>&1 && git rev-parse --short HEAD)
 
 TEST_TIMEOUT ?= 5s
+RELEASE_UID ?= $(shell id -u)
+RELEASE_GID ?= $(shell id -g)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,12 +20,47 @@ $(TARGET):
...
+		--volume `pwd`:/cilium docker.io/library/golang:1.15.7-alpine3.13 \
+		sh -c "apk add --no-cache make && make local-release"
+
+local-release: clean
+	for OS in darwin linux; do \
+		EXT=; \
+		ARCHS=; \
+		case $$OS in \
+			darwin) \
+				ARCHS='amd64'; \
```
kaworu, line 38: Might be outside the scope of this PR, but should we start `arm64` OSX builds?

rolinh: Yes. However, as for Hubble CLI, this is pending on Go 1.16 to be released.

kaworu (approved):

PR #65, opened 2021-01-28
hubble/clustermesh: Add --create-ca flag to create CA if needeed

tgraf:
The CA is typically created at install time. Allow creating it on the
fly if needed. We don't do this automatically as the common case will be
that the CA already exists as sharing between clusters is desirable. The
default behavior should thus be to fail if it does not exist.

### DIFF
--- a/clustermesh/certs.go
+++ b/clustermesh/certs.go
@@ -154,10 +154,22 @@ func (k *K8sClusterMesh) deleteCertificates(ctx context.Context) error {
 func (k *K8sClusterMesh) installCertificates(ctx context.Context) error {
 	err := k.certManager.LoadCAFromK8s(ctx)
 	if err != nil {
-		k.Log("❌ Cilium CA not found: %s", err)
-		return err
+		if !k.params.CreateCA {
+			k.Log("❌ Cilium CA not found: %s", err)
+			return err
+		}
+
+		k.Log("🔑 Generating CA...")
+		if err := k.certManager.GenerateCA(); err != nil {
+			return fmt.Errorf("unable to generate CA: %w", err)
+		}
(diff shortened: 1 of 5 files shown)


clustermesh/certs.go
```diff
@@ -154,10 +154,22 @@ func (k *K8sClusterMesh) deleteCertificates(ctx context.Context) error {
...
-		k.Log("❌ Cilium CA not found: %s", err)
-		return err
+		if !k.params.CreateCA {
+			k.Log("❌ Cilium CA not found: %s", err)
+			return err
+		}
+
+		k.Log("🔑 Generating CA...")
+		if err := k.certManager.GenerateCA(); err != nil {
+			return fmt.Errorf("unable to generate ca: %w", err)
```
tklauser, line 164: Spelling nit (`CA` is used in other log messages here):

```suggestion
			return fmt.Errorf("unable to generate CA: %w", err)
```

hubble/hubble.go
```diff
@@ -135,9 +136,23 @@ func (k *K8sHubble) Disable(ctx context.Context) error {
...
+	err := k.certManager.LoadCAFromK8s(ctx)
+	if err != nil {
+		if !k.params.CreateCA {
+			k.Log("❌ Cilium CA not found: %s", err)
+			return err
+		}
+
+		k.Log("🔑 Generating CA...")
+		if err := k.certManager.GenerateCA(); err != nil {
+			return fmt.Errorf("unable to generate ca: %w", err)
```
tklauser, line 148: Same nit here:
```suggestion
			return fmt.Errorf("unable to generate CA: %w", err)
```

PR #71, opened 2021-02-01
install: fix cluster name autodetection with dots

tklauser:
Previously, dots were disallowed in cluster names per commit
46db7b821ad9 ("install: Reject cluster names with dots"). This was due
to the fact that the Hubble server certificate would otherwise include
the dots which would lead to an unexpected commonName in the
certificate, making Relay unable to authenticate the Hubble servers.
This was fixed in commit 942d4325208c ("hubble: support clusters with
dots in their name").

Update the cluster name autodetection to allow cluster names with dots.
This e.g. allows to seamlessly use `cilium install` (i.e. without
setting `--cluster-name`) on clusters created by `eksctl` which by
default uses the naming scheme `<cluster-name>.<region>.eksctl.io`.

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -42,7 +42,7 @@ var (
 		},
 	}
 
-	clusterNameValidation = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])$`)
+	clusterNameValidation = regexp.MustCompile(`^[a-z0-9]([-.a-z0-9]*[a-z0-9])$`)
 )
 
 func (p InstallParameters) checkDisabled(name string) bool {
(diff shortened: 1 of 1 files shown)


kaworu (changes requested):
I think that the clustermesh certs currently lack support for dots in clustername. That is why I haven't removed the check in #15, sorry for not mentioning it.

This is something we've discussed with @aanm at some point but I don't know the status yet.

https://github.com/cilium/cilium-cli/blob/32b611bd2619ae4c111bafceef33e8d17bda023e/clustermesh/certs.go#L34-L40
https://github.com/cilium/cilium-cli/blob/32b611bd2619ae4c111bafceef33e8d17bda023e/clustermesh/clustermesh.go#L602

tklauser:
> I think that the clustermesh certs currently lack support for dots in clustername. That is why I haven't removed the check in #15, sorry for not mentioning it.
> 
> This is something we've discussed with @aanm at some point but I don't know the status yet.
> 
> https://github.com/cilium/cilium-cli/blob/32b611bd2619ae4c111bafceef33e8d17bda023e/clustermesh/certs.go#L34-L40
> 
> https://github.com/cilium/cilium-cli/blob/32b611bd2619ae4c111bafceef33e8d17bda023e/clustermesh/clustermesh.go#L602

I see, thanks for the explanation. In that case this PR doesn't make sense (yet), closing.

PR #75, opened 2021-02-04
internal/k8s: fix cluster name autodetection for eksctl created clusters

tklauser:
Clusters created using `eksctl create cluster--name foo` will have their
cluster name set to `foo.<region>.eksctl.io`. This will break cluster
name autodetection due to the fact that currently, dots are not allowed
in cluster names (see cilium/cilium-cli#71). Thus, replace dots by
hyphens for the purpose of the Cilium CLI.

This will allow to seamlessly use `cilium install` (i.e. without setting
`--cluster-name` explicitly) on `eksctl`-created clusters.

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -327,7 +327,7 @@ func (c *Client) AutodetectFlavor(ctx context.Context) (f Flavor, err error) {
 	}
 
 	// When creating a cluster with kind create cluster --name foo,
-	// the context and clustername are kind-foo.
+	// the context and cluster name are kind-foo.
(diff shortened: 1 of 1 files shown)


kaworu:
Have you considered replacing dots with hypens (`-`) instead of stripping? That is how we workaround this issue for the Hubble TLS stuff, has the benefit to preserve information and can be generalized.

tklauser:
> Have you considered replacing dots with hypens (`-`) instead of stripping? That is how we workaround this issue for the Hubble TLS stuff, has the benefit to preserve information and can be generalized.

Good point, thanks for the suggestion. I've implemented this in the latest revision.

kaworu (approved):
Thanks!

PR #89, opened 2021-02-19
ci: Run all workflows in intervals every 6 hours

tgraf:


pchaigno:
@tgraf Do you remember what was the purpose of this? These workflows already run for every single PR and push to master. Running them also every 6h seems a bit overkill, especially with our limit of 20 concurrent jobs per organization.

PR #90, opened 2021-02-19
multicluster: Split test into more steps for simplified troubleshooting

tgraf:

### DIFF
--- a/.github/workflows/multicluster.yaml
+++ b/.github/workflows/multicluster.yaml
@@ -5,10 +5,11 @@ on:
   push:
     branches:
       - master
+
 env:
-  KIND_VERSION: v0.9.0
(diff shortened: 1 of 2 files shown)


.github/workflows/multicluster.yaml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request: {}
   push:
     branches:
-      - master
+      - pr/tgraf/fix-clustermesh
```
rolinh, line 7: I guess this change needs to be reverted before merging.

rolinh (approved):

.github/workflows/multicluster.yaml
tgraf: Yes, instead of disabling GitHub actions, we'll need to change the branch trigger in the future so we can test re-enabling in a particular PR.

PR #96, opened 2021-02-23
cmd: add cli flag to specify flavor manually instead of using auto detection

fristonio:
See the commit message for more details.


internal/cli/cmd/clustermesh.go
```diff
@@ -63,6 +63,8 @@ func newCmdClusterMeshEnable() *cobra.Command {
 	cmd.Flags().StringVar(&params.ApiserverImage, "apiserver-image", "", "Container image for clustermesh-apiserver")
 	cmd.Flags().BoolVar(&params.CreateCA, "create-ca", false, "Automatically create CA if needed")
 	cmd.Flags().StringVar(&contextName, "context", "", "Kubernetes configuration context")
+	cmd.Flags().StringVar(&params.K8sFlavor, "k8s-flavor", "",
+		"K8s flavor of the cluster to install Cilium in, this will override autodetect { kind | minikube | gke | aks | eks }")
```
joestringer, line 67: I'll probably get around to a microk8s flavor at some point, so would be nice to generate these list of flavours programatically from a static list under `internal/k8s` rather than copy+pasting them everywhere.

fristonio: Yeah, that makes sense. I will update the pull request.

PR #98, opened 2021-02-25
disable hubble on uninstall

### LINKED ISSUE #97: ServiceAccount hubble-relay left in-place after `cilium uninstall`
Steps to reproduce:

```
$ cilium install --cluster-name $CLUSTER_NAME
$ cilium hubble enable
$ cilium uninstall
```

After not using the cluster for a while, I attempted to re-install `cilium` using `helm`:

```
$ helm repo add cilium https://helm.cilium.io
$ helm repo update
$ helm install -n kube-system --set hubble.relay.enabled=true cilium cilium/cilium
```

This lead to the following error:

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ServiceAccount "hubble-relay" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "cilium-enterprise"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"
```

The `ServiceAccount` is indeed still there:

```
$ kubectl -n kube-system get serviceaccount hubble-relay
NAME                 SECRETS   AGE
hubble-relay         1         23m
```

Naturally, I would have expected that `ServiceAccount` to be uninstalled as well by `cilium uninstall`. Note that I can still get rid of it using `cilium hubble disable` before or after `cilium uninstall`. It might be more user-friendly though to remove as part of `cilium uninstall` as well, given all other `hubble-relay` objects are uninstalled.

kaworu:
Fix #97

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/cilium/cilium-cli/hubble"
 	"github.com/cilium/cilium-cli/install"
 
 	"github.com/spf13/cobra"
@@ -89,6 +90,13 @@ func newCmdUninstall() *cobra.Command {
 		Short: "Uninstall Cilium",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			h := hubble.NewK8sHubble(k8sClient, hubble.Parameters{
+				Namespace: params.Namespace,
+				Writer:    params.Writer,
+			})
+			if err := h.Disable(context.Background()); err != nil {
+				fatalf("Unable to disable Hubble:  %s", err)
+			}
 			uninstaller := install.NewK8sUninstaller(k8sClient, params)
 			if err := uninstaller.Uninstall(context.Background()); err != nil {
 				fatalf("Unable to uninstall Cilium:  %s", err)
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -130,9 +130,7 @@ func (k *K8sHubble) Validate(ctx context.Context) error {
 }
 
 func (k *K8sHubble) Disable(ctx context.Context) error {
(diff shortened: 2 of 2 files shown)


internal/cli/cmd/install.go
```diff
@@ -89,6 +90,13 @@ func newCmdUninstall() *cobra.Command {
 		Short: "Uninstall Cilium",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			h := hubble.NewK8sHubble(k8sClient, hubble.Parameters{
+				Namespace: params.Namespace,
+				Writer:    params.Writer,
+			})
+			if err := h.Disable(context.Background()); err != nil {
+				fatalf("Unable to disable Hubble:  %s", err)
```
tklauser, line 98: Not sure whether we should fatal here, e.g. in case Hubble wasn't enabled previously. We should still make sure to uninstall Cilium in case we fail to disable Hubble.

kaworu: Make sense, although arguably `hubble.Disable()` should not return an error when hubble is not enabled. I expect an error to be returned when something fatal happen (e.g. cannot connect to the cluster) that would prevent cilium to be uninstalled as well.

Note that, in the current implementation, there is no path were `h.Disable()` can return a non-nil error.

tklauser: Fair enough. I was slightly confused by the second commit in this PR which makes `(*K8sHubble).Disable` return the error from `(*K8sHubble).disableRelay` but I didn't realize that the latter currently also always returns `nil`. Let's leave it as is then.

tklauser (approved):

tgraf (approved):

PR #105, opened 2021-03-01
install: Add support for microk8s clusters

joestringer:
Add the conditional parameters necessary to install into a microk8s cluster.

This doesn't do any changes in the Hubble side for now, but it at least gets Cilium installed & running in the cluster.


joestringer:
Closing since CI jobs won't pass on PRs submitted from external forks.

PR #112, opened 2021-03-03
connectivity: Some package refactoring

errordeveloper:
- add check & tests subpackages
- use shorter type names for each of the tests


errordeveloper:
@tgraf PTAL! I'd like to improve check package next and add some tests using fake clientset and ensure that API interaction patterns are a bit more clear, we need to be able to do _create-or-update_ for each thing rather then having conditions around namespace existence.

tgraf (approved):

tgraf:
@errordeveloper I like it. Good refactoring. So far, there is no create-or-update assumption anywhere. We can do so but that's not what any of the code does right now.

errordeveloper:
@tgraf yeah, I understood it is create-or-skip right now, that has a few challenges... E.g. if namespace exists, nothing gets created. I have an idea of how to do this, will leave that for another PR.

tgraf:
>  I have an idea of how to do this, will leave that for another PR.
We should discuss that first because it has wide implications.

tgraf:
@errordeveloper Are you planning to rebase and submit? I have a couple of new PRs coming which change all tests with regard to flow validation and it would require an even bigger rebase for you.

errordeveloper:
@tgraf thanks for the heads-up, I've rebased this just now!

errordeveloper:
@tgraf assuming that the tests that are failing are basically not working on forks yet, this is ready to merge.

PR #121, opened 2021-03-12
version: report the cilium images version

kaworu:
Before this patch:
```
cilium-cli: v0.5-dev@pr/kaworu/version-report-cilium-version-08338b2 compiled with go1.16.2 on linux/amd64
```
After:
```
cilium-cli: v0.5-dev@pr/kaworu/version-report-cilium-version-08338b2 (v1.9.4 images) compiled with go1.16.2 on linux/amd64
```

### DIFF
--- a/internal/cli/cmd/version.go
+++ b/internal/cli/cmd/version.go
@@ -16,7 +16,10 @@ package cmd
 
 import (
 	"fmt"
+	"io"
+	"net/http"
 	"runtime"
+	"strings"
 
 	"github.com/cilium/cilium-cli/defaults"
 
@@ -32,6 +35,21 @@ var (
 	// GitHash is the git checksum of the most recent commit in HEAD.
 	GitHash string
 )
+
+func getLatestStableVersion() string {
+	resp, err := http.Get("https://raw.githubusercontent.com/cilium/cilium/master/stable.txt")
+	if err != nil {
+		return "unknown"
+	}
+	defer resp.Body.Close()
+
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return "unknown"
+	}
+
+	return strings.TrimSpace(string(b))
+}
 
 func newCmdVersion() *cobra.Command {
 	return &cobra.Command{
@@ -50,6 +68,7 @@ func newCmdVersion() *cobra.Command {
 			// the cluster, if any. See https://github.com/cilium/cilium-cli/issues/131
 			fmt.Printf("cilium-cli: v%s%s compiled with %v on %v/%v\n", Version, gitInfo, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 			fmt.Printf("cilium image (default): %s\n", defaults.Version)
+			fmt.Printf("cilium image (stable): %s\n", getLatestStableVersion())
 		},
 	}
 }
--- a/internal/cli/cmd/version.go
+++ b/internal/cli/cmd/version.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"runtime"
 
+	"github.com/cilium/cilium-cli/defaults"
+
 	"github.com/spf13/cobra"
 )
 
@@ -44,8 +46,10 @@ func newCmdVersion() *cobra.Command {
 			case GitHash != "":
 				gitInfo = fmt.Sprintf("@%s", GitHash)
 			}
-			// TODO: add support for reporting the Cilium version
+			// TODO: add support for reporting the Cilium version running in
+			// the cluster, if any. See https://github.com/cilium/cilium-cli/issues/131
 			fmt.Printf("cilium-cli: v%s%s compiled with %v on %v/%v\n", Version, gitInfo, runtime.Version(), runtime.GOOS, runtime.GOARCH)
+			fmt.Printf("cilium image (default): %s\n", defaults.Version)
 		},
 	}
 }


rolinh:
@kaworu I think the idea here was not to report the default cilium version but the version of Cilium that is used in the cluster (if any).

kaworu:
@rolinh I see, that make sense! While working on https://github.com/cilium/cilium-cli/pull/120 I was interested to know which version would be installed by default. Then, I came across the comment in `version.go` which I mistakenly took for the default Cilium version.

@tgraf Is there any interest in having the default version displayed somewhere in the version output? If not, feel free to close this PR.

tgraf:
> @tgraf Is there any interest in having the default version displayed somewhere in the version output? If not, feel free to close this PR.

Yes, although I was actually planning to fetch the latest version from GitHub directly so we would always automatically install the latest version. I'm planning to do that after the operator version is in.

kaworu:
I see! So what do you think about a `defaults.Version()` func? It could just return the constant for now, and allow to implement the "fetch from GitHub" logic in the future.

rolinh:
> I see! So what do you think about a `defaults.Version()` func? It could just return the constant for now, and allow to implement the "fetch from GitHub" logic in the future.

This makes sense to me as this would allow users to know which Cilium version would be installed by the tool. The Cilium version of a running cluster is already given by `cilium status`. I think that the fetch from github logic should be easy given that we have a `VERSION` file in the Cilium repository.

kaworu:
> I think that the fetch from github logic should be easy given that we have a VERSION file in the Cilium repository.

What I had in mind was using the GitHub API to get the latest release because our VERSION file has `1.10.90`, e.g.
```
curl \
  -H "Accept: application/vnd.github.v3+json" \
  https://api.github.com/repos/cilium/cilium/releases | jq -r '.[0].tag_name'
```
We could do some version compare with the releases to ensure we get the "highest release" rather than the latest (in the curl example we take the latest, but that could well be a v1.9.x release I believe).

rolinh:
> We could do some version compare with the releases to ensure we get the "highest release" rather than the latest (in the curl example we take the latest, but that could well be a v1.9.x release I believe).

In the Hubble repo, we have [`stable.txt`](https://github.com/cilium/hubble/blob/master/stable.txt) for this purpose actually. It always contains the latest stable released version. Introducing the same concept to the Cilium repository would make sense IMHO.

kaworu:
Closing to re-open to trigger the PR assignment bot.

kaworu:
"latest stable Cilium version" detection discussion moved at https://github.com/cilium/cilium-cli/issues/272

internal/cli/cmd/version.go
```diff
@@ -44,8 +46,9 @@ func newCmdVersion() *cobra.Command {
 			case GitHash != "":
 				gitInfo = fmt.Sprintf("@%s", GitHash)
 			}
-			// TODO: add support for reporting the Cilium version
-			fmt.Printf("cilium-cli: v%s%s compiled with %v on %v/%v\n", Version, gitInfo, runtime.Version(), runtime.GOOS, runtime.GOARCH)
+			// TODO: add support for reporting the Cilium version running in
+			// the cluster, if any.
```
tklauser, line 50: Might want to add a reference to #131 here?

defaults/defaults.go
```diff
@@ -117,3 +117,8 @@ var (
 	// CiliumPodSelector is the pod selector to be used for the Cilium agents.
 	CiliumPodSelector = "k8s-app=cilium"
 )
+
+// CiliumVersion return the default Cilium version used by the cli.
```
tklauser, line 121: Nit:

```suggestion
// CiliumVersion returns the default Cilium version used by the cli.
```

internal/cli/cmd/version.go
```diff
@@ -44,8 +46,9 @@ func newCmdVersion() *cobra.Command {
...
+			fmt.Printf("cilium-cli: v%s%s (%s images) compiled with %v on %v/%v\n", Version, gitInfo, defaults.CiliumVersion(), runtime.Version(), runtime.GOOS, runtime.GOARCH)
```
tklauser, line 51: Given that we also want to report the Cilium version running in the cluster (if any) in the future and to clearly separate the different versions, I'd probably add a separate line reporting this, e.g.

```suggestion
			fmt.Printf("cilium-cli: v%s%s compiled with %v on %v/%v\n", Version, gitInfo, runtime.Version(), runtime.GOOS, runtime.GOARCH)
			fmt.Printf("images: %s\n", defaults.CiliumVersion())
```

And then for the Cilium version running in the cluster, we'd probably add another line. This would also approximately correspond to what the in-cluster `cilium` CLI reports and it would keep the version lines somewhat consistent across tools.

tklauser:
Rebased and amended the review comments.

tklauser (approved):

rolinh (approved):
Awesome, thanks for adding the logic to fetch the latest stable version as well @tklauser :rocket:

tklauser:
After out-of-band discussion with @rolinh we've decided that it probably too risky using the latest stable version as provided in cilium/cilium `stable.txt` by default. For example, we've had some problems with the update from v1.9.8 to v1.10.1 (see https://github.com/cilium/cilium-cli/pull/358#issuecomment-866745869)

Thus, we've decided to keep the built-in default version (`defaults.Version`) for now and additionaly report the latest stable version in `cilium version`:

```
$ cilium version
cilium-cli: v0.8.5-dev@pr/kaworu/version-report-cilium-version-5c7fca26a2b7 compiled with go1.16.5 on darwin/amd64
cilium image (default): v1.10.2
cilium image (stable): v1.10.2
```

tklauser:
Connectivity tests on AKS failed with #367: https://github.com/cilium/cilium-cli/runs/3017925195?check_suite_focus=true

PR #134, opened 2021-03-22
install: add support for --install-no-conntrack-iptables-rules option

jibi:
This commit adds the same logic present in the configmap helm template
to disable the `--install-no-conntrack-iptables-rules` option (set to
true by default) in case Cilium is going to be installed in a managed
Kubernetes environment (such as AKS, EKS or GKE) or with a chained CNI
setup.

Signed-off-by: Gilberto Bertin <gilberto@isovalent.com>

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -1084,7 +1084,7 @@ func (k *K8sInstaller) Log(format string, a ...interface{}) {
 	fmt.Fprintf(k.params.Writer, format+"\n", a...)
 }
 
-func (k *K8sInstaller) generateConfigMap() *corev1.ConfigMap {
+func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 	m := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: defaults.ConfigMapName,
@@ -1182,8 +1182,9 @@ func (k *K8sInstaller) generateConfigMap() *corev1.ConfigMap {
 			"masquerade":            "true",
 			"enable-bpf-masquerade": "true",
 
-			"enable-xt-socket-fallback": "true",
-			"install-iptables-rules":    "true",
+			"enable-xt-socket-fallback":           "true",
+			"install-iptables-rules":              "true",
+			"install-no-conntrack-iptables-rules": "false",
 
 			"auto-direct-node-routes":             "false",
 			"enable-bandwidth-manager":            "true",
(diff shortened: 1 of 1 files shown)


pchaigno (approved):
Do we need to wait for https://github.com/cilium/cilium/pull/15264 to be part of a release before we can merge this?

jibi:
> Do we need to wait for [cilium/cilium#15264](https://github.com/cilium/cilium/pull/15264) to be part of a release before we can merge this?

I don't think we need to. I tested these changes with the default v1.9.5 and the installation works fine (this is just adding an extra item to the `cilium-config` cm, for the current release that field has no meaning and so it's just ignored)

install/install.go
```diff
@@ -1276,6 +1280,10 @@ func (k *K8sInstaller) generateConfigMap() *corev1.ConfigMap {
 		m.Data[key] = value
 	}
 
+	if _, ok := m.Data["cni-chaining-mode"]; ok {
+		m.Data["install-no-conntrack-iptables-rules"] = "false"
```
tgraf, line 1284: Does this need to be extended to also cover `kube-proxy-replacement: disabled`?

tgraf (changes requested):

install/install.go
jibi: No, the kube-proxy-replacement check is performed by the agent during initialization to make it work also with `--kube-proxy-replacement=probe` in case all KPR features end up being enabled

tgraf: agent init is horrible from a user experience. It means that `install` will block forever and the user has to check logs. Is there any situation where `install-no-conntrack-iptables-rules` can succeed if `kube-proxy-replacement: disabled`? I think not so we are better off preventing this.

We are also moving away from `kube-proxy-replacement: probe` (see https://github.com/cilium/cilium-cli/pull/136) because it will just suddenly break for users and is unpredictable. Users can't really know whether they still have to run kube-proxy or not.

jibi: > It means that install will block forever and the user has to check logs. 

I'm not sure I'm following. Install should not block in case the necessary KPR setup is not available as the agent will just disable the option: https://github.com/cilium/cilium/blob/b64afe42e9868970e69eeebb9b810bdf6bfe0480/daemon/cmd/kube_proxy_replacement.go#L337-L346

> Is there any situation where install-no-conntrack-iptables-rules can succeed if kube-proxy-replacement: disabled? I think not so we are better off preventing this.

Correct, install-no-conntrack-iptables-rules cannot be enabled with kube-proxy-replacement: disabled

We can add some additional logic here in the CLI and in the helm templates to prevent install-no-conntrack-iptables-rules from being enabled in case kube-proxy-replacement is set to disabled or is not set to strict, but given the previous point this feels redundant to me (unless I'm missing something)

tgraf: The agent automatically disabling something that was requested by the user is very intransparent. How does a user know whether the option was effective? We should move away from it. In general, the CLI should complain if a configuration cannot be fulfilled. There might be exceptions to this but the general rule should be to not hide decisions on what to enable/disable into log files.

jibi: I updated the PR:
* the default value for `--install-no-conntrack-iptables-rules` is now `false`
* `generateConfigMap()` will return an error in case any of the required conditions for enabling the option is not satisfied

tgraf (approved):

PR #136, opened 2021-03-23
install: Disable kube-proxy-replacement by default

tgraf:
Related: https://github.com/cilium/cilium/pull/15422

### DIFF
--- a/.github/in-cluster-test-scripts/external-workloads-install.sh
+++ b/.github/in-cluster-test-scripts/external-workloads-install.sh
@@ -3,7 +3,12 @@
 set -x
 set -e
 
-cilium install --cluster-name "${CLUSTER_NAME}" --restart-unmanaged-pods=false --config monitor-aggregation=none --config tunnel=vxlan --native-routing-cidr="${CLUSTER_CIDR}"
+cilium install --cluster-name "${CLUSTER_NAME}"\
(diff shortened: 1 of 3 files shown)


jrajahalme (approved):

pchaigno (approved):

pchaigno:
I rebased the PR as it was quite old and closed/reopen to trigger the tests (I think something is broken in the workflow triggers, I'll have a look). Let's see if the e2e tests still pass.

twpayne (approved):

pchaigno:
Tests are currently failing because KPR is disabled but the bandwidth manager is enabled. Thus, devices are autodetected and we end up in https://github.com/cilium/cilium/issues/12205's case for the native routing setups. I'll disable the bandwidth manager.

michi-covalent (approved):

pchaigno:
The only test failing, External Workloads, is failing with a different error than on master. That's because this PR is editing `external-workloads-install.sh` but the workflow runs the code from master. I've opened https://github.com/cilium/cilium-cli/pull/291 to run the PR version of the workflow and show that this workflow fails with the same error as on master. That error is tracked at https://github.com/cilium/cilium-cli/issues/224.

Only waiting for a review from @nebril now :rocket:

nebril (approved):

PR #137, opened 2021-03-26
connectivity: Add filter for ICMPv6, match ICMPv6 in pod-to-host test

eyanulis:
The pod-to-host test fails in IPv6 (or v6-first-dualstack) clusters when Hubble is enabled, because the test checks for ICMP(v4) codes 8 and 0 (echo request, echo reply) but the logged flows have ICMPv6 codes 128 (echo) and 129 (reply).

Add a filter for ICMPv6 mimicking the existing ICMPv4 filter, and use `filters.Or()` to make the pod-to-host test IP-version-agnostic.

### DIFF
--- a/connectivity/filters/filters.go
+++ b/connectivity/filters/filters.go
@@ -136,6 +136,37 @@ func ICMP(typ uint32) FlowFilterImplementation {
 	return &icmpFilter{typ: typ}
 }
 
+type icmpv6Filter struct {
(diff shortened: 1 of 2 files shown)


tklauser (approved):

tgraf (approved):

tgraf:
Retriggered the CI

PR #139, opened 2021-03-26
CI: add coverage for external workloads on GKE/GCP

tklauser:
This uses a GKE cluster with a GCP VM instance as an external workload.
For now it verifies basic connectivity according to the external
workloads GSG, see
https://docs.cilium.io/en/latest/gettingstarted/external-workloads


jrajahalme (approved):
Awesome, thanks!

PR #151, opened 2021-04-02
Add a restart flag to config set

aditighag:
After setting config map changes, Cilium pods need to be restarted for the changes to take effect.
Currently, this is a manual step. Add a new flag with default set to `true`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,6 +41,7 @@ type K8sConfig struct {
 
 type Parameters struct {
 	Namespace string
+	Restart   bool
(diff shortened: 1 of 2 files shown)


pchaigno:
@aditighag Do you remember why you closed this? It seems it would have helped with https://github.com/cilium/cilium-cli/issues/165 (according to https://github.com/cilium/cilium-cli/pull/176).

aditighag:
@pchaigno The changes are already merged - https://github.com/cilium/cilium-cli/pull/154.

PR #154, opened 2021-04-02
Add restart flag to config set and delete commands

aditighag:
After setting config map changes, Cilium pods need to be restarted for the changes to take effect.
Currently, this is a manual step. Add a new flag to restart the pods by default. 
Tested the changes on a local kind cluster.

### DIFF
--- a/internal/cli/cmd/config.go
+++ b/internal/cli/cmd/config.go
@@ -77,7 +77,7 @@ func newCmdConfigSet() *cobra.Command {
 		Args:  cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			check := config.NewK8sConfig(k8sClient, params)
-			if err := check.Set(context.Background(), args[0], args[1]); err != nil {
+			if err := check.Set(context.Background(), args[0], args[1], params); err != nil {
 				fatalf("Unable to set config:  %s", err)
 			}
 			return nil
@@ -85,7 +85,7 @@ func newCmdConfigSet() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
-
+	cmd.Flags().BoolVarP(&params.Restart, "restart", "r", true, "Restart Cilium pods")
 	return cmd
 }
 
@@ -101,7 +101,7 @@ func newCmdConfigDelete() *cobra.Command {
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			check := config.NewK8sConfig(k8sClient, params)
-			if err := check.Delete(context.Background(), args[0]); err != nil {
+			if err := check.Delete(context.Background(), args[0], params); err != nil {
 				fatalf("Unable to delete config:  %s", err)
 			}
 			return nil
@@ -109,6 +109,7 @@ func newCmdConfigDelete() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
+	cmd.Flags().BoolVarP(&params.Restart, "restart", "r", true, "Restart Cilium pods")
 
 	return cmd
 }
--- a/config/config.go
+++ b/config/config.go
@@ -32,6 +32,7 @@ import (
 type k8sConfigImplementation interface {
 	GetConfigMap(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*corev1.ConfigMap, error)
 	PatchConfigMap(ctx context.Context, namespace, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions) (*corev1.ConfigMap, error)
+	DeletePodCollection(ctx context.Context, namespace string, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error
 }
 
 type K8sConfig struct {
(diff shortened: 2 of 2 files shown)


jrajahalme (approved):

internal/cli/cmd/config.go
```diff
@@ -80,12 +82,33 @@ func newCmdConfigSet() *cobra.Command {
...
+			if err := k8sClient.DeletePodCollection(context.Background(), params.Namespace, metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: defaults.CiliumPodSelector}); err != nil {
+				fmt.Printf("⚠️  Unable to restart Cilium pods: %s\n", err)
+			} else {
+				fmt.Println("♻️  Restarted Cilium pods")
+			}
 			return nil
 		},
 	}
 
-	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
```
tklauser, line 87: We use the single line formatting in all other files in `internal/cli/cmd`. I think we should stay consistent across the code base. If we decide to use multi line formatting, let's change all other occurences too.

internal/cli/cmd/config.go
```diff
@@ -108,7 +131,13 @@ func newCmdConfigDelete() *cobra.Command {
...
```
tklauser, line 111: Same applies here.

tklauser (changes requested):

internal/cli/cmd/config.go
aditighag: I'm happy to change other occurrences in this file since I've already made similar changes. As for other files, I'm inclined to defer it until we have a consensus around the formatting structure.

tklauser: I think in this case we should stick to the single line version in this file as well for this PR and - if we decide that we want to change the format - change all occurrences in the code at once in a follow-up PR.

tklauser (changes requested):
See https://github.com/cilium/cilium-cli/pull/154#discussion_r611744410

internal/cli/cmd/config.go
aditighag: Why is that blocking for this PR?

tklauser: Because this PR actively changes the formatting and the formatting is IMO unrelated to the functionality being introduced.

aditighag: Correct, I'm changing the ones in the file that I'm touching. 

>I dont' think this was caused by gofmt, right?

That's right. I've made the change for improving readability. But this is a good point. I don't see any opinionated comments in the official go documentation. Do you know of any, either for or against such a formatting change?

I was able to find a related GitHub issue for the "stricter fmt" in favor of this formatting style - https://github.com/mvdan/gofumpt/issues/2.

tklauser: > Correct, I'm changing the ones in the file that I'm touching.
> 
> > I dont' think this was caused by gofmt, right?
> 
> That's right. I've made the change for improving readability. But this is a good point. I don't see any opinionated comments in the official go documentation. Do you know of any, either for or against such a formatting change?

I don't know whether there are any "official" Go guidelines. Personally, I prefer for the options to stay single line as it allows for an easy `git grep cmd\.Flags` to get an overview of options with their name and description from the code. Though, I agree that readability might suffer a bit depending on line length.

My review was primarily on the point that we shouldn't switch single occurrence to a multi-line format while keeping others in the existing single-line format. IMO we should strive to stay consistent for cases like these where `gofmt` doesn't enforce a particular formatting.

> I was able to find a related GitHub issue for the "stricter fmt" in favor of this formatting style - [mvdan/gofumpt#2](https://github.com/mvdan/gofumpt/issues/2). Looks like they finally supported the multi-line feature - [mvdan/gofumpt#2 (comment)](https://github.com/mvdan/gofumpt/issues/2#issuecomment-796878776)

Interesting, will check this out.

tgraf: Let's stick to the existing notion. Single line. We use it in the entire codebase.

internal/cli/cmd/config.go
```diff
@@ -86,6 +88,20 @@ func newCmdConfigSet() *cobra.Command {
 			if err := check.Set(context.Background(), args[0], args[1]); err != nil {
 				fatalf("Unable to set config:  %s", err)
 			}
+			if !params.Restart {
+				fmt.Println("⚠️  Restart Cilium pods for configmap changes to take effect")
+				return nil
+			}
+			if err := k8sClient.DeletePodCollection(
```
tgraf, line 95: This should be moved into `Set()` instead as the goal is to make the entire logic by the CLI accessible via a Go API as well.

tgraf (changes requested):

internal/cli/cmd/config.go
aditighag: I'm not sure which Go API you are referring here. ` Set()` is currently called only from newCmdConfigSet, which is the handler that backs the CLI command `set config`. Are you suggesting that `Set()` can also be called from some other API in future?

aditighag:
I realized that we need to add the restart behavior to `delete config` command as well. @tgraf I've also addressed your comments in the latest commit. PTAL.

aditighag:
>AWS::EC2::EIP/NATIP: CREATE_FAILED – "The maximum number of addresses has been reached. 

EKS failures are due to resource unavailability issue.

tgraf (approved):

PR #160, opened 2021-04-13
Install: Upgrade/downgrade support for micro releases

tgraf:
Adds a new "cilium upgrade" command that sets image tags on all required
container images. The command can later be extended to also apply the
required diff to move between minor releases.

```
cilium upgrade
🚀 Upgrading cilium-operator to version quay.io/cilium/operator-generic:v1.9.5...
🚀 Upgrading cilium to version quay.io/cilium/cilium:v1.9.5...
⌛ Waiting for Cilium to be upgraded...
```

### DIFF
--- a/install/upgrade.go
+++ b/install/upgrade.go
@@ -1,0 +1,106 @@
+// Copyright 2021 Authors of Cilium
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package install
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/status"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (k *K8sInstaller) Upgrade(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


install/upgrade.go
```diff
@@ -0,0 +1,106 @@
...
+
+	deployment, err := k.client.GetDeployment(ctx, k.params.Namespace, defaults.OperatorDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to retrieve Deployment of cilium-operator: %s", err)
+	}
+
+	var patched int
+
+	if deployment.Spec.Template.Spec.Containers[0].Image == k.fqOperatorImage() {
+		k.Log("✅ Cilium-operator is already up to date")
```
errordeveloper, line 43: Is this capitalised quite deliberately?

install/upgrade.go
```diff
@@ -0,0 +1,106 @@
...
+			initContainerPatches = append(initContainerPatches, `{"name":"`+c.Name+`", "image":"`+agentImage+`"}`)
+		}
+	}
+
+	if len(containerPatches) == 0 && len(initContainerPatches) == 0 {
+		k.Log("✅ Cilium is already up to date")
+	} else {
+		k.Log("🚀 Upgrading cilium to version %s...", k.fqAgentImage())
+
+		patch := []byte(`{"spec":{"template":{"spec":{"containers":[` + strings.Join(containerPatches, ",") + `], "initContainers":[` + strings.Join(initContainerPatches, ",") + `]}}}}`)
```
errordeveloper, line 75: Just a few thoughts...

Personally, I find this simple patch already quite hairy in the way it's constructed... but I'm also not fully convinced patches are the way to go, but we discussed this and it's worth a shot for sure. We can revisit the approach or the code later.

I guess one thing that is very certain is that this assumes all operations are performed against the API directly, so therefore it doesn't accommodate for a dry-run mode where YAML documents would written to stdout or files. But that applies to the rest of the code, so that'd be a major aspect to revisit down the line anyway.

errordeveloper (commented):
Looks good overall, good starting point for upgrades functionality!

install/upgrade.go
tgraf: This it the minimal PR to support upgrades of micro versions for 1.10. It is convenient for simple default installs. It does not exclude any advanced methods which support generating yamls manually such as via `helm template`.

errordeveloper (approved):

PR #162, opened 2021-04-13
hubble: Add cilium hubble port-forward command

tgraf:
Simplify the local forwarding of the relay port and allow users to run
the command `cilium hubble port-forward`

```
cilium hubble port-forward
Forwarding from 0.0.0.0:4245 -> 4245
Forwarding from [::]:4245 -> 4245
```

### DIFF
--- a/hubble/relay.go
+++ b/hubble/relay.go
@@ -17,6 +17,7 @@ package hubble
 import (
 	"context"
 	"fmt"
+	"os/exec"
 	"time"
 
 	"github.com/cilium/cilium-cli/defaults"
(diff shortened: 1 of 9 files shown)


hubble/relay.go
```diff
@@ -409,3 +410,24 @@ func (k *K8sHubble) createRelayClientCertificate(ctx context.Context) error {
...
+		"svc/hubble-relay",
+		"--address", "0.0.0.0",
+		"--address", "::",
+		fmt.Sprintf("%d:80", k.params.PortForward)}
+
+	c := exec.Command(cmd, args...)
+	c.Stdout = k.params.Writer
+	c.Stderr = k.params.Writer
+
+	if err := c.Run(); err != nil {
```
errordeveloper, line 428: I don't quite recollect if SIGINT will get passed to the child, or do we need to do something about it?

hubble/relay.go
```diff
@@ -409,3 +410,24 @@ func (k *K8sHubble) createRelayClientCertificate(ctx context.Context) error {
 
 	return nil
 }
+
+func (k *K8sHubble) PortForwardCommand(ctx context.Context) error {
+	cmd := "kubectl"
```
errordeveloper, line 415: There is also https://github.com/justinbarrick/go-k8s-portforward

hubble/relay.go
tgraf: Interrupting it works fine so not sure we need to do something

hubble/relay.go
tgraf: I think this is interesting if we want to build it directly into the Hubble client at some point

hubble/relay.go
```diff
@@ -409,3 +410,24 @@ func (k *K8sHubble) createRelayClientCertificate(ctx context.Context) error {
...
+	args := []string{
+		"port-forward",
+		"-n", k.params.Namespace,
+		"svc/hubble-relay",
+		"--address", "0.0.0.0",
+		"--address", "::",
+		fmt.Sprintf("%d:80", k.params.PortForward)}
```
kaworu, line 422: note: this is assuming that the relay service was deployed without TLS enabled (server side).

PR #176, opened 2021-04-20
restart pod after patch configmap

### LINKED ISSUE #165: hubble: Only patch ConfigMap and restart if required
When enabling Hubble, only patch the ConfigMap and restart the Cilium pods when actually required.

jomenxiao:
fix #165


tgraf:
Just noticed this PR. @aditighag was already working on this in #151 which just got merged.

PR #180, opened 2021-04-21
ci: run gke commands via a pod and not from binary

nebril:


nebril:
needs https://github.com/cilium/cilium-cli/pull/181

nebril:
GKE build passed! https://github.com/cilium/cilium-cli/pull/180/checks?check_run_id=2419590535

I will mark it as ready for review, let's keep in mind that we want to move this workflow from `pull_request` to `pull_request_target` before merging. Will do that after reviews are in.

.github/workflows/gke.yaml
```diff
@@ -20,19 +20,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
```
aanm, line 21: I think we can also remove this line so it is explicit that we are not doing anything from any line of code?

aanm (changes requested):
LGTM, only one minor request

.github/workflows/gke.yaml
nebril: We need to check out to get access to master version of the `.github/get-kubeconfig.sh` and helm chart.

aanm: oh I miss that, you are right.

aanm (approved):

PR #196, opened 2021-04-29
Validate image version tag

### LINKED ISSUE #187: Cilium CLI should validate the --version argument
If you pass `--version 1.9.6` into the Cilium CLI, it will pass this directly into the image tag and fail to deploy Cilium (ImagePullBackoff) because the real version number for the image is `v1.9.6`. The CLI should validate the arg and either transparently add the `v` if not specified, or reject the user argument if it is specified in the wrong format.

GoGstickGo:
Attempt to fix #187 . Probably error handling needs more work.


GoGstickGo:
I just realized this approach breaks this:  `cilium clustermesh enable --service-type NodePort`.

GoGstickGo:
What do you think @joestringer  ?

internal/utils/utils.go
```diff
@@ -27,8 +28,14 @@ func BuildImagePath(userImage, defaultImage, userVersion, defaultVersion string)
 		userImage = defaultImage
 	}
 
+	var checkNumber = regexp.MustCompile(`^[1-9].*`).MatchString
```
joestringer, line 31: I think we should be a little more strict here, maybe something more like in this link:

https://github.com/cilium/cilium/blob/006140eca1950690de5cf310219ce6f10c5aa656/contrib/release/lib/gitlib.sh#L42

Note though the regex syntax might not be 100% the same.

What I have in mind is we check something that may or may not have a beginning `v`, followed by a semantic version (also supporting values like `v1.10.0-rc1`). If this doesn't pass, then the version is nonsense.

If the regex is OK, we can do something like `strings.StartsWith(userVersion, "v")` to check whether there is a v in front, if it's not there, just prepend it. If it's there, nothing to do - can just continue using the value.

internal/utils/utils.go
```diff
@@ -27,8 +28,14 @@ func BuildImagePath(userImage, defaultImage, userVersion, defaultVersion string)
...
+
 	if userVersion == "" {
 		userVersion = defaultVersion
+	} else if checkNumber(userVersion) {
+		fmt.Println("version must start v; attaching v to the tag...")
```
joestringer, line 36: I think we probably don't even need to log anything here, we can just take whatever the user gives us as long as it passes the checks I suggested.
```suggestion
```

joestringer (changes requested):
Thanks for picking this up! Looks close to what I had in mind, I have a few extra suggestions below.

joestringer:
oh yeah and for the error case, if the regex doesn't match then maybe we can return an error up to the caller so that it tries to abort the operation?

hubble/relay.go
```diff
@@ -265,7 +266,12 @@ disable-server-tls: true
 }
 
 func (k *K8sHubble) relayImage() string {
-	return utils.BuildImagePath(k.params.RelayImage, defaults.AgentImage, k.params.RelayVersion, defaults.Version)
+	image, err := utils.BuildImagePath(k.params.RelayImage, defaults.AgentImage, k.params.RelayVersion, defaults.Version)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
joestringer, line 272: I think it may be a bit brittle to have this kind of "print and exit" logic sprinkled around these various functions, since the caller likely does not expect that calling into functions like `relayImage()` could cause the whole process to terminate. This could end up leaving the cluster in a weird partially-initialized state if the user passes a bad argument.

Ideally the validation of the version should occur very early in the CLI before the CLI takes any action in the cluster, then these "image path build" functions can just trust that they will successfully build a valid version.

GoGstickGo: Do you mean place  the validation into for example: https://github.com/cilium/cilium-cli/blob/37d6bd6f4c75a9bf4e415b33a255c16d128c2ddf/internal/cli/cmd/install.go#L50 ? In this case CLI wouldn't even pass it if it fails.

joestringer: That's close. Digging from there through `NewK8sInstaller()` gets to `InstallParameters.validate()`, which sounds exactly like the right place for such logic:

https://github.com/cilium/cilium-cli/blob/37d6bd6f4c75a9bf4e415b33a255c16d128c2ddf/install/install.go#L1045

GoGstickGo:
I also added logic to `clustermesh` to standardized image path creation.

defaults/defaults.go
```diff
@@ -39,6 +39,8 @@ const (
 
 	HubbleSocketPath       = "/var/run/cilium/hubble.sock"
 	HubbleServerSecretName = "hubble-server-certs"
+  	HubbleServerSecretCertName = "tls.crt"
```
joestringer, line 42: Why are these changes necessary? :thinking:

install/install.go
```diff
@@ -1052,6 +1053,10 @@ func (p *InstallParameters) validate() error {
 
 		p.configOverwrites[t[0]] = t[1]
 	}
+	var checkVersion = regexp.MustCompile(`(0|[1-9][0-9]*)*\.*(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?`).MatchString
```
joestringer, line 1056: This regex seems a bit different from the one we had discussed before, in particular there seems to be extra `*` following some of the `\.`, which means that it could accept versions like "0..1.0", which I don't think is intended.

I wonder if there's a better way to get a canonical version regex instead of rolling our own here :thinking: 

On the other hand, if we do need to make our own here, it would be good to create a small unit test to try out various correct and incorrect versions against the regex to check that the regex is linting the input as we would expect.

joestringer (changes requested):
Couple more notes below. At some point I think it would be good also to squash the commits together all into one commit. If you're not familiar with how to do this, there are guides available to help:

https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec

install/install.go
joestringer: ^ On the same train of thought, it would be good to declare the regex only once and use it in multiple places so that this one and the other one don't get out-of-sync (and if we need to fix it again, we can do that once).

defaults/defaults.go
GoGstickGo: There was conflict with master, so I fixed it.

joestringer (changes requested):
LGTM, thanks! I like the idea of not enforcing when a custom image is specified, as that will do a good job of covering the case where developers want to test something out with a custom image tag.

Could you please rebase against the latest master branch to avoid the merge commit and sign off your commits (`git commit --amend --signoff`) according to the [Developer's Certificate of Origin](https://docs.cilium.io/en/v1.9/contributing/development/contributing_guide/#developer-s-certificate-of-origin)?

internal/utils/utils.go
```diff
@@ -17,20 +17,29 @@ package utils
...
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/cilium/cilium-cli/defaults"
 )
 
+func CheckVersion(version string) bool {
+	var check = regexp.MustCompile(`(0|[1-9][0-9]*)*\.*(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?`).MatchString
```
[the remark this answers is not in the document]
GoGstickGo: Actually it works better with `?` instead  `\.*`. Let me update that.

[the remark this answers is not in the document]
joestringer: Yeah I was wondering about checking for whether the image legitimately exists or not too but it seemed a bit overkill. Also, it's plausible that you could define an image tag that is accessible from the cluster but not accessible from where your CLI is running (eg private mirror for your organization) so it might be just a bit too fancy.

GoGstickGo:
I am closing this PR due to local git issues.

PR #193, opened 2021-04-28
Pin image tags for arm64 support, and update hubble tags

aanm:
Read per commit.
Fixes #191

PR #198, opened 2021-04-29
connectivity/check: bump connectivity check images for arm64 support

tklauser:
Follows cilium/cilium#15894

Also define the image name/version in package defaults.


aanm:
being addressed in https://github.com/cilium/cilium-cli/pull/193

tklauser:
Oops, didn't noticed #193 which already does this. Closing.

PR #199, opened 2021-04-29
Prepare for v0.5 release

joestringer:
- Prepare for release v0.5.0
- Prepare for 0.6 development

After merge, I can tag the commit with the v0.5.0 and trigger the release process.


VERSION
```diff
@@ -1 +1 @@
-0.5-dev
+0.6.0-dev
```
joestringer, line 1: I can alternatively set this to 0.5.1-dev if we expect to follow up with other fixes shortly and don't want to bump the version too quickly

tgraf (approved):

joestringer:
Did we break authentication tokens for some of the CI checks?

joestringer:
Ah right, we don't support PRs with a base branch outside the main repository at this time so it's expected that all the checks fail.

Issue #205, opened 2021-05-02
Idempotent Cilium Install ?


lfundaro:
Just noticed that when issuing `cilium install` 2x in a row, the second execution fails because resources (service accounts, config maps, daemonsets, etc) have been created already. Is this intended ? 

How to reproduce:

`Release v0.5`
```
$> cilium install
...
...
✅ 
$> cilium install 
...
..
Error: Unable to install Cilium:  unable to create secret kube-system/hubble-server-certs: secrets "hubble-server-certs" already exists
```

From a security perspective, it could make sense to abort installation, i.e.: one could argue that the cluster has been compromised with a certificate that an attacker has access to and all cilium-related bytes can be read, so in this case it's good to abort the installation. If we want to abort installation, should we at least say why we are aborting ? From a users perspective, reading this message doesn't tell me much on what next steps I should take to fix this. 

P.S.: I opened a [PR](https://github.com/cilium/cilium-cli/pull/206) that goes for idempotent. No tests yet, since I'm not sure if idempotency is not intended here. Alternatively, we could add a friendly message. Let me know what you think.

Cheers !

lfundaro:
closing as we don't need idempotency. The issue evolved into telling users to do a `cilium upgrade` if they are trying to `reinstall` cilium.

lfundaro:
closing as we don't need idempotency. The issue evolved into telling users to do a `cilium upgrade` if they are trying to reinstall cilium.

joestringer:
I'd like to reopen this feature request as it would simplify declarative approaches to defining the way that I want Cilium to be created in an environment. From the PR discussion, it looks like there is some desire still to do idempotency, it just needs to take into account that upgrading an existing environment should not replace secrets etc.

PR #206, opened 2021-05-02
Make cilium install idempotent

lfundaro:
[No tests yet, waiting for discussion with Authors]

This PR makes the command cilium install idempotent.
It checks the error returned from the API and continues
in case the resource has been created already.

Fixes: #205

Signed-off-by: Lorenzo Fundaró <lorenzofundaro@gmail.com>

### DIFF
--- a/install/certs.go
+++ b/install/certs.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cloudflare/cfssl/config"
 	"github.com/cloudflare/cfssl/csr"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -61,7 +62,9 @@ func (k *K8sInstaller) createHubbleServerCertificate(ctx context.Context) error 
 
 	_, err = k.client.CreateSecret(ctx, k.params.Namespace, k8s.NewTLSSecret(defaults.HubbleServerSecretName, k.params.Namespace, data), metav1.CreateOptions{})
 	if err != nil {
-		return fmt.Errorf("unable to create secret %s/%s: %w", k.params.Namespace, defaults.HubbleServerSecretName, err)
+		if ok := errors.IsAlreadyExists(err); !ok {
+			return fmt.Errorf("unable to create secret %s/%s: %w", k.params.Namespace, defaults.HubbleServerSecretName, err)
+		}
 	}
 
 	return nil
--- a/install/install.go
+++ b/install/install.go
@@ -33,6 +33,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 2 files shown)


lfundaro:
No tests yet, since there remains the question whether `cilium install` should be idempotent.

pchaigno:
From the issue reactions, it looks like it would be appreciated :-)

Could you update the PR with tests? I switched to draft mode in the meantime.

lfundaro:
> From the issue reactions, it looks like it would be appreciated :-)
> 
> Could you update the PR with tests? I switched to draft mode in the meantime.

sure  👍 . I'll ping you back when I have something to review. Cheers.

lfundaro:
@pchaigno, one question, in this PR we are skipping the installation of certificates if we see there is a resource already there. Is this the behavior that we want ? I'm thinking some edge cases:
1. As an operator, if I want to update all cilium certificates I could just issue `cilium install` and expect that all certificates would be refreshed. In this case, instead of skipping like in [this block](https://github.com/cilium/cilium-cli/pull/206/files#diff-3c22b152a3ea57656642ac018f2fb593bc485682cb44f2475328299745f52940R63-R67) we would `upsert` the certificate, i.e.: if the k8s secret already exits, we just refresh `data` with a new cert, otherwise we create a new k8s secret.
2. From a security perspective, we may want `cilium-cli` to be authoritative on the installation process as a whole, i.e.: we guarantee that installed certificates come exclusively from the `cli` instead from unknown sources. 

wdyt ? 

I'm leaning towards `upsert` but I'm open to discuss.

pchaigno:
AFAIU, option 1 also means the command won't actually be idempotent. Probably worth asking the wider community for opinions on #development (although note many folks are currently OOO with the long weekend).

lfundaro:
> AFAIU, option 1 also means the command won't actually be idempotent. Probably worth asking the wider community for opinions on #development (although note many folks are currently OOO with the long weekend).

yes, we would lose idempotency, but if option 2 is more important then we may as well forget about that property. I will ask on a wider forum. Cheers !

ti-mo:
> As an operator, if I want to update all cilium certificates I could just issue cilium install and expect that all certificates would be refreshed. In this case, instead of skipping like in this block we would upsert the certificate, i.e.: if the k8s secret already exits, we just refresh data with a new cert, otherwise we create a new k8s secret.

Making `cilium install` perform cert upserts would be surprising to me as an operator. This probably belongs in `cilium upgrade` (following Helm behaviour) or might even warrant a new `cilium rotate` or `cilium encryption rotate` command (for clustermesh/wg/ipsec).

As discussed on Slack, introducing skips does not add idempotency. It will cause headaches when we make this install routine create more resources and a user runs `cilium install` instead of `cilium upgrade`. Either we merge `cilium upgrade` into `cilium install` and make the whole lot fully idempotent (including diffs and dry-run), or we simply point the user to `cilium upgrade` when one of the resources created in `cilium install` already exists, with a graceful exit code.

lfundaro:
> or we simply point the user to cilium upgrade when one of the resources created in cilium install already exists, with a graceful exit code.

I would lean towards this approach. In practical terms, it yields the same value as merging `upgrade` into `install` and is waaaay much simpler. 😄 👍 

I'll ping you both when I have a commit. 

Cheers !

lfundaro:
closing this PR and opening #282 since this is no longer about making install idempotent. cc @ti-mo @pchaigno

PR #207, opened 2021-05-03
handle default image and versions at the flag level

kaworu:
I got bitten trying `cilium install --version 1.10.0-rc1` (notice the missing `v` prefix). `cilium install --help` wasn't helpful about the expected "format" of the `--version` flag.

I've considered checking and adding the `v` prefix to `version` in `utils.BuildImagePath` when it was missing, but ultimately elected against it in order to support using non-version tags (e.g. `stable`). Maybe we could try to do something smart like checking against `^\d+\.\d+\.d+` and adding the `v` prefix if it matches, happy to hear thoughts about this.

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -20,7 +20,6 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/internal/k8s"
 )
 
@@ -98,10 +97,7 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
 		}
 	}
 
-	if k.params.Version == "" {
-		k.Log("ℹ️  Cilium version not set, using default version %q", defaults.Version)
-		k.params.Version = defaults.Version
-	}
+	k.Log("ℹ️  using Cilium version %q", k.params.Version)
 
 	if k.params.ClusterName == "" {
 		if f.ClusterName != "" {
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -74,8 +74,8 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringVar(&params.Encryption, "encryption", "disabled", "Enable encryption of all workloads traffic { disabled | ipsec | wireguard }")
 	cmd.Flags().BoolVar(&params.NodeEncryption, "node-encryption", false, "Enable encryption of all node to node traffic")
 	cmd.Flags().StringSliceVar(&params.ConfigOverwrites, "config", []string{}, "Set ConfigMap entries (key=value)")
-	cmd.Flags().StringVar(&params.AgentImage, "agent-image", "", "Image path to use for Cilium agent")
-	cmd.Flags().StringVar(&params.OperatorImage, "operator-image", "", "Image path to use for Cilium operator")
+	cmd.Flags().StringVar(&params.AgentImage, "agent-image", defaults.AgentImage, "Image path to use for Cilium agent")
+	cmd.Flags().StringVar(&params.OperatorImage, "operator-image", defaults.OperatorImage, "Image path to use for Cilium operator")
 	cmd.Flags().DurationVar(&params.CiliumReadyTimeout, "cilium-ready-timeout", 5*time.Minute,
 		"Timeout for Cilium to become ready before restarting unmanaged pods")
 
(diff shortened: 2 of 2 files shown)


tklauser (approved):
This bit me in the past as well. Thanks for the fix! And sorry for the delayed review.

> Maybe we could try to do something smart like checking against ^\d+\.\d+\.d+ and adding the v prefix if it matches, happy to hear thoughts about this.

That would certainly be nice as a follow-up. Adding the `v` prefix in case a version numeric version is matched SGTM. This would still allow to use non-version tags such as `latest` (e.g. when using `quay.io/cilium/cilium-ci` to test the latest Cilium `master` image).

tklauser:
> This bit me in the past as well. Thanks for the fix! And sorry for the delayed review.
> 
> > Maybe we could try to do something smart like checking against ^\d+.\d+.d+ and adding the v prefix if it matches, happy to hear thoughts about this.
> 
> That would certainly be nice as a follow-up. Adding the `v` prefix in case a version numeric version is matched SGTM. This would still allow to use non-version tags such as `latest` (e.g. when using `quay.io/cilium/cilium-ci` to test the latest Cilium `master` image).

Just noticed that #219 is implementing something along these lines.

kaworu:
test-me-please

internal/utils/utils.go
```diff
@@ -22,16 +22,8 @@ import (
...
-		userImage = defaultImage
-	}
-
-	if userVersion == "" {
-		userVersion = defaultVersion
-	}
-
-	return userImage + ":" + userVersion
+func BuildImagePath(image, version string) string {
+	return image + ":" + version
```
aditighag, line 26: Nyc, but since you are making changes anyway, I'll post my comment here. The version string shouldn't be appended to a custom image.

aditighag: Looks like the correct format to pass a custom image is - `--agent-image quay.io/foo/ --version bar`. This will then be computed as `quay.io/foo:bar`. This isn't intuitive! @kaworu Could you add an example to the help dump?

aditighag (commented):
>cilium install --help wasn't helpful about the expected "format" of the --version flag.

I've left a comment along the lines of your commit description. Let me know if you are opposed to accommodating it in this PR.

internal/utils/utils.go
aditighag: Sounds good to me! Thanks!

internal/cli/cmd/install.go
```diff
@@ -60,7 +60,7 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace to install Cilium into")
 	cmd.Flags().StringVar(&params.ClusterName, "cluster-name", "", "Name of the cluster")
 	cmd.Flags().StringSliceVar(&params.DisableChecks, "disable-check", []string{}, "Disable a particular validation check")
-	cmd.Flags().StringVar(&params.Version, "version", "", "Cilium version to install")
+	cmd.Flags().StringVar(&params.Version, "version", defaults.Version, "Cilium version to install")
```
aditighag, line 63: I think this would mention the current default version in the help output. If not, can we add that?

aditighag (approved):

internal/cli/cmd/install.go
kaworu: Yes, the help output show the default version after this patch.

Before the patch:
```console
% go run ./cmd/cilium install --help
...
      --version string                  Cilium version to install
...
```

After the patch:
```console
% go run ./cmd/cilium install --help
...
      --version string                  Cilium version to install (default "v1.9.7")
...
```

gandro (approved):
LGTM! One thing to point out is that this line here will basically not be printed anymore after this patch (unless the user goes out of their way to specify `--version=""`).

https://github.com/cilium/cilium-cli/blob/42d087f9df4876b8c7acfe5ab79c19e0c5edccc9/install/autodetect.go#L102

kaworu:
@gandro nice catch! I've updated the patch to print the version unconditionally.

install/autodetect.go
```diff
@@ -98,10 +98,7 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
 		}
 	}
 
-	if k.params.Version == "" {
-		k.Log("ℹ️  Cilium version not set, using default version %q", defaults.Version)
-		k.params.Version = defaults.Version
-	}
+	k.Log("ℹ️  Cilium version not set, using Cilium version %q", defaults.Version)
```
gandro, line 101: I think this statement here not correct anymore. If the user specifies `k.params.Version`, then this will still print `defaults.Version`?

gandro (changes requested):

install/autodetect.go
kaworu: ouch :facepalm: my bad, thank you for pointing it out

gandro (approved):

michi-covalent (approved):

rolinh (approved):

PR #208, opened 2021-05-03
install: don't delete the 'aws-node' DaemonSet

bmcustodio:
Instead of deleting the `aws-node` DaemonSet, which is permanent, we can be gentler and just evict its pods by setting a node selector that won't ever be matched (unless someone wants it to by explicitly setting the required label on a node).

This allows `cilium uninstall` to gracefully revert the `aws-node` `DaemonSet` to its previous state, as it might contain customizations that would otherwise be lost.

### DIFF
--- a/.github/workflows/eks.yaml
+++ b/.github/workflows/eks.yaml
@@ -106,6 +106,10 @@ jobs:
       - name: Wait for test job
         run: |
           kubectl -n kube-system wait job/cilium-cli --for=condition=complete --timeout=10m
+      
+      - name: Make sure the 'aws-node' DaemonSet exists but has no scheduled pods
+        run: |
+          [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 0 ]]
 
       - name: Post-test information gathering
         if: ${{ failure() }}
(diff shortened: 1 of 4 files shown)


tgraf (approved):

pchaigno (approved):
This is very nice! :rocket: 

Could we test it by adding a check in the EKS workflows, to verify the aws-node DaemonSet's pods exists after Cilium is uninstalled? Or is it not worth it?

tklauser:
Hi @bmcustodio. Could you please rebase the PR branch against latest `master`? This should hopefully get rid of most of the test flakes.

Also there's a [review remark](https://github.com/cilium/cilium-cli/pull/208#pullrequestreview-670101619) from @pchaigno which you might want to address, if feasible?

bmcustodio:
@tklauser @pchaigno sorry for totally missing this review 🤦🏼 I've added said check and rebased!

tklauser (approved):

tklauser:
For some reason the GitHub actions weren't triggered when you pushed a new version. I'll close and re-open the PR hoping that this will trigger the tests.

joestringer (approved):

joestringer:
AKS failure is only due to the PR being opened from outside the cilium-cli repo. The code changes themselves only touch the EKS-specific jobs and codepaths so there should be no regression in the AKS job. Should be good to merge.

PR #212, opened 2021-05-05
run most actions tests from within cluster

nebril:
This change allows all GH Actions workflows aside from AKS one to be run against forks from external contributors without exposing our cloud provider secrets.

AKS integration is too tightly coupled with our secrets (we are passing aks secrets to Cilium datapath config) which makes in impossible to not run "unsafe" cilium-cli binary in an environment where the secrets are not available.


nebril:
Kind run is not related, being fixed in https://github.com/cilium/cilium-cli/pull/221

nebril:
Ready for review! After reviews are in, I will change all workflows to use `pull_request_target` event.

michi-covalent (approved):
let's go

tklauser (approved):
Thanks

nbusseneau (commented):
I'm afraid I won't have time to review this until Friday, this is a big piece. I've glanced over and from a high-level overview it looks great, but if at all possible I'd love to go through it again with a comb when I have more time so that I'm 100% convinced :D

clustermesh/clustermesh.go
```diff
@@ -1401,7 +1401,7 @@ ${SUDO} docker cp cilium:/usr/bin/cilium /usr/bin/cilium
 
 # Wait for cilium agent to become available
 cilium_started=false
-for ((i = 0 ; i < 24; i++)); do
+for ((i = 0 ; i < 48; i++)); do
```
nbusseneau, line 1404: Leftover from testing?

.github/workflows/externalworkloads.yaml
```diff
@@ -85,21 +104,42 @@ jobs:
...
+          kubectl logs -n kube-system job/cilium-cli-install
+
+      - name: Post-test information gathering
         if: ${{ always() }}
         run: |
-          cilium status
-          cilium clustermesh status
-          cilium clustermesh vm status
+          kubectl logs -n kube-system job/cilium-cli
+          gcloud compute ssh ${{ env.vmName }} --zone ${{ env.zone }} --command "cilium status"
```
nbusseneau, line 131: Can we do something similar on other workflows to retrieve `cilium status` as we did before?
Likewise in this workflow for `cilium clustermesh vm status` which seems to have disappeared here.

nbusseneau (commented):
Hey, some nits below. Also, there's an empty `trolo` file in last commit :)

.github/workflows/externalworkloads.yaml
nebril: `cilium status` and `cilium clustermesh vm status` is added at the end of test jobs (after connectivity check).

.github/in-cluster-test-script-eks.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cilium hubble enable
+
+cilium status --wait
+
+cilium hubble port-forward&
+
+sleep 5
+
+cilium connectivity test --test '!/pod-to-local-nodeport' --all-flows
```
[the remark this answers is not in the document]
nebril: This is copy pasted from what was in the script, I would rather change what tests we run in a separate PR to minimize chances of breaking something.

.github/in-cluster-test-script-eks-install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+set -x
+set -e
+
+cilium install --cluster-name "${CLUSTER_NAME}" --restart-unmanaged-pods=false --config monitor-aggregation=none
```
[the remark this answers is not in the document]
nebril: Good catch

nebril:
Moved all changed workflows to `pull_request_target`, that's why they don't show up in this PR now.

.github/in-cluster-test-script-eks.sh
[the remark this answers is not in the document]
nbusseneau: OK.

.github/workflows/eks-tunnel.yaml
```diff
@@ -50,38 +54,58 @@ jobs:
         run: |
           ./eksctl create cluster --name ${{ env.clusterName }} --without-nodegroup
 
-      - name: Install cilium
+      - name: Create gcloud-free kubeconfig and load it in configmap
```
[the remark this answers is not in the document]
nbusseneau: Forgot to mention there are a bunch of other `gcloud-free` in the comments, but it's not really important.

nbusseneau (approved):
Thanks, one small nit left from previous conversations above but it's not very important.

pchaigno:
This PR needs a rebase after which it should be ready to merge.

.github/workflows/kind.yaml
```diff
@@ -50,6 +50,7 @@ jobs:
       - name: Enable Relay
         run: |
           cilium hubble enable --ui
+          cilium status --wait --wait-duration=5m
```
nbusseneau, line 53: Still needed since we now have `cilium status --wait` in next step?

.github/workflows/externalworkloads.yaml
```diff
@@ -53,27 +40,59 @@ jobs:
...
+            OWNER=${{ github.sha }}
+          fi
+          echo ::set-output name=sha::${SHA}
+
       - name: Create GCP VM
         run: |
           gcloud compute instances create ${{ env.vmName }} --image-project ubuntu-os-cloud --image-family ubuntu-2010 --machine-type n1-standard-4 --zone ${{ env.zone }} --metadata hostname=${{ env.vmName }} --metadata-from-file startup-script=${{ env.vmStartupScript }}
 
-      - name: Install cilium in cluster
+      - name: Create gcloud-free kubeconfig and load it in configmap
```
nbusseneau, line 66: [nit] Still some `gloud-free` mentions here and in other workflows :D

nbusseneau (approved):
LGTM, small nits.

.github/workflows/externalworkloads.yaml
nebril: This runs on gcloud, so I think it can stay :D.

.github/workflows/kind.yaml
nebril: It wasn't there xD. Dropping the commit.

.github/workflows/externalworkloads.yaml
```diff
@@ -87,21 +106,42 @@ jobs:
...
+      - name: Create cilium-cli test job
+        run: |
+          helm install .github/cilium-cli-test-job-chart --generate-name \
+          --set tag=${{ steps.vars.outputs.sha }} \
+          --set job_name=cilium-cli \
+          --set test_script_cm=cilium-cli-test-script \
+
+      - name: Wait for test job
+        run: |
+          kubectl -n kube-system wait job/cilium-cli --for=condition=complete --timeout=10m
```
pchaigno, line 122: Doesn't this completely hide the test output in case of failure?

It used to fail with the output of cilium connectivity test clearly detailed what failed. Now it just says a timeout was reached...

.github/workflows/eks.yaml
```diff
@@ -2,12 +2,12 @@ name: EKS (ENI)
...
 # Any change in triggers needs to be reflected in the concurrency group.
 on:
-  pull_request: {}
+  pull_request_target: {}
   schedule:
     - cron:  '10 */6 * * *'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after || 'scheduled' }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request_target.number || github.event.after || 'scheduled' }}
```
pchaigno, line 10: I don't think there is such a thing as `github.event.pull_request_target`. All jobs are currently cancelling each other because of this. I'll send a fix.

PR #219, opened 2021-05-11
Validate image version

GoGstickGo:
Reopening this PR: https://github.com/cilium/cilium-cli/pull/196

Fixes: #187

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -164,7 +164,7 @@ etcdctl auth enable;
 exit`}
 
 func (k *K8sClusterMesh) apiserverImage() string {
-	return utils.BuildImagePath(k.params.ApiserverImage, defaults.ClusterMeshApiserverImage, k.params.ApiserverVersion, defaults.Version)
+	return utils.BuildImagePath(k.params.ApiserverImage, defaults.ClusterMeshApiserverImage, k.params.ApiserverVersion, k.imageVersion)
 }
 
 func (k *K8sClusterMesh) generateDeployment() *appsv1.Deployment {
@@ -435,6 +435,7 @@ type K8sClusterMesh struct {
 	params          Parameters
 	clusterName     string
 	clusterID       string
+	imageVersion    string
 }
 
 type Parameters struct {
@@ -518,6 +519,19 @@ func (k *K8sClusterMesh) GetClusterConfig(ctx context.Context) error {
 	return nil
 }
 
+func (k *K8sClusterMesh) getRunningCiliumVersion(ctx context.Context) error {
+	pods, err := k.client.ListPods(ctx, k.params.Namespace, metav1.ListOptions{LabelSelector: "k8s-app=cilium"})
+	if err != nil {
+		return fmt.Errorf("unable to list pods: %w", err)
+	}
+	for _, pod := range pods.Items {
+		image := pod.Spec.Containers[0].Image
+		version := strings.SplitN(image, ":", 2)
+		k.imageVersion = version[1]
+	}
+	return nil
+}
+
 func (k *K8sClusterMesh) Disable(ctx context.Context) error {
 	k.Log("🔥 Deleting clustermesh-apiserver...")
 	k.client.DeleteService(ctx, k.params.Namespace, defaults.ClusterMeshServiceName, metav1.DeleteOptions{})
@@ -556,6 +570,10 @@ func (k *K8sClusterMesh) Enable(ctx context.Context) error {
 	}
 
 	if err := k.GetClusterConfig(ctx); err != nil {
+		return err
+	}
+
+	if err := k.getRunningCiliumVersion(ctx); err != nil {
 		return err
 	}
 
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -462,6 +462,15 @@ type Parameters struct {
 	Retries              int
 }
 
+func (p Parameters) validateParams() error {
+	if p.ApiserverImage != defaults.ClusterMeshApiserverImage {
+		return nil
+	} else if !utils.CheckVersion(p.ApiserverVersion) && p.ApiserverVersion != "" {
+		return fmt.Errorf("invalid syntax %q for image tag", p.ApiserverVersion)
+	}
+	return nil
+}
+
 func (p Parameters) waitTimeout() time.Duration {
 	if p.WaitDuration != time.Duration(0) {
 		return p.WaitDuration
@@ -542,6 +551,10 @@ func (p Parameters) validateForEnable() error {
 }
 
 func (k *K8sClusterMesh) Enable(ctx context.Context) error {
+	if err := k.params.validateParams(); err != nil {
+		return err
+	}
+
 	if err := k.params.validateForEnable(); err != nil {
 		return err
 	}
@@ -566,6 +579,7 @@ func (k *K8sClusterMesh) Enable(ctx context.Context) error {
 	}
 
 	k.Log("✨ Deploying clustermesh-apiserver...")
+
 	if _, err := k.client.CreateServiceAccount(ctx, k.params.Namespace, k8s.NewServiceAccount(defaults.ClusterMeshServiceAccountName), metav1.CreateOptions{}); err != nil {
 		return err
 	}
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -592,8 +592,7 @@ func (k *K8sClusterMesh) Enable(ctx context.Context) error {
 		return err
 	}
 
-	k.Log("✨ Deploying clustermesh-apiserver...")
-
+	k.Log("✨ Deploying clustermesh-apiserver from %s...", k.apiserverImage())
 	if _, err := k.client.CreateServiceAccount(ctx, k.params.Namespace, k8s.NewServiceAccount(defaults.ClusterMeshServiceAccountName), metav1.CreateOptions{}); err != nil {
 		return err
 	}
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -164,11 +164,7 @@ etcdctl auth enable;
 exit`}
 
 func (k *K8sClusterMesh) apiserverImage() string {
-	if k.params.ApiserverImage != "" {
-		return k.params.ApiserverImage
-	}
-
-	return defaults.ClusterMeshApiserverImage
+	return utils.BuildImagePath(k.params.ApiserverImage, defaults.ClusterMeshApiserverImage, k.params.ApiserverVersion, defaults.Version)
 }
 
 func (k *K8sClusterMesh) generateDeployment() *appsv1.Deployment {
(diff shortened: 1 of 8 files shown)


internal/utils/utils.go
```diff
@@ -17,20 +17,34 @@ package utils
...
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/cilium/cilium-cli/defaults"
 )
 
+func CheckVersion(version string) bool {
+	var check1 = regexp.MustCompile(`[^\.](0|[1-9][0-9]*\.)?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?`).MatchString
```
GoGstickGo, line 28: I was trying to do something like this: `(?:^[v0-9]|[^a-zA-Z\.])(0|[1-9][0-9]*\.)?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?` but I couldn't overcome issue whether string starts with or without `v`.

internal/utils/utils_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestCheckVersion(t *testing.T) {
+	tests := []struct {
+		name string
+		want bool
+	}{
+		{"v1.9.6", true},
+		{"1.9.6", true},
+		{"1.9", true},
+		{"v1.9", true},
```
joestringer, line 27: Interesting. Incidentally, we don't distribute official images with these tags so we should probably prevent these (unless we have some other logic to somehow resolve this to a specific valid tag).

internal/utils/utils.go
```diff
@@ -17,20 +17,29 @@ package utils
...
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/cilium/cilium-cli/defaults"
 )
 
+func CheckVersion(version string) bool {
+	var check = regexp.MustCompile(`^([v]|0|[1-9][0-9]*\.)?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?`).MatchString
```
joestringer, line 28: Does this now allow `v.1.0`? I'm not sure about the `[v]|` since the OR operator (`|`) seems like it is selecting between either starting with a v or a 0 or a number series starting with a 1.

joestringer: And for the other question about discouraging `v1.9`, I think that's because the first section here (before the first `?`) may all occur zero or one times due to the `?`. But we probably just want that first section to always occur, so removing the question mark and moving the period literal (`\.`) outside the brackets.

joestringer (changes requested):
Awesome, those unit tests are making it pretty clear what is allowed or disallowed, the feedback cycle should be much quicker now since it's easy to add examples we may want to allow / disallow and test them out.

I found a couple more interesting corner cases, described below.

internal/utils/utils.go
```diff
@@ -17,20 +17,29 @@ package utils
...
 	if userImage == "" {
 		userImage = defaultImage
 	}
 
-	if userVersion == "" {
+	switch {
+	case userVersion == "":
 		userVersion = defaultVersion
+	case !strings.HasPrefix(userVersion, "v"):
+		userVersion = "v" + userVersion
```
joestringer, line 41: If a developer passes `--agent-image docker.io/alice/cilium`  and `--version my-custom-version`, will this produce `docker.io/alice/cilium:vmy-custom-version`?

I'm wondering if we should actually do this `v` prepend earlier as well, at the same place where we choose whether to perform the regex validation or not.

joestringer (commented):
Looks good to me. I have one more question below for if someone passes a custom non-version image tag, would be good to keep support for that too.

internal/utils/utils.go
GoGstickGo: Good catch fixed. I realized another thing as there is only one `--version` command in https://github.com/cilium/cilium-cli/blob/373236d85ef4812f9602b7f449dbfa56dc527675/install/install.go#L1034 however in the same place  there are two different commands for `--agent-image` and ` --operator-image`.   Can `cilium agent` and `cilium operator` run in different versions , if so shouldn't we support this case ?

joestringer: I wouldn't worry too much about handling that case for now. For most users, it will be one consistent version like `v1.10.0`. For developers, they will often specify an image built from the CI which should use a consistent tag as well I think.

clustermesh/clustermesh.go
```diff
@@ -510,6 +517,19 @@ func (k *K8sClusterMesh) GetClusterConfig(ctx context.Context) error {
...
 
+func (k *K8sClusterMesh) getRunningCiliumVersion(ctx context.Context) error {
+	pods, err := k.client.ListPods(ctx, k.params.Namespace, metav1.ListOptions{LabelSelector: "k8s-app=cilium"})
+	if err != nil {
+		return fmt.Errorf("unable to list pods: %w", err)
+	}
+	for _, pod := range pods.Items {
+		image := pod.Spec.Containers[0].Image
+		version := strings.Split(image, ":")
+		k.imageVersion = version[1]
```
joestringer, line 528: I don't remember seeing this before, but FYI image tags can be of the form `docker.io/cilium/operator-generic:v1.9.7@sha256:8f121bf312654de62eeec0076755411b088ce4c48843b174d206ea1b141fbb7c` which has two colons, so the [1] index isn't exactly correct for this case (in fact, it will be wrong because it doesn't include the full sha).

GoGstickGo: I believe `strings.SplitN("string", ":", 2)` solve this issue as image:tag syntax set in stone.

internal/utils/utils.go
```diff
@@ -17,20 +17,28 @@ package utils
...
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/cilium/cilium-cli/defaults"
 )
 
+func CheckVersion(version string) bool {
+	var check = regexp.MustCompile(`^([v]|0|[1-9][0-9]*\.)?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-[a-zA-Z0-9]+)*\.*(0|[1-9][0-9]*)?`).MatchString
```
christarazi, line 28: This can be moved out of the function and can be created in the package scope instead. That way it doesn't have to be recreated on each call to `CheckVersion()`

internal/cli/cmd/clustermesh.go
```diff
@@ -68,6 +68,7 @@ func newCmdClusterMeshEnable() *cobra.Command {
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
 	cmd.Flags().StringVar(&params.ServiceType, "service-type", "", "Type of Kubernetes service to expose control plane { ClusterIP | LoadBalancer | NodePort }")
 	cmd.Flags().StringVar(&params.ApiserverImage, "apiserver-image", "", "Container image for clustermesh-apiserver")
+	cmd.Flags().StringVar(&params.ApiserverVersion, "apiserver-version", "", "Container image version for clustermesh-apiserver")
```
christarazi, line 71: I think this should probably be in its own commit, `Add flag for clustermesh-apiserver image version`

christarazi (changes requested):
Thanks for the PR. Some comments below.

* Why only Hubble and Clustermesh are calllng `validateParams()`? What about the others?
* In commit `Validate image version `, please write what changed and why in the commit body rather than the current list in there now.

Thanks for adding the unit tests btw :rocket:

GoGstickGo:
> Thanks for the PR. Some comments below.
> 
>     * Why only Hubble and Clustermesh are calllng `validateParams()`? What about the others?
> 
>     * In commit `Validate image version `, please write what changed and why in the commit body rather than the current list in there now.
> 
> 
> Thanks for adding the unit tests btw 🚀

What do you mean others for `validateParams()`. I only added to Hubble and Clustermesh as cilium main install already has it:https://github.com/cilium/cilium-cli/blob/d73fbfe058356e23c2cc4cf01b773233f31b0da2/install/install.go#L1058
I couldn't name the same as there is already `validate()` in https://github.com/cilium/cilium-cli/blob/d73fbfe058356e23c2cc4cf01b773233f31b0da2/hubble/hubble.go#L102 and https://github.com/cilium/cilium-cli/blob/d73fbfe058356e23c2cc4cf01b773233f31b0da2/clustermesh/clustermesh.go#L612

christarazi:
@GoGstickGo What is meant is that if we are going to validate the images for Clustermesh and Hubble, then why not do it for everything else like the agent and the operator images: https://github.com/cilium/cilium-cli/blob/d73fbfe058356e23c2cc4cf01b773233f31b0da2/install/install.go#L1037

> In commit `Validate image version `, please write what changed and why in the commit body rather than the current list in there now.

The above item is not addressed. While the list in useful, it would be better if we separated out fixes into their own commits. Not every item in the list corresponds to a commit, so some things can be combined together. For example, the change that adds the Clustermesh arg for an image can be separated. This is useful in case we have regressions and we need to revert the change. Instead of reverting unnecessary code, we only revert the code that potentially causes issues.

GoGstickGo:
@christarazi  , 1.Changes break into commits.  2. Image validation for agent and operator can be found https://github.com/cilium/cilium-cli/pull/219/commits/37aba765ef648a0af6f6b5633ab7f79ff306aecf

christarazi (approved):

clustermesh/clustermesh.go
joestringer: Yep, that should work great. Thanks! :+1:

joestringer (approved):
Apologies about the delay getting back to this, I was out on PTO :-)  LGTM.

The Github actions on cloud environments seem to be unhappy, but I suspect that this is because you are submitting the change from outside the Cilium repository and we didn't yet support cloud CI logins from such PRs. Might be worth raising a question on Slack in #development or #testing to see what we should do about this.

twpayne (approved):

nbusseneau:
https://github.com/cilium/cilium-cli/actions/runs/927870806 hit flake fixed by #302:
```
📋 Test Report
❌ 1/9 tests failed (2/69 actions), 0 warnings, 0 tests skipped, 4 scenarios skipped:
Test [echo-ingress]:
  ❌ echo-ingress/pod-to-pod/curl-1: cilium-test/client2-657df6649d-zc7vr (10.20.1.152) -> cilium-test/echo-other-node-646f4f6797-ztgp6 (10.112.1.69:8080)
  ❌ echo-ingress/pod-to-pod/curl-2: cilium-test/client2-657df6649d-zc7vr (10.20.1.152) -> cilium-test/echo-other-node-646f4f6797-ztgp6 (10.112.1.69:8080)
```
I think we're good :)

GoGstickGo:
Thanks @nbusseneau ! 

@joestringer @christarazi, can we move forward with it ?

PR #225, opened 2021-05-17
sysdump: initial implementation

bmcustodio:
Adds a Go port of [`cilium-sysdump`](https://github.com/cilium/cilium-sysdump) as `cilium sysdump`. The implementation doesn't depend on `kubectl` or any other tool to be present.


internal/cli/cmd/sysdump.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/cilium/cilium-cli/sysdump"
+)
+
+var (
+	d bool
+	o = sysdump.SysdumpOptions{
+		LargeSysdumpAbortTimeout: sysdump.DefaultLargeSysdumpAbortTimeout,
+		LargeSysdumpThreshold:    sysdump.DefaultLargeSysdumpThreshold,
+	}
+	t time.Duration
```
rolinh, line 36: `d`, `o` and `t` are one letter vars that are global to the `cmd` package. It would be good to avoid this by, for example, creating one global struct for the sysdump command options.

internal/utils/utils.go
```diff
@@ -106,3 +106,12 @@ func (w *WaitObserver) Retry(err error) error {
 
 	return nil
 }
+
+func Contains(l []string, v string) bool {
+	for _, l := range l {
```
rolinh, line 111: Could we avoid variable shadowing please?

sysdump/sysdump.go
```diff
@@ -0,0 +1,825 @@
...
+	RESTClientGetter            genericclioptions.RESTClientGetter
+}
+
+type Sysdump interface {
+	Collect(context.Context) error
+}
+
+type sysdump struct {
+	k KubernetesClient
+	o SysdumpOptions
```
rolinh, line 66: I think that one letter variables are fine for scoped contexts, not so much as a struct member.

sysdump/sysdump.go
```diff
@@ -0,0 +1,825 @@
...
+	var merr error
+loop:
+	for {
+		select {
+		case <-wgdCh:
+			break loop
+		case err := <-errCh:
+			merr = multierror.Append(merr, err)
+		}
+	}
```
rolinh, line 590: Have you considered using [`cilium/workerpool`](https://github.com/cilium/workerpool) here? I think you could spare some lines of code and gain in readability here.

sysdump/sysdump.go
```diff
@@ -0,0 +1,825 @@
...
+					r = append(r, n.Name)
+					continue loop
+				}
+			}
+		}
+	}
+	return r, nil
+}
+
+func (c *sysdump) collectBugtool(ctx context.Context, l *log.Entry, pods []*corev1.Pod, containerName string, path func(string) string) error {
```
rolinh, line 646: I think the work being done as part of this method could also benefit from `cilium/workerpool`.

sysdump/sysdump.go
```diff
@@ -0,0 +1,825 @@
...
+	}
+	return nil
+}
+
+func buildNodeNameList(n *corev1.NodeList, filter string) ([]string, error) {
+	w := strings.Split(strings.TrimSpace(filter), ",")
+	r := make([]string, 0)
+loop:
+	// Iterate nodes and try to match either names or IPs to any of the specified filter strings.
+	for _, n := range n.Items {
```
rolinh, line 623: Could we maybe extract the inner loop logic into a dedicated function for better readability?

rolinh (changes requested):
That's awesome! I did a first pass, will likely do another one with testing later but here is already a couple of suggestions.

internal/utils/utils.go
bmcustodio: Sure, nice catch! :+1:

sysdump/sysdump.go
bmcustodio: Interesting, I wasn't aware of this! Let me 🔍!

bmcustodio: Done — though a `WaitGroup` is still in place to wait for all sub-tasks to be submitted before calling `Drain()`. Thank you for your help on this @rolinh. 💯

michi-covalent:
not sure why cluster creation is failing 🤔 , but probably makes sense to start by fixing these lint errors: https://github.com/cilium/cilium-cli/pull/225/checks?check_run_id=2601847837#step:6:8

internal/cli/cmd/sysdump.go
```diff
@@ -0,0 +1,71 @@
...
+			// Collect the sysdump.
+			if err := sysdump.NewSysdump(k8sClient, sysdumpOptions).Collect(); err != nil {
+				return fmt.Errorf("failed to collect sysdump: %v", err)
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&contextName, "context", "", "Kubernetes configuration context")
+	cmd.Flags().StringVar(&sysdumpOptions.CiliumLabelSelector, "cilium-label-selector", sysdump.DefaultCiliumLabelSelector, "The labels used to target Cilium pods")
```
glibsm, line 52: it's an interesting concept (using an intermediary struct to collect all the options together), but I find this whole section really messy and hard to read primarily to the length of each line.

A good layout for cobra options that I have found is pretty much using the usual golang line breaks with a few minor tweaks:

```
cmd.Flags().<TYPE>Var<P?>(&location.Var, // can be in-line since for reader it often doesn't matter
    "my-custom-var", defaults.DefaultValue, // name clearly visible with default
    "This is a custom var to do xyz", // description also clear on it's own line
)
.. other flags
```

This allows for easy section scan to see all the possible values (I think)

internal/k8s/client.go
```diff
@@ -29,39 +30,44 @@ import (
...
+	Config           *rest.Config
+	RawConfig        clientcmdapi.Config
+	restClientGetter genericclioptions.RESTClientGetter
+	contextName      string
 }
 
 func NewClient(contextName, kubeconfig string) (*Client, error) {
-	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+	// Register the Cilium types in the default scheme.
+	_ = ciliumv2.AddToScheme(scheme.Scheme)
```
glibsm, line 57: this function doesn't return an error by chance, does it?

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+func (s *sysdump) Collect() error {
+	// Grab the current timestamp and create a temporary directory to hold the files.
+	t := time.Now()
+
+	// replaceTimestamp can be used to replace the special timestamp placeholder in file and directory names.
+	replaceTimestamp := func(f string) string {
+		return strings.Replace(f, timestampPlaceholderFileName, t.Format(timeFormat), -1)
+	}
+
+	d, err := ioutil.TempDir("", "*")
```
glibsm, line 96: use `os.TempDir`

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+		return fmt.Errorf("failed to build node list: %w", err)
+	}
+	nodeList := l
+	s.logDebug("Restricting bugtool and logs collection to pods in %v", nodeList)
+
+	// tasks is the list of base tasks to be run.
+	tasks := []sysdumpTask{
+		// Get Kubernetes nodes.
+		func(ctx context.Context) error {
+			defer s.logDebug("Finished collecting Kubernetes nodes")
```
glibsm, line 146: defining the `sysdumpTask` as containing some sort of `desc string` and auto-logging it on completely would avoid these defers in each task

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+	// Build the list of node names in which the user is interested.
+	l, err := buildNodeNameList(n, s.options.NodeList)
+	if err != nil {
+		return fmt.Errorf("failed to build node list: %w", err)
+	}
+	nodeList := l
+	s.logDebug("Restricting bugtool and logs collection to pods in %v", nodeList)
+
+	// tasks is the list of base tasks to be run.
+	tasks := []sysdumpTask{
```
glibsm, line 143: This concept is pretty neat. one thing I'd suggest to do is instead of using in-lined anonymous functions is to define pure functions and then just collect them all here.

That would hopefully avoid having a several hundred lines long function (and also optionally allow for testing individual components)

Something like
```
tasks := []sysdumpTask{
    getVersion(s),
    collectEvents(s),
    ... etc
}
```

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+	LogsSinceTime               time.Duration
+	NodeList                    string
+	OutputFileName              string
+	Quick                       bool
+	RESTClientGetter            genericclioptions.RESTClientGetter
+	WorkerCount                 int32
+	Writer                      io.Writer
+}
+
+type Sysdump interface {
```
glibsm, line 63: not sure why this is an interface. Are we planning to have multiple sysdump implementations? (maybe they already exist and I haven't seen it yet)

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+	"github.com/cilium/workerpool"
+	"github.com/hashicorp/go-multierror"
+	archiver "github.com/mholt/archiver/v3"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+type SysdumpOptions struct {
```
glibsm, line 39: all the public structs (and their members :O) technically require comments to properly lint

sysdump/sysdump.go
```diff
@@ -0,0 +1,815 @@
...
+	// Try to remove the temporary directory.
+	s.logDebug("Removing the temporary directory %s", d)
+	if err := os.RemoveAll(d); err != nil {
+		s.logWarn("failed to remove temporary directory %s: %v", d, err)
+	}
+	return nil
+}
+
+func (s *sysdump) log(msg string, args ...interface{}) {
+	fmt.Fprintf(s.options.Writer, msg+"\n", args...)
```
glibsm, line 646: Timestamps may be nice here, but I'm not sure if we're intentionally keeping parity with previous output

glibsm (commented):
👏 awesome stuff. I tried to do a first pass with some comments without being too nitpicky.

Due to the nature of `rewrite X in Y` this is obviously a huge PR. We're going to have to balance the reviews carefully so we don't block this for too long. I think for me the main priority is to get things settled into a shape where we can easily jump in and iterate as needed.

michi-covalent:
> not sure why cluster creation is failing 🤔 , but probably makes sense to start by fixing these lint errors: https://github.com/cilium/cilium-cli/pull/225/checks?check_run_id=2601847837#step:6:8

https://github.com/cilium/cilium-cli/pull/212 should fix the cluster creation issue.

michi-covalent:
... for the short term it might be easier to just open a pull request from a branch in this repo instead of using a fork.

sysdump/sysdump.go
michi-covalent: https://github.com/cilium/cilium-cli/pull/229 should catch this.

internal/cli/cmd/sysdump.go
```diff
@@ -0,0 +1,71 @@
...
+		LargeSysdumpThreshold:    sysdump.DefaultLargeSysdumpThreshold,
+		Writer:                   sysdump.DefaultWriter,
+	}
+)
+
+func newCmdSysdump() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sysdump",
+		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
+		Long:  ``,
```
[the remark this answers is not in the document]
bmcustodio: I saw this being used in most sub-commands so I just followed the same pattern. `Short` seems to take over in this case, and I believe that to be enough:

```
➜  cilium-cli git:(bruno-cilium-sysdump) ✗ go run -race ./cmd/cilium/main.go sysdump --help
Collects information required to troubleshoot issues with Cilium and Hubble
(...)
```

WDYT?

internal/cli/cmd/sysdump.go
bmcustodio: This makes total sense :+1: I'm pushing the change shortly.

internal/k8s/client.go
bmcustodio: I've never seen it actually return an error — and the [usage examples](https://github.com/cilium/cilium/blob/v1.9.7/pkg/k8s/client/clientset/versioned/scheme/register.go#L45) seem to suggest that. I think it's OK.

sysdump/sysdump.go
bmcustodio: Hmmm I think [`os.TempDir`](https://golang.org/pkg/os/#TempDir) is a slightly different thing — it would just return, say, `/tmp` (or something weird in MacOS). We'd still have to create a directory there. [`ioutil.TempDir`](https://golang.org/pkg/io/ioutil/#TempDir) actually creates a directory inside the temporary directory, so it's one step instead of two. WDYT?

sysdump/sysdump.go
bmcustodio: I actually thought about this and I like the idea. I'll try to push the changes shortly.

sysdump/sysdump.go
bmcustodio: I think this makes sense too — I'll see how it plays with the suggestion above.

sysdump/sysdump.go
bmcustodio: We're using this pattern in [some](https://github.com/cilium/cilium-cli/blob/master/hubble/hubble.go#L42) [places](https://github.com/cilium/cilium-cli/blob/master/connectivity/check/check.go#L173), and I think that even if it doesn't add anything for now, it doesn't hurt either. WDYT?

sysdump/sysdump.go
bmcustodio: It's more of keeping parity with current output of the CLI — I stole this from somewhere else. However, I agree that logging should be made uniform throughout (of course, as part of a separate PR 😛).

sysdump/sysdump.go
bmcustodio: Makes sense :+1:

bmcustodio:
@michi-covalent I can't create a branch in this repo (and not in https://github.com/cilium/cilium) either 😢

sysdump/sysdump.go
bmcustodio: Done!

internal/cli/cmd/sysdump.go
bmcustodio: Done!

sysdump/sysdump.go
bmcustodio: Done!

sysdump/sysdump.go
rolinh: Note that link would also complain about the following:
```
type name will be used as sysdump.SysdumpOptions by other packages, and that stutters; consider calling this Options
```

sysdump/sysdump.go
rolinh: The whole `ioutil` package [is deprecated](https://golang.org/doc/go1.16#ioutil) since the release of Go v1.16. Instead of `ioutil.TempDir`, [`os.MkdirTemp`](https://golang.org/pkg/os/#MkdirTemp) should be used instead.

internal/cli/cmd/sysdump.go
```diff
@@ -0,0 +1,107 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Silence klog to avoid displaying "throttling" messages - those are expected.
+			klog.SetOutput(ioutil.Discard)
```
rolinh, line 42: Use [`io.Discard`](https://golang.org/pkg/io/#Discard) instead.

sysdump/writers.go
```diff
@@ -0,0 +1,57 @@
...
+	"io/ioutil"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/printers"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func writeBytes(p string, b []byte) error {
+	return ioutil.WriteFile(p, b, fileMode)
```
rolinh, line 28: Use [`os.WriteFile`](https://golang.org/pkg/os/#WriteFile) instead.

sysdump/sysdump.go
rolinh: I would only create the interface if it's strictly required, e.g. to implement tests.

sysdump/sysdump.go
```diff
@@ -0,0 +1,868 @@
...
+	NodeList string
+	// The name of the resulting file (without extension)\n'<ts>' can be used as the placeholder for the timestamp.
+	OutputFileName string
+	// Whether to enable quick mode (i.e. skip collection of 'cilium-bugtool' output and logs).
+	Quick bool
+	// A 'RESTClientGetter' that can be used to create REST clients for the Kubernetes API.
+	// Required at least for getting the proper output of 'kubectl get pod -o wide' without actually using 'kubectl'.
+	RESTClientGetter genericclioptions.RESTClientGetter
+	// The number of workers to use.
+	WorkerCount int32
```
rolinh, line 81: I'd stick to using `int` unless there's a good reason to use `int32`. I see that this member is being cast to an `int` when used anyway.

rolinh (approved):
I believe there's room for improvements but given that this PR is an entire rewrite of `cilium-sysdump` and that all existing functionality seems to have been re-implemented, I'm in favor or merging it as is and improve the code with follow-up PRs. The command could possibly be marked `hidden` perhaps before the work on the command is considered finalized.

PR #229, opened 2021-05-18
Use golangci-lint for linting

michi-covalent:
Using `revive` instead of `golint` based on this warning:

    level=warning msg="[runner] The linter 'golint' is deprecated
    (since v1.41.0) due to: The repository of the linter has been
    archived by the owner.  Replaced by revive."

Ref: https://github.com/golangci/golangci-lint


rolinh (changes requested):
Since `lint` is no longer maintained, would you please remove the `lint` target from the `Makefile` as well and the dep itself from `/tools`?

As `golangci-lint` seems to be a replacement for all of `ineffasign`, `lint`, `staticcheck`, `gofmt` and `vet` given the right set of options, would you mind pruning these targets from the `Makefile` as well (including the vendored deps) and add a new make target for `golangci-lint`?

michi-covalent:
> Since `lint` is no longer maintained, would you please remove the `lint` target from the `Makefile` as well and the dep itself from `/tools`?
> 
> As `golangci-lint` seems to be a replacement for all of `ineffasign`, `lint`, `staticcheck`, `gofmt` and `vet` given the right set of options, would you mind pruning these targets from the `Makefile` as well (including the vendored deps) and add a new make target for `golangci-lint`?

sounds good! i wanted to get feedback from people before starting to remove all the existing static checks.

Makefile
```diff
@@ -68,23 +68,7 @@ test:
 bench:
 	go test -timeout=30s -bench=. $$(go list ./...)
 
-check: gofmt ineffassign lint staticcheck vet
+check:
+	golangci-lint run -v
```
michi-covalent, line 72: @rolinh how do you feel about having to install `golangci-lint` binary? running this inside docker was too slow, and they don't recommend building from source 🤔

rolinh: Not really fond of it but on the other hand, I don't manage to install it from source because of that same old grpc breaking change issue that breaks the etcd client :confused:  :
```
go: finding module for package github.com/golangci/golangci-lint/cmd/golangci-lint
go: found github.com/golangci/golangci-lint/cmd/golangci-lint in github.com/golangci/golangci-lint v1.40.1
go: finding module for package google.golang.org/grpc/naming
go: finding module for package google.golang.org/grpc/examples/helloworld/helloworld
go: found google.golang.org/grpc/examples/helloworld/helloworld in google.golang.org/grpc/examples v0.0.0-20210519181852-3dd75a6888ce
go: finding module for package google.golang.org/grpc/naming
github.com/cilium/cilium-cli/internal/certs imports
        github.com/cloudflare/cfssl/helpers imports
        github.com/google/certificate-transparency-go imports
        go.etcd.io/etcd imports
        go.etcd.io/etcd/etcdmain imports
        go.etcd.io/etcd/proxy/grpcproxy imports
        google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.38.0), but does not contain package google.golang.org/grpc/naming
```
Another option would be to run it as a docker container, which I think I prefer because this means control over the version that is run.

```
docker run --rm -v $(pwd):/app -w /app docker.io/golangci/golangci-lint:v1.40.1 golangci-lint run -v
```

michi-covalent: > Another option would be to run it as a docker container, which I think I prefer because this means control over the version that is run.
> 
> ```
> docker run --rm -v $(pwd):/app -w /app docker.io/golangci/golangci-lint:v1.40.1 golangci-lint run -v
> ```

yeah i tried that as well, it's much slower (takes more than 1m each time i run the command)  😐

ti-mo: :wave: Switching to golangci-lint would be great, was thinking of proposing the same change.

I see you're using a pre-built golangci-lint GH action here, but on most other CIs it's usually installed as a preparation step as described here: https://golangci-lint.run/usage/install.
```
~ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.40.1
```

This pins the version of the golangci-lint binary, so whatever runs in CI will always be checked against the same version just to make sure.

Just curious, what's the issue with installing the binary locally? The script above can be used, personally I just install it from AUR, and it's also available on Homebrew for mac users. Never had any trouble with that.

michi-covalent: i just went with their recommendation but no strong preference so long as we get those nice annotations for lint errors:

> We recommend using our GitHub Action for running golangci-lint in CI for GitHub projects

personally i don't have strong preference regarding installing the binary locally either. one benefit of using docker is that it's easier to control which version everybody uses, but it is super slow on mac 😐

michi-covalent: updated the pr. in summary:

- still using golangci-lint github actions based on their recommendation.
- `make check` runs golangci-lint in a docker container.
- updated the commit message with an alternative way to run`golangci-lint` with a locally installed binary.

.golangci.yml
```diff
@@ -0,0 +1,10 @@
+# See https://golangci-lint.run/usage/configuration/ for available options.
+linters:
+  enable:
```
ti-mo, line 3: For this list, we could draw inspiration from the [Cilium .golangci.yaml file](https://github.com/cilium/cilium/blob/master/.golangci.yaml#L82-L92), although I would add `goimports` and `revive`.

michi-covalent: thanks for the pointer! updated the list. now running with:
```
INFO [lintersdb] Active 9 linters: [goerr113 goimports govet ineffassign misspell revive staticcheck unused varcheck]
```

christarazi (approved):

Makefile
rolinh: > Just curious, what's the issue with installing the binary locally?

First, it requires to install the binary. Second, to keep it updated. Third, if the version doesn't match with the version in the repo, discrepancies along what is being reported will likely emerge. For overall consistency, I think it's best to stick to a single version.

Makefile
```diff
@@ -68,23 +68,7 @@ test:
 bench:
 	go test -timeout=30s -bench=. $$(go list ./...)
 
-check: gofmt ineffassign lint staticcheck vet
+check:
+	docker run --rm -v `pwd`:/app -w /app docker.io/golangci/golangci-lint:v1.40.1 golangci-lint run -v
```
rolinh, line 72: How does it pick the same list of plugins as the ones configured in the CI job?

ti-mo: By mounting pwd onto `/app` and setting `/app` as cwd, it should use the ` .golangci.yml` file added in this PR.

rolinh (approved):

tklauser (approved):
:rocket:

Makefile
ti-mo: 1, 2.: AUR/brew have the package, so it's set and forget. Keeping the *repo* version up to date would be more of a chore than a local workstation. :sweat_smile: 
3: Yes, very annoying, but whatever linter errors you get running locally should be indicative anyway. As long as CI has the final word, we're not at risk of breaking the build.

Haven't hit 3. in a while (not since using golangci-lint) since it acts as a buffer. New releases usually appear around Go releases. As long as `.golangci.yaml` is in the repo, my editor will be happy. :shrug: If the Makefile target runs whatever CI runs, it's good as a final check.

ti-mo (approved):

twpayne (approved):

PR #236, opened 2021-05-21
sysdump: support collecting eks-specific data

bmcustodio:
Adds support for collecting EKS-specific data — namely the `kube-system/aws-node` `DaemonSet`, if present, and any `eniconfigs` and `securitygrouppolicies` that may be present as well. Useful when Cilium is running in chaining mode. Also includes some cleaning-up bits.

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -33,9 +33,12 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -47,6 +50,7 @@ import (
 
 type Client struct {
 	Clientset        kubernetes.Interface
+	DynamicClientset dynamic.Interface
 	CiliumClientset  ciliumClientset.Interface
 	Config           *rest.Config
 	RawConfig        clientcmdapi.Config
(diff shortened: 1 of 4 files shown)


internal/k8s/client.go
```diff
@@ -576,6 +586,13 @@ func (c *Client) GetPodsTable(ctx context.Context) (*metav1.Table, error) {
 	return unstructuredToTable(i[0].Object)
 }
 
+func (c *Client) ListUnstructured(ctx context.Context, gvr schema.GroupVersionResource, namespace *string, o metav1.ListOptions) (*unstructured.UnstructuredList, error) {
```
errordeveloper, line 589: Good job on figuring this one out! These things are a bit of a maze to navigate. I found that [`sigs.k8s.io/controller-runtime/pkg/client`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client) is kind of best way to go about types, it accepts typed resources as well as `Unstructured`.

nathanjsweet (approved):
Nice. Thanks for the patch!

sysdump/sysdump.go
```diff
@@ -824,6 +842,51 @@ func (c *Collector) submitLogsTasks(ctx context.Context, pods []*corev1.Pod, sin
...
 
+func (c *Collector) submitFlavorSpecificTasks(ctx context.Context, f k8s.Flavor, path func(string) string) error {
+	switch f.Kind {
+	case k8s.KindEKS:
+		if err := c.pool.Submit(awsNodeDaemonSetName, func(ctx context.Context) error {
+			// Collect the 'kube-system/aws-node' DaemonSet.
+			d, err := c.client.GetDaemonSet(ctx, awsNodeDaemonSetNamespace, awsNodeDaemonSetName, metav1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					c.logWarn("DaemonSet %q not found in namespace %q - this is expected when running in ENI mode", awsNodeDaemonSetName, awsNodeDaemonSetNamespace)
```
aanm, line 853: If it is expected why are we telling that to the user in a form of a warning message? Maybe we could not print the message at all

aanm (changes requested):

sysdump/sysdump.go
bmcustodio: Makes sense — I've changed it to `logDebug` as it may still be useful. WDYT?

aanm (approved):

bmcustodio:
@tklauser rebased this one too 👍🏼

tklauser:
AKS job is failing because this PR was opened from a fork and is thus lacking Azure credentials. Merging, sorry for the delay @bmcustodio.

Issue #239, opened 2021-05-23
install would take production down


lfundaro:
If `cilium install` is meant to behave well with production clusters, then at the end of the installation, when restarting unmanaged pods, it should  issue pod deletions through the eviction API (honoring Pod Disruption Budgets), otherwise plain deletes would cause a service outage.

PR #240, opened 2021-05-23
Evict pods at the end of install command

lfundaro:
This PR tries to make the install command more
gentle with a production cluster. Instead of deleting
all unmanaged pods, we try to evict them, which essentially
achieves a pod deletion but it does so by honoring Pod Disruption
Budgets. This way deletions won't cause service outages.

Fixes: #239

Signed-off-by: Lorenzo Fundaró <lorenzofundaro@gmail.com>


lfundaro:
⚠️  No tests yet, waiting for Authors input first.

install/install.go
```diff
@@ -988,6 +991,7 @@ type k8sInstallerImplementation interface {
 	GetNamespace(ctx context.Context, namespace string, options metav1.GetOptions) (*corev1.Namespace, error)
 	ListPods(ctx context.Context, namespace string, options metav1.ListOptions) (*corev1.PodList, error)
 	DeletePod(ctx context.Context, namespace, name string, options metav1.DeleteOptions) error
```
kaworu, line 993: Is `DeletePod` not needed anymore after this patch?

kaworu (commented):
Hi @lfundaro and thanks for the PR! Small question but the idea make sense to me. cc @aanm

install/install.go
lfundaro: I didn't check if `DeletePod` was referenced in other parts of the code. If it's not, it could make sense to remove that.

aanm: Yeah it doesn't look like it's used anywhere else so we can remove it.

install/install.go
```diff
@@ -1445,27 +1447,71 @@ func (k *K8sInstaller) restartUnmanagedPods(ctx context.Context) error {
...
-			}
-			err := k.client.DeletePod(ctx, pod.Namespace, pod.Name, metav1.DeleteOptions{})
-			if err != nil {
-				k.Log("⚠️  Unable to restart pod %s/%s: %s", pod.Namespace, pod.Name, err)
-			} else {
+	// group pods by namespace to make evictions faster
+	nsToPods := make(map[string][]corev1.Pod)
+	for _, p := range pods.Items {
+		nsToPods[p.Namespace] = append(nsToPods[p.Namespace], p)
+	}
```
aanm, line 1463: I don't think we need to create a map with a copy of all pods in here again. We could simply store the state that we care about which is `hostNetwork`, `name`, `namespace` and `TerminationGracePeriodSeconds`.

aanm (changes requested):

install/install.go
lfundaro: just pushed a new commit with the function deleted

install/install.go
lfundaro: I think I should explain the motivation behind. Pod Disruption Budgets (PDB) are namespaced and applied to pods that satisfy a given selector. When you try to evict more pods than the PDB allows, you need to wait some time and retry.  
Given the following scenario:
```
NS1:
   pod1
   pod2
   pod3

PDB_1 -> selects pods from NS1  (maxUnavailable = 1, aka 1 disruption at a time)

NS2:
   pod1
   pod2
```

We have two namespaces and one has a PDB defined. When getting the list of pods you get something like:
```
pod.Items = [ pod1_ns1, pod2_ns1, pod3_ns1, pod1_ns2, pod2_ns2 ]
```
As you iterate over `pod.Items`, assuming pods on `ns1` take long time to terminate, you delete pod1 and immediately right after try to delete pod2, but you would get throttled by PDB_1 and need to wait some time X until you can keep with the loop. You'll get again throttled when trying to evict pod3. With this method, you haven even started to delete pods from ns2, which may or may not have a PDB. So, for the sake of making progress in this loop, and knowing how PDBs work, we should group pods by namespace (and maybe also by selector) so make evictions in the overall cluster faster. 

**Note:** I just realized I took the simplification of just grouping by namespace but in reality we should group by namespace and by selector. Otherwise, this will not work if you have all your cluster pods in a single namespace (e.g.: `default`) where you can actually have multiple PDBs that select different pods within the same namespace

aanm: I'm fine with the motivation of grouping pods per namespace. My feedback is regarding the unnecessary memory consumption of duplicating a full pod structure while we can simply copy the fields that we care about.

lfundaro: Oh ok, sorry for the confusion, I thought you were questioning the need for grouping in general. I think we can do that. Only that, if we copy those fields, we have to put them into a struct and I would prefer that we don't introduce throw-away structs just to keep things simple and easy to read. What about we take a pointer to those pods ? 

```
// current
var map[string][]corev1.Pod
// proposed
var map[string][]*corev1.Pod
```

wdyt ?

aanm: yeah pointers work as fine

lfundaro: I would like to understand why you changed your mind with pointers. I think the intent on `ceMap` is to create a `set`  , whereas here, we're trying to do a `groupBy` operation with a subset of the pods values. Two different purposes. Are we discussing a matter of style here ? I'm confused because we started with the argument that `memory` is a problem and we end up now with a code that is more verbose (need to add local struct) and more memory (copying values that already exist in memory).  

> while instead of struct{}{} we would have hostNetwork and TerminationGracePeriodSeconds

we would also need to add the pod's Name.

lfundaro: I pushed two commits:

[ed13f5a](https://github.com/cilium/cilium-cli/pull/240/commits/ed13f5a370f51511a6675db1df95499aa777e5ec): using pointers
[7e6f72a](https://github.com/cilium/cilium-cli/pull/240/commits/7e6f72a1a6967b899ba1db8ba34ab57145f6aca2): using structs

Maybe it would be more evident to see the differences between the two approaches. Cheers !

aanm: Sorry, I completely miss the point of having them grouped by namespace as I was reviewing the PR thinking that we can't perform a podList for the entire cluster. However that can be fixed in a follow up PR. Using pointers for now is fine.

lfundaro: I just reverted the last commit that uses a struct and now we are back to using pointers. Cheers

aanm (approved):

kaworu:
@lfundaro thanks! The PR LGTM.

> warning No tests yet, waiting for Authors input first.

I think we can start address testing now. I guess it would help if the logic was split into a dedicated function (for readability if not for testing).

lfundaro:
> @lfundaro thanks! The PR LGTM.
> 
> > warning No tests yet, waiting for Authors input first.
> 
> I think we can start address testing now. I guess it would help if the logic was split into a dedicated function (for readability if not for testing).

Hi @kaworu ! I refactored the code and split functions. Also, I started testing some of the newly introduced functions. I'm delaying the test of `runEvictions` as it's a bit more involved and thought I might discuss with you before. For that function I see two ways of testing: 

1. Mocking the eviction API. A bit involved I think.
2. Not testing the function _per se_ but rather use CI to run a `cilium install` and verify that restarting pods honors PDBs.

Those are the ones off the top of my head. What do you think ? 

btw, I'm gonna be off the radar for a week, but as soon as I'm back, I can pick up any comments left. 

Cheers !

kaworu:
@lfundaro yes using CI make sense to me. /cc @cilium/ci-structure

christarazi (commented):
Please squash all commits into the first

errordeveloper:
Quick high-level question before I look at the code, was this an optional behaviour already, or was pod deletion done unconditionally?

lfundaro:
> Quick high-level question before I look at the code, was this an optional behaviour already, or was pod deletion done unconditionally?

`cilium install --restart-unmanaged-pods` (default true)

install/install.go
```diff
@@ -1416,15 +1419,83 @@ func (k *K8sInstaller) deployResourceQuotas(ctx context.Context) error {
...
+				}
+
+				err := k.client.EvictPod(ctx, pod.Namespace, pod.Name, metav1.DeleteOptions{})
+				if err != nil {
+					if errors.IsTooManyRequests(err) { //PodDisruptionBudget enforcement
+						waitTime := time.Duration(*pod.Spec.TerminationGracePeriodSeconds)
+						k.Log("ℹ️  Restarting %s/%s would violate PodDisruptionBudget. Sleeping %ds before retrying.", pod.Namespace, pod.Name, waitTime)
+						time.Sleep(waitTime * time.Second)
+						retry++
+						if retry > maxRetry {
```
twpayne, line 1470: The retry count check here can happen before sleeping and logging.

install/install.go
```diff
@@ -1416,15 +1419,83 @@ func (k *K8sInstaller) deployResourceQuotas(ctx context.Context) error {
...
+					} else if errors.IsInternalError(err) {
+						k.Log("⚠️  Could not restart pod %s/%s. Please take manual action: %s", pod.Namespace, pod.Name, err)
+					} else {
+						k.Log("⚠️  Unable to restart pod %s/%s: %s.", pod.Namespace, pod.Name, err)
+					}
+					pods = tail(pods)
+					continue
+				}
+				k.Log("♻️  Restarted unmanaged pod %s/%s", pod.Namespace, pod.Name)
+				retry = 0
```
twpayne, line 1484: This is the only place where `retry` is reset to `0`. Shouldn't other calls to `pods = tail(pods); continue` also reset `retry`?

install/install.go
```diff
@@ -1416,15 +1419,83 @@ func (k *K8sInstaller) deployResourceQuotas(ctx context.Context) error {
...
+	for ns, pods := range nsToPods {
+		ns := ns
+		pods := pods
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			k.Log("♻️  Restarting unmanaged pods on namespace=%s", ns)
+			var retry int
+			maxRetry := 5
+			for len(pods) > 0 {
```
twpayne, line 1451: I think this loop can be written more clearly as a standard `for` ... `range` loop rather than treating `pods` as a list and replacing it with its tail.

twpayne (changes requested):

internal/cli/cmd/install.go
```diff
@@ -67,8 +67,8 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringVar(&params.InheritCA, "inherit-ca", "", "Inherit/import CA from another cluster")
 	cmd.Flags().StringVar(&params.KubeProxyReplacement, "kube-proxy-replacement", "probe", "Enable/disable kube-proxy replacement { disabled | probe | strict }")
 	cmd.Flags().BoolVar(&params.Wait, "wait", true, "Wait for status to report success (no errors)")
-	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
-	cmd.Flags().BoolVar(&params.RestartUnmanagedPods, "restart-unmanaged-pods", true, "Restart pods which are not being managed by Cilium")
+	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time (minutes) to wait for status")
+	cmd.Flags().BoolVar(&params.RestartUnmanagedPods, "restart-unmanaged-pods", true, "Restart pods which are not being managed by Cilium. It honors PodDisruptionBudgets")
```
errordeveloper, line 71: It's not not an actual sentence, let's not introduce punctuation as it if was one. How about this:
```suggestion
	cmd.Flags().BoolVar(&params.RestartUnmanagedPods, "restart-unmanaged-pods", true, "Evict all pods that are not being managed by Cilium"
```

install/install_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	nsToPods := groupPodsByNamespace(&t)
+	c.Assert(nsToPods["Berlin"], check.DeepEquals, []*corev1.Pod{{ObjectMeta: v1.ObjectMeta{Name: "Foo", Namespace: "Berlin"}},
+		{ObjectMeta: v1.ObjectMeta{Name: "Bar", Namespace: "Berlin"}}})
+	c.Assert(nsToPods["Caracas"], check.DeepEquals, []*corev1.Pod{{ObjectMeta: v1.ObjectMeta{Name: "Qux", Namespace: "Caracas"}}})
+}
+
+func (b *InstallSuite) TestRunEvictions(c *check.C) {
+	// it should not evict cep pods
```
errordeveloper, line 59: Are you planning on implementing this?

install/install_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func (b *InstallSuite) TestTail(c *check.C) {
+	cases := []struct {
+		given  []*corev1.Pod
+		expect []*corev1.Pod
+	}{
+		{
+			given: []*corev1.Pod{
+				{ObjectMeta: v1.ObjectMeta{Name: "Foo"}}},
```
errordeveloper, line 16: Why use capitals here? I find this a bit disorienting, please use all-lower-case.

install/install.go
```diff
@@ -1416,15 +1419,83 @@ func (k *K8sInstaller) deployResourceQuotas(ctx context.Context) error {
...
+
+				err := k.client.EvictPod(ctx, pod.Namespace, pod.Name, metav1.DeleteOptions{})
+				if err != nil {
+					if errors.IsTooManyRequests(err) { //PodDisruptionBudget enforcement
+						waitTime := time.Duration(*pod.Spec.TerminationGracePeriodSeconds)
+						k.Log("ℹ️  Restarting %s/%s would violate PodDisruptionBudget. Sleeping %ds before retrying.", pod.Namespace, pod.Name, waitTime)
+						time.Sleep(waitTime * time.Second)
+						retry++
+						if retry > maxRetry {
+							k.Log("⚠️  Unable to restart pods in namespace=%s, after retries=%d every t=%ds.", pod.Namespace, maxRetry, waitTime)
```
errordeveloper, line 1471: Since other status message are not using the variable notation, it'd be more consistent to avoid it here also, e.g.:
```suggestion
							k.Log("⚠️  Unable to restart pods in namespace %q, after %d retries every %ds.", pod.Namespace, maxRetry, waitTime)
```

install/install.go
lfundaro: Hi @twpayne , thanks for your feedback. I understand that it may seem an exotic way to do a `for` ... `range` loop, but I'm doing it this way because of the nature of the eviction process. The slice we are looping has a set of pods that belong to the same namespace and _likely_ have the same Pod Disruption Budget in common. On a given iteration, you take one pod and it may fail because of PDB contraints, in that case it's of no use to go to the next pod because you're just evicting another replica of the same pod family and you'll be hit again by the PDB contraints. That's why I came up with a stack where we loop from the top and only pop that element (tail) as soon as we make progress on the eviction.  I reckon we could make the code clearer to convey the fact that we are using a stack here.

If I use a `for` ... `range` loop the code becomes confusing and possibly unsafe(?) because for each iteration I may need to re-introduce the same element to ranged over it again. I think that breaks the purpose that a range loop is intended for. 

What do you think ?

install/install_test.go
lfundaro: I will remove this. I discussed this with @kaworu , and it seems that this kind of test can be done on the CI, e.g.: asserting that evictions happen as expected. This will make the whole thing easier, otherwise we would need to mock too many things for the value that we get.

install/install.go
lfundaro: I think you have a point, but I need to reason about it more thoroughly. Let me get back to you on this one. Thanks !

install/install.go
errordeveloper: > I reckon we could make the code clearer to convey the fact that we are using a stack here.

I think that would be quite helpful indeed.

tklauser:
The build for this PR is failing and there seem to be review comments which haven't been addressed yet. I'll thus convert the PR to draft status. Please mark it as ready for review once everything is addressed.

install/install.go
lfundaro: Hi @errordeveloper  and @twpayne . I refactor the code and simplify the eviction logic. Instead of grouping pods by namespace I just fire concurrent go routines that would try (and retry) to evict a single pod each. IMHO the code is easier to read and understand. Let me know what you think. Cheers !

lfundaro: btw, I will fix the conflicts later.

PR #244, opened 2021-05-26
.github: Stop running tests for every push to master

pchaigno:
This pull request changes the triggers of all tests that spin up clusters to only run on pull requests and every 6h. That matches what is done at https://github.com/cilium/cilium and probably enough to catch potential regressions.

We are limited to 20 concurrent GitHub jobs for the entire Cilium organization so we should avoid running long workflows unless strictly necessary.


ti-mo:
Thanks, was wondering if anything could be done about this. :+1:

tklauser (approved):
Thanks!

aanm (approved):

PR #251, opened 2021-05-27
connectivity: egress HTTP and policy w/ toFQDNs

ti-mo:
Best reviewed per commit.

- Added test `client-egress-l7` for HTTP filtering with and without `toFQDNs`. Only client Pod 'other' is allowed to generate HTTP traffic.
- Some fixups to the yaml parser and error reporting.
- Run PodToWorld calls to google.com from all client Pods.
- Execute `connectivity.Run()` inside its own goroutine.

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -195,21 +195,25 @@ func (a *Action) ExecInPod(ctx context.Context, cmd []string) {
 	cmdName := cmd[0]
 	cmdStr := strings.Join(cmd, " ")
 
-	if stderr.Len() > 0 {
-		a.test.Debugf("%s stderr: %s", cmdName, stderr.String())
-	} else if stdout.Len() > 0 {
-		a.test.Debugf("%s stdout: %s", cmdName, stdout.String())
-	}
-
 	if err != nil || stderr.Len() > 0 {
 		if a.shouldSucceed() {
+			// Command failed unexpectedly, display stderr.
 			a.Failf("command %q failed: %s", cmdStr, err)
+			a.test.Infof("%s stderr:", cmdName)
+			a.test.Log(strings.TrimSpace(stderr.String()))
+			a.test.Log()
 		} else {
 			a.test.Debugf("command %q failed as expected: %s", cmdStr, err)
 		}
 	} else {
 		if !a.shouldSucceed() {
+			// Command succeeded unexpectedly, display stdout. (stderr should be empty)
 			a.Failf("command %q succeeded while it should have failed: %s", cmdStr, stdout.String())
+			if stdout.Len() > 0 {
+				a.test.Infof("%s stdout:", cmdName)
+				a.test.Log(strings.TrimSpace(stdout.String()))
+				a.test.Log()
+			}
 		}
 	}
 }
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -49,6 +49,9 @@ var (
 
 	//go:embed manifests/client-egress-l7-http.yaml
 	clientEgressL7HTTPPolicyYAML string
+
+	//go:embed manifests/echo-ingress-l7-http.yaml
+	echoIngressL7HTTPPolicyYAML string
 )
 
 func Run(ctx context.Context, ct *check.ConnectivityTest) error {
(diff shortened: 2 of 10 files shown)


internal/cli/cmd/connectivity.go
```diff
@@ -87,9 +90,29 @@ func newCmdConnectivityTest() *cobra.Command {
 				cancel()
 			}()
 
-			if err := connectivity.Run(ctx, cc); err != nil {
-				fatalf("Connectivity test failed: %s", err)
+			done := make(chan bool)
```
christarazi, line 93: This doesn't need to be `chan bool` and instead can be a "signal" channel, aka `chan struct{}`

connectivity/suite.go
```diff
@@ -166,6 +169,29 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 			return check.ResultOK, check.ResultNone
 		})
 
+	// Test L7 HTTP introspection abilities and egress filtering.
+	// The policy allows ambient DNS resolution,
+	ct.NewTest("client-egress-l7").
```
christarazi, line 174: Can we also add `echo-ingress-l7` as was pointed out in #217?

connectivity/check/action.go
```diff
@@ -161,21 +161,21 @@ func (a *Action) ExecInPod(ctx context.Context, cmd []string) {
...
-		a.test.Debugf("%s stderr: %s", cmdName, stderr.String())
-	} else if stdout.Len() > 0 {
-		a.test.Debugf("%s stdout: %s", cmdName, stdout.String())
-	}
-
 	if err != nil || stderr.Len() > 0 {
 		if a.shouldSucceed() {
+			// Command failed unexpectedly, display stderr.
 			a.Failf("command %q failed: %s", cmdStr, err)
+			a.test.Infof("%s stderr:\n%s\n", cmdName, strings.TrimSpace(stderr.String()))
```
christarazi, line 168: In general, I personally prefer to not include newlines in formatted output, but instead just issue another `Infof`. It makes it easier to read the string being printed from the code point of view. WDYT?

christarazi (changes requested):
LGTM to me overall, I left a few comments to be addressed

connectivity/check/action.go
ti-mo: Oh, totally! stdout/err is usually multi-line, so would probably use a `Log()` for the second call. Thanks!

connectivity/suite.go
ti-mo: Okay, sure. Does it make a difference in practice whether an L7 policy is applied using an `endpointSelector` or a `toEndpoints` clause? We should try to avoid redundant tests, but I'm not opposed to having a separate L7 ingress test.

internal/cli/cmd/connectivity.go
ti-mo: Sure, that eliminates a single alloc. :stuck_out_tongue: Will fix, thanks.

connectivity/suite.go
ti-mo: Okay, reading through https://github.com/cilium/cilium/issues/15826, it looks like there is indeed a difference. Will add an ingress equivalent as well. :+1:

ti-mo:
Converted to draft as it's currently blocked on https://github.com/cilium/cilium/issues/16376 and L7 ingress policy test needs to be added.

connectivity/suite.go
ti-mo: Added `echo-ingress-l7` test.

christarazi (approved):

internal/cli/cmd/connectivity.go
```diff
@@ -77,16 +80,34 @@ func newCmdConnectivityTest() *cobra.Command {
...
 
-			if err := connectivity.Run(ctx, cc); err != nil {
-				fatalf("Connectivity test failed: %s", err)
+			done := make(chan struct{})
+			var finished bool
+
+			// Execute connectivity.Run() in its own goroutine, it might call Fatal()
+			// and end the goroutine without returning.
+			go func() {
+				defer func() { done <- struct{}{} }()
```
twpayne, line 96: nit: `close(done)` allows multiple goroutines to block on `done`. Not needed here of course.

twpayne (approved):
Very nice!

PR #227, opened 2021-05-18
install: allow to set Azure subscription ID

tklauser:
In some cases, e.g. when working with the `cilium-dev` subscription,
leading, Cilium installation fails with the derived subscription ID:
    
```
$ cilium install --version 1.9.5 --azure-resource-group "${NAME}" --azure-tenant-id $(echo ${AZURE_SERVICE_PRINCIPAL} | jq -r '.tenant') --azure-client-id "${APP_ID}" --azure-client-secret $(echo ${AZURE_SERVICE_PRINCIPAL} | jq -r '.password')
🔮 Auto-detected Kubernetes kind: AKS
✨ Running "AKS" validation checks
✅ Detected az binary
🔮 Auto-detected cluster name: tklauser-test-12960
🔮 Auto-detected IPAM mode: azure
🔮 Auto-detected datapath mode: azure
✅ Using manually configured principal for cilium operator with App ID <redacted> and tenant ID <redacted>
✅ Derived Azure subscription id <redacted>
Error: Unable to install Cilium: unable to execute "az aks show --resource-group foo-test-12960 --name dwifootklauser-test-12960": exit status 3   
```
    
Fix this by allowing to override the subscription ID using the
`--azure-subscription-id` command line flag and pass its value to the
`az aks show` command. Keep using the derived subscription ID for
`AZURE_SUBSCRIPTION_ID` though.

PR #252, opened 2021-05-27
azure: refactor subscription handling + small improvements

nbusseneau:
Please see individual commits.

This is a follow-up to #227.

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -78,10 +78,10 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringVar(&params.OperatorImage, "operator-image", "", "Image path to use for Cilium operator")
 
 	cmd.Flags().StringVar(&params.Azure.ResourceGroupName, "azure-resource-group", "", "Azure resource group name the cluster is in")
-	cmd.Flags().StringVar(&params.Azure.TenantID, "azure-tenant-id", "", "Azure tenant ID")
-	cmd.Flags().StringVar(&params.Azure.ClientID, "azure-client-id", "", "Azure client (application) ID")
+	cmd.Flags().StringVar(&params.Azure.TenantID, "azure-tenant-id", "", "Tenant ID of Azure Service Principal to use for installing Cilium (will create one if none provided)")
+	cmd.Flags().StringVar(&params.Azure.ClientID, "azure-client-id", "", "Client (application) ID of Azure Service Principal to use for installing Cilium (will create one if none provided)")
 	cmd.Flags().StringVar(&params.Azure.SubscriptionID, "azure-subscription-id", "", "Azure subscription ID")
-	cmd.Flags().StringVar(&params.Azure.ClientSecret, "azure-client-secret", "", "Azure client secret")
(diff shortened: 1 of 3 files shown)


nbusseneau:
I've fixed this warning:
image

But I've no idea why it complains about this when there are multiple identical error returns in there:
image

tklauser:
> But I've no idea why it complains about this when there are multiple identical error returns in there:
> image

I guess this is because it is newly added and the linter only checks changed code. To me, it doesn't really make sense to define a static sentinel error `var` in this case as nothing will check for that error exactly. Maybe we could somehow disable that check?

tklauser (approved):
Thanks a lot for cleaning up after me and implementing this properly :)

LGTM once we figured out a way to please/silence the `goerr113` checker.

ti-mo:
@nbusseneau See https://github.com/cilium/cilium-cli/pull/253, added an exception there to allow dynamic errors.

michi-covalent (approved):

internal/cli/cmd/install.go
```diff
@@ -77,11 +77,11 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringVar(&params.AgentImage, "agent-image", "", "Image path to use for Cilium agent")
 	cmd.Flags().StringVar(&params.OperatorImage, "operator-image", "", "Image path to use for Cilium operator")
 
-	cmd.Flags().StringVar(&params.Azure.ResourceGroupName, "azure-resource-group", "", "Azure resource group name the cluster is in")
-	cmd.Flags().StringVar(&params.Azure.TenantID, "azure-tenant-id", "", "Azure tenant ID")
-	cmd.Flags().StringVar(&params.Azure.ClientID, "azure-client-id", "", "Azure client (application) ID")
-	cmd.Flags().StringVar(&params.Azure.SubscriptionID, "azure-subscription-id", "", "Azure subscription ID")
-	cmd.Flags().StringVar(&params.Azure.ClientSecret, "azure-client-secret", "", "Azure client secret")
+	cmd.Flags().StringVar(&params.Azure.ResourceGroupName, "azure-resource-group", "", "Azure resource group name the cluster is in (required)")
+	cmd.Flags().StringVar(&params.Azure.SubscriptionName, "azure-subscription", "", "Azure subscription name the cluster is in (default `az account show`)")
```
ti-mo, line 81: The parameter should probably be called `"azure-subscription-name"`, since it's not clear if this now takes a name or an ID.

ti-mo (changes requested):
Thanks, this should remove some friction. One small nit.

Is there documentation we need to update after this change? We've started including cilium-cli in GSGs with the 1.10 release (although we're still installing 1.9..)

internal/cli/cmd/install.go
nbusseneau: I decided to leave out `name` to be consistent with how `--azure-resource-group` is (does not have the name either, while the viarable does).

nbusseneau:
> Is there documentation we need to update after this change? We've started including cilium-cli in GSGs with the 1.10 release (although we're still installing 1.9..)

I don't think so: the current GSG guide uses the simplest CLI forms, and so will still use the default subscription (no changes).

https://docs.cilium.io/en/v1.10/gettingstarted/k8s-install-default/

internal/cli/cmd/install.go
ti-mo: True! Didn't catch that, fine by me. :+1:

PR #253, opened 2021-05-27
Address all outstanding linter errors

ti-mo:
It came to my attention that `golangci-lint` was enabled in 'new' mode, which ignores all existing linter errors present in the codebase. This PR addresses that issue.

Also allowed exported consts without docstrings, as well as dynamic errors. Sentinels are really not required everywhere.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,5 +14,10 @@ linters:
     - varcheck
 
 issues:
-  new: true
-
+  exclude-rules:
+      # Disallows any kind of `fmt.Errorf("%s is too high", bar)`, too opinionated.
+    - linters: [goerr113]
+      text: "do not define dynamic errors, use wrapped static errors instead"
+      # Most exported consts are self-descriptive.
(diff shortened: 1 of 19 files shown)


twpayne (approved):
Nice cleanup, thanks!

michi-covalent (approved):

ti-mo:
EKS tests unable to run due to provisioning failures, so this is probably good to go. Marking as ready.

rolinh:
@ti-mo Can't we then remove this from the golangci-lint config then?
```
issues:
  new: true
```

ti-mo:
> @ti-mo Can't we then remove this from the golangci-lint config then?
> 
> ```
> issues:
>   new: true
> ```

Sure, it's the default anyway, but didn't want to touch it. Not sure why it was added to begin with.

rolinh (approved):
:rocket:

kaworu (approved):

PR #265, opened 2021-06-01
.github: Fix concurrency groups

pchaigno:
There's currently a single concurrency group per workflow, `workflow_name-scheduled`, because of a regression in d9957fb ("ci: change all in-cluster test runs to pull_request_target"). This commit started using `github.event.pull_request_target` for the concurrency group, but that variable doesn't exist. As a consequence, scheduled workflows may cancel workflows triggered by pull requests and the other way around.

This pull request also fixes commit f998cca (".github: Stop running tests for every push to master") in which triggers were updated without updating the corresponding concurrency group. That mistake was without consequences, but left unnecessary code.

Fixes: https://github.com/cilium/cilium-cli/pull/212.
Fixes: https://github.com/cilium/cilium-cli/pull/244.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -7,7 +7,7 @@ on:
     - cron:  '0 */6 * * *'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.after || 'scheduled' }}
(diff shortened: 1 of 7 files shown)


nbusseneau (approved):
Yes, `pull_request_target` uses the same event as `pull_request`.

nebril (approved):

PR #266, opened 2021-06-01
CODEOWNERS: Update for cilium/ci-structure

pchaigno:
Update code owners for @cilium/ci-structure to include scripts and manifests in `.github/`.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -14,6 +14,12 @@
 * @cilium/janitors
 /CODEOWNERS @cilium/contributing
 /.github/ @cilium/contributing
+/.github/cilium-cli-test-job-chart/ @cilium/github-sec @cilium/ci-structure
+/.github/gcp-vm-startup.sh @cilium/ci-structure
+/.github/get-kubeconfig.sh @cilium/github-sec @cilium/ci-structure
+/.github/in-cluster-test-scripts/ @cilium/ci-structure
+/.github/kind-config*.yaml @cilium/ci-structure
+/.github/tools/ @cilium/ci-structure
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -14,6 +14,12 @@
 * @cilium/janitors
 /CODEOWNERS @cilium/contributing
 /.github/ @cilium/contributing
+/.github/cilium-cli-test-job-chart/ @cilium/ci-structure
+/.github/gcp-vm-startup.sh @cilium/ci-structure
+/.github/get-kubeconfig.sh @cilium/ci-structure
+/.github/in-cluster-test-scripts/ @cilium/ci-structure
```
pchaigno, line 20: @nebril Which of these need to be owned by cilium/maintainers due to potential security impact?

michi-covalent (approved):

christarazi (approved):

CODEOWNERS
nebril: `cilium-cli-test-job-chart` should probably be owned by maintainers, I can imagine a scenario when templating helm allows to leak secrets
`gcp-vm-startup.sh` runs in gcp vm, so it should be safe
`get-kubeconfig.sh` should be owned by maintainers
`in-cluster-test-scripts.sh` are sent to cluster and executed there, so probably doesn't need maintainers review.

nebril (changes requested):

CODEOWNERS
pchaigno: Thanks! :pray: 

I added @cilium/github-sec to the security-sensitive entries and added you as a co-author.

nebril (approved):

Issue #260, opened 2021-05-31
upgrade: `cilium upgrade` does not patch the Hubble Relay image


kaworu:
When Hubble Relay has been installed (for example through `cilium hubble enable`), `cilium upgrade` doesn't update the Hubble Relay image to the upgraded version.

GoGstickGo:
Do you want to have `cilium hubble upgrade` cmd or when one runs `cilium upgrade`  it should upgrade hubble as well if it installed ?

rolinh:
@GoGstickGo Hubble Relay lives in the Cilium tree and depends on Cilium releases. I think the aim should thus be to handle the upgrade as part of `cilium upgrade`.

GoGstickGo:
Roger that, thanks for the  clarification . @rolinh

PR #267, opened 2021-06-01
 hubble upgrade with `cilium upgrade`

GoGstickGo:
https://github.com/cilium/cilium-cli/issues/260


GoGstickGo:
Please review, @kaworu @rolinh.

install/upgrade.go
```diff
@@ -81,7 +82,24 @@ func (k *K8sInstaller) Upgrade(ctx context.Context) error {
 		patched++
 	}
 
-	if patched > 0 && k.params.Wait {
+	hubbleDeployment, err := k.client.GetDeployment(ctx, k.params.Namespace, defaults.RelayDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		k.Log("❌hubble is not enabled, unable to upgrade")
```
twpayne, line 87: This isn't quite right: if `err != nil` then it doesn't necessarily mean that Hubble is not enabled, it might be some other error. It would be better to detect Hubble not being enabled directly, and return other errors.

twpayne (changes requested):

install/upgrade.go
GoGstickGo: Just for my own sake . Only another error would be if the client fails. How can Hubble running without the deployment existing ?

install/upgrade.go
```diff
@@ -81,7 +88,55 @@ func (k *K8sInstaller) Upgrade(ctx context.Context) error {
 		patched++
 	}
 
-	if patched > 0 && k.params.Wait {
+	hubbleDeployment, err := k.client.GetDeployment(ctx, k.params.Namespace, defaults.RelayDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		k.Log("ℹ️  Unable to retrieve Deployment of hubble-relay")
+		message++
```
twpayne, line 94: Incrementing `message` has no effect here, as any new value is overwritten at line 97.

install/upgrade.go
```diff
@@ -81,7 +88,55 @@ func (k *K8sInstaller) Upgrade(ctx context.Context) error {
...
+	}
+
+	message, err = k.ValidateHubble(ctx)
+	if err != nil {
+		k.Log("Error:%q", err)
+	}
+
+	if message >= 1 && err == nil {
```
twpayne, line 102: Rather than accumulating state in local variables (e.g. counting `message`s and populating `err`) and then trying to work out what happened from the state (as is done here), I'd suggest structuring this a clear sequence of steps, each of which does its own error handling.

install/upgrade.go
twpayne: Exactly, the call to the client could fail in many ways (e.g. because the network is down). Receiving an error from the `k.client.GetDeployment` does not imply that Hubble is not enabled.

GoGstickGo: Actually I don't agree as if the client down the whole `cilium-cli` is dead as it is. It wouldn't even get there to check. Second of all if it can't retrieve  deployment template doesn't matter if the `Hubble` enabled or not. Upgrade wouldn't work as `cilium-cli` uses  the deployment template to upgrade. I do  agree to check Hubble enabled properly however deployment template is must have. It must skip the upgrade if can't retrieve deployment template hence the error message correct.

kaworu: Agree here that when the k8s client is unusable then the cli should fail (there is not much we can do), but the cli should fail with an error message that can help understand the underlying issue.

Reporting "hubble is not enabled" when encountering a network issue with a cluster where hubble _is_ enabled would be confusing, and I think that was @twpayne's point.

GoGstickGo: However if there is network issue , it's going to error out here:https://github.com/cilium/cilium-cli/blob/94cba869fc0d176a4a9a917dd9269c83dd1ee50f/internal/cli/cmd/cmd.go#L45 I believe.

kaworu: @GoGstickGo the issue could happen between the k8s client initialization and the commented line, and in that case then your assumption doesn't hold (it won't error at the codepath you've pointed out).

GoGstickGo: You are saying cli successfully established client BUT when it gets to hubble upgrade it fails because of network error. 🤔

Error out with network error message  it could be also misleading as it's simply not true. Further more any network error should be captured much earlier as other functions: https://github.com/cilium/cilium-cli/blob/e55101c56e8947e1dd986383f0da0bb99df51aaa/install/upgrade.go#L34 & https://github.com/cilium/cilium-cli/blob/e55101c56e8947e1dd986383f0da0bb99df51aaa/install/upgrade.go#L40 don't do any error handling related to the network. To return network error only when the Hubble upgrade it's also far from correct.

Look ,this is becoming nitpicking ,actually I lost interest. I am going to close this PR. Thanks for your feedback.

Issue #224, opened 2021-05-14
External workload flake - vm Cilium agent cannot connect to the clustermesh apiserver.


nebril:
This is a transient failure that I observed when working on https://github.com/cilium/cilium-cli/pull/212

Cilium install script on external vm failed timing out, waiting for Cilium agent to come up.

Cilium logs showed that agent was unable to connect to the ip of the clustermesh apiserver deployed in gke cluster.

This issue tend to happen several times in a row and then it "fixes" itself. My working theory is that sometimes gke cluster and vm are created in a way that allows vm->cluster connectivity (then the workflow passes), and sometimes it doesn't.

If seen again, we might try creating firewall rules explicitly allowing vm->cluster traffic.

PR #285, opened 2021-06-04
.github/workflows: move Go module vendoring check to build checks

tklauser:
Instead of instantiating a separate job checking out its own copy of the
code, move the module vendoring check to the build job which already has
the code checked out.

Also move the static checks and the module vendoring check before the
build and tests steps.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -23,12 +23,6 @@ jobs:
       with:
         go-version: 1.16.4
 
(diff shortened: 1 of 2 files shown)


christarazi (approved):

joestringer (approved):

tklauser:
Only failing test is a known flake: #224. Marking as ready to merge.

Issue #203, opened 2021-04-30
azure: remove Azure Service Principal-related code


nbusseneau:
**Dependent on https://github.com/cilium/cilium/issues/15966 being integrated into `cilium-operator-azure`**

After removing requirement for Azure SP from the operator, remove SP-related code from the CLI:

- CLI options/flags
- Automated SP creation if not provided via options/flags

PR #286, opened 2021-06-04
workflows: overhaul and standardize all jobs

nbusseneau:
Please see individual commits :)

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -39,7 +39,7 @@ jobs:
       - name: Login to Azure
         uses: azure/login@77f1b2e3fb80c0e8645114159d17008b8a2e475a
         with:
-          creds: ${{ secrets.AZURE_CREDENTIALS }}
+          creds: ${{ secrets.AZURE_PR_SP_CREDS }}
 
       - name: Display Azure CLI info
         uses: azure/CLI@4b58c946a0f48d82cc2b6e31c0d15a6604859554
@@ -94,9 +94,9 @@ jobs:
           cilium install \
             --cluster-name=${{ env.name }} \
             --azure-resource-group ${{ env.name }} \
-            --azure-tenant-id ${{ secrets.AZURE_TENANT_ID }} \
-            --azure-client-id ${{ secrets.AZURE_CLIENT_ID }} \
-            --azure-client-secret ${{ secrets.AZURE_CLIENT_SECRET }} \
+            --azure-tenant-id ${{ secrets.AZURE_PR_SP_TENANT_ID }} \
+            --azure-client-id ${{ secrets.AZURE_PR_SP_CLIENT_ID }} \
+            --azure-client-secret ${{ secrets.AZURE_PR_SP_CLIENT_SECRET }} \
             --wait=false \
             --config monitor-aggregation=none
 
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -19,10 +19,10 @@ jobs:
     timeout-minutes: 30
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@37335c7bb261b353407cff977110895fa0b4f7d8
         with:
           go-version: 1.16.4
 
@@ -35,12 +35,12 @@ jobs:
         run: sudo make install
 
       - name: Azure Login
-        uses: azure/login@v1
+        uses: azure/login@77f1b2e3fb80c0e8645114159d17008b8a2e475a
         with:
           creds: ${{ secrets.AZURE_CREDENTIALS }}
 
       - name: Azure CLI info
-        uses: azure/CLI@v1
+        uses: azure/CLI@4b58c946a0f48d82cc2b6e31c0d15a6604859554
         with:
           azcliversion: 2.0.72
           inlineScript: |
@@ -93,7 +93,7 @@ jobs:
 
       - name: Upload Artifacts
         if: ${{ always() }}
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@ee69f02b3dfdecd58bb31b4d133da38ba6fe3700
         with:
           name: cilium-sysdump-out.zip
           path: cilium-sysdump-out.zip
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -12,7 +12,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  clusterName: cilium-cli-ci-${{ github.run_number }}
+  name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+  location: westeurope
 
 jobs:
   installation-and-connectivity:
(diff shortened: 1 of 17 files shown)


nbusseneau:
@cilium/maintainers I will need you to setup our new credentials for `cilium-pr` containers on all platforms in `cilium-cli` secrets. I think @nebril has access? If yes, we can coordinate directly on Slack as we did previously with `cilium/cilium`.

.github/workflows/aks.yaml
```diff
@@ -35,64 +36,103 @@ jobs:
...
+              -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+              ${{ github.event.issue.pull_request.url || github.event.pull_request.url }})
+            SHA=$(echo "$PR_API_JSON" | jq -r ".head.sha")
+            OWNER=$(echo "$PR_API_JSON" | jq -r ".number")
+          else
+            SHA=${{ github.sha }}
+            OWNER=${{ github.sha }}
+          fi
+
+          echo ::set-output name=sha::${SHA}
```
nebril, line 66: I don't see `sha` used in this file.

.github/workflows/externalworkloads.yaml
```diff
@@ -12,9 +12,9 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  clusterName: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}-vm
```
nebril, line 15: nit: do we want `-vm` suffix in cluster name?

.github/workflows/multicluster.yaml
```diff
@@ -21,42 +22,72 @@ jobs:
...
+              -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+              ${{ github.event.issue.pull_request.url || github.event.pull_request.url }})
+            SHA=$(echo "$PR_API_JSON" | jq -r ".head.sha")
+            OWNER=$(echo "$PR_API_JSON" | jq -r ".number")
+          else
+            SHA=${{ github.sha }}
+            OWNER=${{ github.sha }}
+          fi
+          CLUSTER_CIDR_1=$(gcloud container clusters describe ${{ env.clusterName1 }} --zone ${{ env.zone }} --format="value(clusterIpv4Cidr)")
+          CLUSTER_CIDR_2=$(gcloud container clusters describe ${{ env.clusterName2 }} --zone ${{ env.zone }} --format="value(clusterIpv4Cidr)")
```
nebril, line 53: neither of these clusters is created yet

nebril (changes requested):
That's a nice cleanup!

I found some issues, it might be worth trying to run these with `pull_request` to make sure nothing else is broken.

.github/workflows/aks.yaml
nbusseneau: Yea, since we don't have AKS tests running in-cluster... I decided to leave it there mostly for consistency, but we can remove it.

.github/workflows/externalworkloads.yaml
nbusseneau: I mostly wanted to have it explicit when we list clusters from GKE if they are part of the default workflow (nothing), multicluster workflow (`mesh`) or external workloads workflow (`vm`). Do you think we should find another suffix?

.github/workflows/multicluster.yaml
nbusseneau: Good catch, will fix.

.github/workflows/aks.yaml
```diff
@@ -35,64 +36,103 @@ jobs:
...
           python cilium-sysdump.zip --output cilium-sysdump-out
-          az aks delete --yes --name ${{ env.clusterName }} --resource-group cilium-ci --no-wait
         shell: bash {0}
 
-      - name: Upload Artifacts
+      - name: Clean up AKS
+        if: ${{ always() }}
+        run: |
+          az group delete --name ${{ env.name }} --yes --no-wait
+        shell: bash {0}
```
pchaigno, line 133: Is this line needed? We don't have the same on the cilium/cilium side.

.github/workflows/multicluster.yaml
nbusseneau: Fixed in latest push, PTAL.

pchaigno (approved):
LGTM with one question below.

Note: The main commit, `workflows: overhaul and standardize all jobs`, has a lot of changes so I'm not very confidant in my review there.

.github/workflows/aks.yaml
nbusseneau: It is not needed in this specific instance, but I have decided to include it (and will also include on `cilium/cilium`) as I think it is warranted for this specific type of step.

In the case of the multicluster or external workloads workflows for example, this ensures even if the first cleanup fails that the second one will be run anyway.

We might in the future have additional resources that need to be cleaned up and that will get added to the cleanup steps in other workflows. Already having the `shell: bash {0}` ensures that we won't forget about adding it if that happens. Also, it keeps the overall workflow structure consistent :)

pchaigno: > In the case of the multicluster or external workloads workflows for example, this ensures even if the first cleanup fails that the second one will be run anyway.

Could you add a comment to explain that? I don't really mind including it in all cleanup steps even if it's unnecessary, but that also means you'll have to copy the comment everywhere ;-)

nbusseneau: What do you mean, adding a comment?

pchaigno: I mean adding a comment above `shell: bash {0}` to explain why it's sometimes needed and why we included it. So that others don't have to guess or ask you :-)

nbusseneau: Ah, I see. Is that really something we want to do...?
We were already leveraging custom shells in the post-test information gathering steps. It just disables the default fail-fast behaviour: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
Since this is part of the default GH syntax and not a hidden trick / workaround, I would personally vote against commenting to explain why we have it, especially if we'd have to copy/paste it everywhere. It does not make sense to me, just like I would not comment `runs-on` or anything that's part of the default workflow syntax.

pchaigno: What the line does was not obvious to me from (maybe it is to others), but on top of that, it doesn't say why it's needed for one step and not all of them.

If you used e.g. `.strategy.fail-fast` for one workflow and not all of them, I would expect a similar comment to explain the rationale. Since we're using `shell: bash {0}` for one step but not others, I think the same apply. The comment can be as simple as `Disable fail-fast to always try to delete all resources`.

nbusseneau: I think the difference comes from the fact you think this is documenting a design choice, whereas I think this is documenting basic GH syntax (e.g. it would be like if we documented our `id` and `steps.<id>.outputs` usage), because to me having a "fail-fast is disabled" marker is equivalent to having "always try to delete all resources".

But I get why it might be non-obvious, so I'll add a comment.

nbusseneau:
All tests in #289 passed except multicluster: https://github.com/cilium/cilium-cli/actions/runs/918228985
The console indicates an issue at `cilium install` with the CIDR detection:
```
+ cilium install --context gke_***_us-west2-a_cilium-cilium-cli-918228985-mesh-1
🔮 Auto-detected Kubernetes kind: GKE
ℹ️  Cilium version not set, using default version "v1.9.8"
🔮 Auto-detected cluster name: gke-***-us-west2-a-cilium-cilium-cli-918228985-mesh-1
🔮 Auto-detected IPAM mode: kubernetes
🔮 Auto-detected datapath mode: gke
🔮 Custom datapath mode: gke
❌ Unable to auto-detect GKE native routing CIDR. Is "gcloud" installed?

Error: Unable to install Cilium: unable to execute gcloud [container clusters describe cilium-cilium-cli-918228985-mesh-1 --zone us-west2-a --format value(clusterIpv4Cidr)] to extract native routing CIDR: exec: "gcloud": executable file not found in $PATH
ℹ️  You can set the native routing CIDR manually with --native-routing-cidr
```
This is due to a missing `\` in `multicluster.sh`:
```sh
# Install Cilium in cluster1
cilium install \
  --context "${CONTEXT1}"
  --cluster-name "${CLUSTER_NAME_1}" \
  --cluster-id 1 \
  --config monitor-aggregation=none \
  --native-routing-cidr="${CLUSTER_CIDR_1}"
```
Fixing...

.github/workflows/aks.yaml
nbusseneau: Fixed in latest push :)

nbusseneau:
Multicluster test failed again: https://github.com/cilium/cilium-cli/actions/runs/918739021
Because I used `cilium --context "${CONTEXT1}" clustermesh status --wait` instead of `cilium --context "${CONTEXT1}" status --wait` when waiting for Cilium to be ready before enabling clustermesh.

.github/workflows/externalworkloads.yaml
nebril: I think that's fine.

.github/workflows/aks.yaml
nebril: I would remove it, since it just clutters the file (setting it uses several lines above), but it's mostly a nitpick.

nbusseneau:
Multicluster workflow is fixed but seems to be hitting flakes: https://github.com/cilium/cilium-cli/actions/runs/919028776
```
📋 Test Report
❌ 1/9 tests failed (2/69 actions), 0 warnings, 0 tests skipped, 4 scenarios skipped:
Test [echo-ingress]:
  ❌ echo-ingress/pod-to-pod/curl-1: cilium-test/client-8655c47fd7-9n9f9 (10.64.0.72) -> cilium-test/echo-other-node-646f4f6797-wq9xb (10.108.1.105:8080)
  ❌ echo-ingress/pod-to-pod/curl-3: cilium-test/client-8655c47fd7-9n9f9 (10.64.0.72) -> cilium-test/echo-other-node-646f4f6797-wq9xb (10.108.1.105:8080)

Error: Connectivity test failed: 1 tests failed
```
Investigating past builds leads me to believe this PR actually fixed the workflow, because the current version of the workflow can get marked as success even though the connectivity tests actually do not run: https://github.com/cilium/cilium-cli/actions/runs/919407371
```
Error: Connectivity test failed: timeout reached waiting for pod IDs in ipcache of Cilium pod kube-system/cilium-m8cmk
```
This is due to the fact that since #212 the tests are ran without waiting for `cilium status` to be green before enabling cluster mesh (confirmed since this sort of failure has been happening since the first scheduled job following merge of #212: https://github.com/cilium/cilium-cli/actions/runs/897360610). In some cases though it does work (example from this very PR, ran via `pull_request_target`: https://github.com/cilium/cilium-cli/actions/runs/918740247).

In any case, all of this leads me to believe this PR does not introduce the issues.

.github/workflows/aks.yaml
nbusseneau: Since it's the only workflow that would diverge from the norm, I would probably leave it like that but can do either way 🤔 We're supposed to fix the reliance on the SP anyway (#203), so when it's gone it'll rejoin the others.

pchaigno (approved):

nebril (approved):

PR #294, opened 2021-06-09
ci: wait for Cilium start before enabling clustermesh

### LINKED ISSUE #224: External workload flake - vm Cilium agent cannot connect to the clustermesh apiserver.

nebril:
This change fixes a bug when clustermesh apiserver was created directly
after Cilium pods were created which caused the apiserver pod to not be
managed by Cilium

fixes #224

### DIFF
--- a/.github/in-cluster-test-scripts/external-workloads-install.sh
+++ b/.github/in-cluster-test-scripts/external-workloads-install.sh
@@ -4,6 +4,7 @@ set -x
 set -e
 
 cilium install --cluster-name "${CLUSTER_NAME}" --restart-unmanaged-pods=false --config monitor-aggregation=none --config tunnel=vxlan --native-routing-cidr="${CLUSTER_CIDR}"
+cilium status --wait
 
 cilium clustermesh enable
 cilium clustermesh status --wait --wait-duration 5m


nebril:
clustermesh apiserver was in crashloopbackoff, investigating

nebril:
it appears that clustermesh apiserver  was not managed by Cilium which caused failure.

nebril:
The fix seems to reliably allow vm cilium to start up, will redo the PR to be ready to merge and review.

nebril:
This will be included in https://github.com/cilium/cilium-cli/pull/286

nbusseneau:
Included! Closing in favor of #286 :)

PR #295, opened 2021-06-09
Add encryption support to sysdump.

bmcustodio:
Adds a `--encrypt` flag (defaulting to `true`) that encrypts the resulting zip file with either the bundled public key or the public key specified via `--encryption-key`.


sysdump/encryption.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return "", err
+	}
+	// Encrypt the zip file.
+	h := &openpgp.FileHints{
+		FileName: filepath.Base(pathToZipFile),
+		IsBinary: true,
+		ModTime:  time.Now(),
+	}
+	encOut, err := openpgp.Encrypt(o, []*openpgp.Entity{e}, nil, h, nil)
```
gandro, line 50: Maybe I'm missing something, but I think `[]*openpgp.Entity{e}` allows you to specify multiple recipients. You would have to change `Options.EncryptionKey` to a `[]string` and the CLI argument to a `StringSliceVar`, but I think we could support multiple public keys this way?

bmcustodio: That would make a lot of sense, but from my 🔍 the entities would need to be read using `openpgp.ReadKeyRing` and unfortunately that seems to be [broken](https://github.com/golang/go/issues/29082#issuecomment-809621751) 😞

gandro: I don't think I follow. When would `openpgp.ReadKeyRing` have be used?

The way I'm imagining this would be that the user invokes this via `cilium sysdump --encryption-key default  --encryption-key my-own-key.gpg --encryption-key my-bosses-key.gpg`

This then would cause 3 invocations to `readDestinationEntity` - the default key would be read from the goembeded `pubKey`, the other two via `opengpg.ReadEntity`. All three are then appended to a `[]*openpgp.Entity` and passed here.

When decrypting, we will only use a single key anyways. I for example would use `my-own-key.gpg`.

sysdump/sysdump.go
```diff
@@ -687,6 +687,32 @@ func (c *Collector) Run() error {
...
+		}
+		f = e
+	}
+	// Move the output file to the current directory.
+	p, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("failed to get current directory: %w", err)
+	}
+	r := filepath.Join(p, filepath.Base(f))
+	if err := os.Rename(f, r); err != nil {
```
gandro, line 710: Note: This fails on my machine with a `LinkError` due to `/tmp` being on a different filesystem (which I think is quite common on linux).

I would propose creating a temporary file next to the target file (e.g. via `os.CreateTemp`) and then use `os.Rename` instead. It's probably also better not to put potentially sensitive data in `/tmp`

sysdump/encryption.go
gandro: The following patch works for me, i.e. I can decrypt the sysdump via `gpg --decrypt` without the default private key installed:

```diff
diff --git a/internal/cli/cmd/sysdump.go b/internal/cli/cmd/sysdump.go
index a49923f51bf1..2083ce4d918d 100644
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -106,8 +106,8 @@ func newCmdSysdump() *cobra.Command {
        cmd.Flags().BoolVar(&sysdumpOptions.Encrypt,
                "encrypt", sysdump.DefaultEncrypt,
                "Whether to encrypt the resulting zip file")
-       cmd.Flags().StringVar(&sysdumpOptions.EncryptionKey,
-               "encryption-key", sysdump.DefaultEncryptionKey,
+       cmd.Flags().StringSliceVar(&sysdumpOptions.EncryptionKeys,
+               "encryption-key", []string{sysdump.DefaultEncryptionKey},
                "The path to the OpenPGP-compliant public key to use for encrypting the resulting zip file. Defaults to an embeded OpenPGP-compliant public key")
 
        return cmd
diff --git a/sysdump/defaults.go b/sysdump/defaults.go
index d151bda62f7d..e6e8c0618da1 100644
--- a/sysdump/defaults.go
+++ b/sysdump/defaults.go
@@ -31,7 +31,7 @@ const (
        DefaultCiliumOperatorNamespace     = DefaultCiliumNamespace
        DefaultDebug                       = false
        DefaultEncrypt                     = true
-       DefaultEncryptionKey               = ""
+       DefaultEncryptionKey               = "default"
        DefaultHubbleLabelSelector         = labelPrefix + "hubble"
        DefaultHubbleNamespace             = DefaultCiliumNamespace
        DefaultHubbleRelayLabelSelector    = labelPrefix + "hubble-relay"
diff --git a/sysdump/encryption.go b/sysdump/encryption.go
index 44f2f7b42d24..3cc9f8552649 100644
--- a/sysdump/encryption.go
diff --git a/internal/cli/cmd/sysdump.go b/internal/cli/cmd/sysdump.go
index a49923f51bf1..2083ce4d918d 100644
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -106,8 +106,8 @@ func newCmdSysdump() *cobra.Command {
        cmd.Flags().BoolVar(&sysdumpOptions.Encrypt,
                "encrypt", sysdump.DefaultEncrypt,
                "Whether to encrypt the resulting zip file")
-       cmd.Flags().StringVar(&sysdumpOptions.EncryptionKey,
-               "encryption-key", sysdump.DefaultEncryptionKey,
+       cmd.Flags().StringSliceVar(&sysdumpOptions.EncryptionKeys,
+               "encryption-key", []string{sysdump.DefaultEncryptionKey},
                "The path to the OpenPGP-compliant public key to use for encrypting the resulting zip file. Defaults to an embeded OpenPGP-compliant public key")
 
        return cmd
diff --git a/sysdump/defaults.go b/sysdump/defaults.go
index d151bda62f7d..e6e8c0618da1 100644
--- a/sysdump/defaults.go
+++ b/sysdump/defaults.go
@@ -31,7 +31,7 @@ const (
        DefaultCiliumOperatorNamespace     = DefaultCiliumNamespace
        DefaultDebug                       = false
        DefaultEncrypt                     = true
-       DefaultEncryptionKey               = ""
+       DefaultEncryptionKey               = "default"
        DefaultHubbleLabelSelector         = labelPrefix + "hubble"
        DefaultHubbleNamespace             = DefaultCiliumNamespace
        DefaultHubbleRelayLabelSelector    = labelPrefix + "hubble-relay"
diff --git a/sysdump/encryption.go b/sysdump/encryption.go
index 44f2f7b42d24..3cc9f8552649 100644
--- a/sysdump/encryption.go
+++ b/sysdump/encryption.go
@@ -29,11 +29,15 @@ import (
 //go:embed pubkey.gpg
 var pubKey []byte
 
-func createEncryptedZipFile(pathToZipFile string, pathToEncryptionKey string) (string, error) {
+func createEncryptedZipFile(pathToZipFile string, pathsToEncryptionKey []string) (string, error) {
        // Read the destination entity.
-       e, err := readDestinationEntity(pathToEncryptionKey)
-       if err != nil {
-               return "", err
+       var entityList []*openpgp.Entity
+       for _, k := range pathsToEncryptionKey {
+               e, err := readDestinationEntity(k)
+               if err != nil {
+                       return "", err
+               }
+               entityList = append(entityList, e)
        }
        // Create the destination file.
        d := pathToZipFile + ".gpg"
```

bmcustodio: @gandro oh I see now what you mean — read from several different public key files instead of from a keyring file containing the three public keys. I think that's very nice — do you want to push this patch to my branch? 🙂

sysdump/sysdump.go
bmcustodio: Hmmm that's interesting 🤔 I'll try to have a look.

sysdump/encryption.go
gandro: Done!

sysdump/sysdump.go
gandro: Fixed!

errordeveloper (approved):
Great idea to embed the public key instead of asking the users to provide one! LGTM :)

joestringer:
It looks like this PR has gone stale and it would now need a rebase. If you are interested in reviving this, it may be worth syncing with @tklauser / @michi-covalent to agree on the approach, then you would be welcome to reopen a PR to submit this feature again.

PR #298, opened 2021-06-09
refactor: inline context check

ldelossa:
nit pr which removes the isDone context check method
in favor or performing the check inline.


connectivity/check/context.go
```diff
@@ -234,7 +234,7 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 
 	// Execute all tests in the order they were registered by the test suite.
 	for _, t := range ct.tests {
-		if isDone(ctx) {
+		if err := ctx.Err(); err != nil {
 			return context.Canceled
```
ldelossa, line 238: I'm now noticing, this error may swallow up the reason why the context is canceled. 

One could do this:
```
if err := ctx.Err(); err != nil {
    return ctx.Err()
}
```

If you want to keep the nice formating you get by using

```
var Canceled = errors.New("context canceled")
```

We could also error wrap 

```
type ErrCanceled struct {
  e error
}
func (e ErrCanceled) Error() string {
  return "context canceled: " + e.e.Error()
}
// implement unwrap methods if desired...
```

rolinh: I think that the following approach is totally fine:
```go
if err := ctx.Err(); err != nil {
    return ctx.Err()
}
```
If we wanted to error wrap, I'd rather use this simpler approach:
```go
return fmt.Errorf("context done: %w", ctx.Err())
```
Although I'm not sure it would add any value.

ldelossa: Agreed. Will update.

rolinh (approved):

ldelossa:
@rolinh - seems like tests are failing. the AKS test is failing expectedly since I'm opening PR from my fork, but the External one, I'm not too sure. Whats the best way to proceed?

ldelossa:
closing this and re-opening as a branch against upstream for AKS tests.

see:
https://cilium.slack.com/archives/C7PE7V806/p1623337145157800

PR #303, opened 2021-06-10
workflows: use the native sysdump

bmcustodio:
Moves workflows from using `cilium-sysdump.zip` to using the new, native implementation. As it is impossible to copy files from terminated pods without resorting to using PVs (or some sort of base64-encoding and log parsing voodoo), I am proposing we use FIFOs to copy the sysdump from the pods while they're still running, which should lead to graceful job termination once read from.

I am not entirely sure this will work as expected, though, and for that reason I am marking it as draft :)


bmcustodio:
Closing in favour of #304.

PR #304, opened 2021-06-10
workflows: use the native sysdump

bmcustodio:
Moves workflows from using cilium-sysdump.zip to using the new, native implementation. ~As it is impossible to copy files from terminated pods without resorting to using PVs, I am proposing we use FIFOs to copy the sysdump from the pods while they're still running, which should lead to graceful job termination once read from.~

I am not entirely sure this will work as expected, though, and for that reason I am marking it as draft.

### DIFF
--- a/.github/in-cluster-test-scripts/eks-tunnel.sh
+++ b/.github/in-cluster-test-scripts/eks-tunnel.sh
@@ -3,6 +3,7 @@
 set -x
 set -e
 
+
 # Enable Relay
 cilium hubble enable
 
@@ -18,3 +19,7 @@ cilium connectivity test --all-flows
 
 # Retrieve Cilium  status
 cilium status
+
+# Grab a sysdump and move it to the persistent volume.
+cilium sysdump --output-filename cilium-sysdump-out
+mv cilium-sysdump-out.zip /output/cilium-sysdump-out.zip
--- a/.github/workflows/eks.yaml
+++ b/.github/workflows/eks.yaml
@@ -130,6 +130,7 @@ jobs:
           kubectl logs --timestamps -n kube-system job/cilium-cli-install
           kubectl logs --timestamps -n kube-system job/cilium-cli
           kubectl get pods --all-namespaces -o wide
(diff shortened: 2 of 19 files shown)


.github/in-cluster-test-scripts/eks-tunnel.sh
```diff
@@ -3,6 +3,10 @@
 set -x
 set -e
 
+# Create a FIFO through which we'll pipe the sysdump before the job terminates.
+# NOTE: This MUST be the first step in the script.
```
nebril, line 7: nitpick - why does it have to be the first step in the script?

.github/workflows/eks.yaml
```diff
@@ -102,8 +103,9 @@ jobs:
             --set tag=${{ steps.vars.outputs.sha }} \
             --set cluster_name=${{ env.clusterName }}
 
-      - name: Wait for test job
+      - name: Wait for and copy the sysdump in the background and wait for job termination
         run: |
+          kubectl -n kube-system exec job/cilium-cli -- cat /tmp/cilium-sysdump-out > cilium-sysdump-out.zip &
```
nebril, line 108: I have had all kinds of trouble trying to use `exec` in gh workflows before, sometimes connections are dropping and this would probably cause sysdumps to be malformed. Let's see how this works, the alternative to this approach is to put the sysdump in a configmap inside cli pod and retrieve it via `kubectl get` in workflow later (like we do with vm install script in external workloads workflow).

bmcustodio: That won't work unfortunately — configmaps are limited to 1MB 😞 We'll have to resort to PVCs in that case.

.github/in-cluster-test-scripts/eks-tunnel.sh
bmcustodio: That is to reduce the failure window for `kubectl exec -- cat /fifo` when `/fifo` does not exist — for that reason, I guess we should always create them as early as possible.

bmcustodio:
Closing in favour of #405.

PR #305, opened 2021-06-10
refactor: inline context check

ldelossa:
nit pr which removes the isDone context check method
in favor or performing the check inline.

we also stop swallowing the specific ctx error encountered
and report it to the caller.

reference: #298


rolinh (approved):

nbusseneau (approved):
LGTM, thanks for the refactor.

kkourt (approved):

nebril (approved):

PR #306, opened 2021-06-11
getFlows: support deadline flow returns

ldelossa:
previous to this commit an error would be returned
with no flows if a Deadline context was provided
to the target method.

so follow the same logic as a Cancel context being passed
in, the switch statement is updated to also return flows
if a Deadline is exceeded.

in this new case all flows received before the Deadline
will be returned.

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -627,7 +627,7 @@ func (a *Action) getFlows(ctx context.Context, hubbleClient observer.ObserverCli
 	for {
 		res, err := b.Recv()
 		switch err {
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):

rolinh (approved):

ti-mo (approved):
Thanks! I have a feeling this might not be an isolated case. Might be another few occurrences here and there (we have context creep in a large amount of functions).

ldelossa:
@ti-mo I'll do a full sweep of the code next.

jrajahalme (approved):

PR #308, opened 2021-06-11
azure: restrict Service Principal scope to Resource Group

nbusseneau:
Please see invidivual commits :)

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -95,9 +95,6 @@ jobs:
           cilium install \
             --cluster-name=${{ env.name }} \
             --azure-resource-group ${{ env.name }} \
-            --azure-tenant-id ${{ secrets.AZURE_PR_SP_TENANT_ID }} \
-            --azure-client-id ${{ secrets.AZURE_PR_SP_CLIENT_ID }} \
-            --azure-client-secret ${{ secrets.AZURE_PR_SP_CLIENT_SECRET }} \
             --wait=false \
             --config monitor-aggregation=none
 
--- a/install/azure.go
+++ b/install/azure.go
@@ -105,6 +105,16 @@ func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
 			k.params.Azure.ClientID, k.params.Azure.TenantID)
 	}
 
+	// `az aks create` requires an existing resource group in which to create a
+	// new AKS cluster, but a single resource group may hold multiple AKS clusters.
+
+	// Internally, AKS creates an intermediate resource group (named
+	// `MC_{RG_name}_{cluster_name}_{location}`) to regroup all AKS nodes for
+	// this cluster.
+
+	// The CLI installs itself into this intermediate resource group, and thus
+	// derives it from the user-given resource group and cluster name using
+	// `az aks show`.
 	bytes, err := k.Exec("az", "aks", "show", "--subscription", k.params.Azure.SubscriptionID, "--resource-group", k.params.Azure.ResourceGroupName, "--name", k.params.ClusterName)
 	if err != nil {
 		return err
@@ -115,8 +125,8 @@ func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
 		return fmt.Errorf("unable to unmarshal az output: %w", err)
 	}
 
-	k.Log("✅ Derived Azure node resource group %s from resource group %s", clusterInfo.NodeResourceGroup, k.params.Azure.ResourceGroupName)
-	k.params.Azure.ResourceGroup = clusterInfo.NodeResourceGroup
+	k.Log("✅ Derived Azure AKS node resource group %s from resource group %s", clusterInfo.NodeResourceGroup, k.params.Azure.ResourceGroupName)
+	k.params.Azure.AKSNodeResourceGroup = clusterInfo.NodeResourceGroup
 
 	return nil
 }
--- a/install/azure.go
+++ b/install/azure.go
@@ -76,35 +76,6 @@ func (k *K8sInstaller) retrieveSubscriptionID(ctx context.Context) error {
 }
 
 func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
-	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" {
-		k.Log("🚀 Creating Azure Service Principal for Cilium operator...")
-		bytes, err := k.azExec("ad", "sp", "create-for-rbac", "--scopes", "/subscriptions/"+k.params.Azure.SubscriptionID)
-		if err != nil {
-			return err
-		}
-
-		p := azurePrincipalOutput{}
-		if err := json.Unmarshal(bytes, &p); err != nil {
-			return fmt.Errorf("unable to unmarshal az output: %w", err)
-		}
-
-		k.Log("✅ Created Azure Service Principal for Cilium operator with App ID %s and Tenant ID %s", p.AppID, p.Tenant)
-		k.params.Azure.TenantID = p.Tenant
-		k.params.Azure.ClientID = p.AppID
-		k.params.Azure.ClientSecret = p.Password
-	} else {
-		if k.params.Azure.TenantID == "" || k.params.Azure.ClientID == "" || k.params.Azure.ClientSecret == "" {
-			k.Log(`❌ All three parameters are required for using an existing Azure Service Principal:
-   - Tenant ID (--azure-tenant-id)
-   - Client ID (--azure-client-id)
-   - Client Secret (--azure-client-secret)`)
-			return fmt.Errorf("missing at least one of Azure Service Principal parameters")
-		}
-
-		k.Log("✅ Using manually configured Azure Service Principal for Cilium operator with App ID %s and Tenant ID %s",
-			k.params.Azure.ClientID, k.params.Azure.TenantID)
-	}
-
 	// `az aks create` requires an existing resource group in which to create a
 	// new AKS cluster, but a single resource group may hold multiple AKS clusters.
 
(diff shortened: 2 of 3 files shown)


install/azure.go
```diff
@@ -123,7 +123,7 @@ func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
 	}
 
 	k.Log("✅ Derived Azure node resource group %s from resource group %s", clusterInfo.NodeResourceGroup, k.params.Azure.ResourceGroupName)
-	k.params.Azure.ResourceGroup = clusterInfo.NodeResourceGroup
+	k.params.Azure.NodeResourceGroup = clusterInfo.NodeResourceGroup
```
pchaigno: So the commit message says (bold is mine):

> Rename to avoid confusion between the **user-given resource group** and the automatically created **AKS resource group** contained within the former.

And the new name is *node resource group*. I'm now very confused :sweat_smile: Is node resource group the user-given resource group or the AKS resource group?

.github/workflows/aks.yaml
```diff
@@ -94,9 +94,6 @@ jobs:
           cilium install \
             --cluster-name=${{ env.name }} \
             --azure-resource-group ${{ env.name }} \
-            --azure-tenant-id ${{ secrets.AZURE_PR_SP_TENANT_ID }} \
-            --azure-client-id ${{ secrets.AZURE_PR_SP_CLIENT_ID }} \
-            --azure-client-secret ${{ secrets.AZURE_PR_SP_CLIENT_SECRET }} \
```
pchaigno, line 99: What's the implication of this change exactly? Does that mean a new service principal is created for each new test or was that already the case?

pchaigno (approved):
One nit and one question. Nothing blocking on my side :rocket:

pchaigno:
Ah, one other question: why will this help with running AKS on PRs from forks?

.github/workflows/aks.yaml
nbusseneau: A new dedicated SP is created for each test:
https://github.com/cilium/cilium-cli/blob/b70e437690f97110e43df96e4cffedff273ab131/install/azure.go#L81-L100

This is also how most users will/should be using `cilium install`, so this also makes testing closer to what users will actually use.

install/azure.go
nbusseneau: The AKS one ;) It seemed obvious to me considering:

- The existing code and contrast between the RG passed as parameter and the one derived later on.
- The Azure API and its objects (the `az aks show` command pertains to the AKS subcommand group, not the RG subcommand group -- which is `az group`).
- The nodes are entirely part of the AKS cluster, the node resource group is the RG automatically created by AKS to hold the nodes.

I can the change the commit / code to make it clearer to non-AKS persons if need be.

nbusseneau:
> Ah, one other question: why will this help with running AKS on PRs from forks?

This is top secret 😇

Joking, basically once we have proven that the CLI works with Service Principals with minimal privileges, we can have the AKS workflow work like the others: when running from `pull_request_target`, the workflow has access to the secrets -- here, a specific AKS Service Principal which has all access rights on the `cilium-pr` subscription, and which we previously were passing to Cilium. We can stop relying on this Service Principal and passing it to Cilium, and instead have the main SP create an ad-hoc restricted SP which will be the one passed to Cilium and accessible in-cluster from the external code.

In effect, we reach security parity with the other workflows -- because the external code only has power over the cluster itself, but nothing more.

nbusseneau:
AKS testing fails due to insufficient privileges for the default Service Principal to create other Service Principals: https://github.com/cilium/cilium-cli/actions/runs/929715725
```
Error: Unable to install Cilium: unable to execute "az [ad sp create-for-rbac --scopes /subscriptions/986ec55c-1e77-4e2e-9ca5-dcbb34c7a110/resourceGroups/MC_cilium-cilium-cli-929715725_cilium-cilium-cli-929715725_westeurope]": exit status 1
```
I will grant more permissions to the main SP so that it can create other SPs next week.

install/azure.go
```diff
@@ -51,12 +56,20 @@ type aksClusterInfo struct {
...
+func (k *K8sInstaller) aksSetup(ctx context.Context) error {
+	if err := k.azureRetrieveSubscriptionID(ctx); err != nil {
+		return err
+	}
+
+	if err := k.azureRetrieveAksNodeResourceGroup(ctx); err != nil {
+		return err
+	}
+
+	err := k.azureSetupServicePrincipal(ctx)
```
ti-mo, line 68:
```suggestion
	return k.azureSetupServicePrincipal(ctx)
```

install/azure.go
```diff
@@ -72,16 +85,36 @@ func (k *K8sInstaller) retrieveSubscriptionID(ctx context.Context) error {
...
 
-	k.Log("✅ Derived Azure Subscription ID %s from subscription %s", ai.ID, ai.Name)
+	k.Log("✅ Derived Azure subscription ID %s from subscription %s", ai.ID, ai.Name)
 	k.params.Azure.SubscriptionID = ai.ID
 
 	return nil
 }
 
-func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
+func (k *K8sInstaller) azureRetrieveAksNodeResourceGroup(ctx context.Context) error {
```
ti-mo, line 94:
```suggestion
func (k *K8sInstaller) azureRetrieveAKSNodeResourceGroup(ctx context.Context) error {
```

install/azure.go
```diff
@@ -72,16 +85,36 @@ func (k *K8sInstaller) retrieveSubscriptionID(ctx context.Context) error {
...
+
+	return nil
+}
+
+func (k *K8sInstaller) azureSetupServicePrincipal(ctx context.Context) error {
 	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" {
 		k.Log("🚀 Creating Azure Service Principal for Cilium operator...")
-		args := []string{"ad", "sp", "create-for-rbac", "--scopes", "/subscriptions/" + k.params.Azure.SubscriptionID}
+		// Restrict the scope of the Service Principal to the strict minimum (i.e. the resource group)
+		args := []string{"ad", "sp", "create-for-rbac", "--scopes", "/subscriptions/" + k.params.Azure.SubscriptionID + "/resourceGroups/" + k.params.Azure.NodeResourceGroup}
```
ti-mo, line 117: I've always wondered what this does behind the scenes. Will this create a new SP on each invocation, or does it re-use an existing one? I'm yet to discover where to view active SPs in the Azure Portal. :sweat_smile:

ti-mo (commented):
Left a few nits, looks reasonable to me overall.

install/azure.go
pchaigno: @ti-mo See https://github.com/cilium/cilium-cli/pull/308#discussion_r650312351.

Maybe this is a good indication a comment is needed around here :slightly_smiling_face:

nbusseneau: As said above, in this instance it will create a new dedicated SP on each call because we did not pass a name to `az ad sp create-for-rbac`. If `--name` is passed, then on first call it creates a new SP with the name, and on subsequent calls it updates the SP with a new client ID and client secret, but keeping the same SP (this is what we used to have before we decided to implement #201, but is definitely not what we want).

install/azure.go
nbusseneau: Fixed in latest push.

install/azure.go
nbusseneau: Thanks, missed that during the refactor. Fixed in latest push.

install/azure.go
nbusseneau: Fixed in latest push.

install/azure.go
nbusseneau: Documented in latest push.

nbusseneau:
Fixed SP privileges so that it may create other SPs: https://github.com/cilium/cilium-cli/actions/runs/936540358.

install/azure.go
```diff
@@ -110,6 +110,18 @@ func (k *K8sInstaller) createAzureServicePrincipal(ctx context.Context) error {
...
+		new AKS cluster, but a single resource group may hold multiple AKS clusters.
+
+		Internally, AKS creates an intermediate resource group (named
+		`MC_{RG_name}_{cluster_name}_{location}`) to regroup all AKS nodes for
+		this cluster.
+
+		The CLI installs itself into this intermediate resource group, and thus
+		derives it from the user-given resource group and cluster name using
+		`az aks show`.
+	*/
```
pchaigno: Note we don't generally use that format for comments. Seems to be the first use in both cilium-cli and cilium/cilium.

pchaigno (approved):

install/azure.go
nbusseneau: OK, what format to we use?

pchaigno: The C-style formats it seems :shrug: 

It seems to be `//` in front of each line most of the time

nbusseneau: Aaaah OK I thought you were talking about something else, my bad 🤦🏻 I will switch to `//` in next push, just waiting for Timo's review in case there are other nits.

install/install.go
```diff
@@ -1019,13 +1019,13 @@ const (
...
 type AzureParameters struct {
-	ResourceGroupName string
-	ResourceGroup     string
-	SubscriptionName  string
-	SubscriptionID    string
-	TenantID          string
-	ClientID          string
-	ClientSecret      string
+	ResourceGroupName    string
+	AksNodeResourceGroup string
```
ti-mo, line 1023:
```suggestion
	AKSNodeResourceGroup string
```
:grin:

install/azure.go
ti-mo: Very solid, thank you :+1:

install/azure.go
ti-mo: I even forgot Go supported this, hadn't seen it in so long. :grin:

install/install.go
nbusseneau: Fixed!

nbusseneau:
Rebasing to fix conflicts with `master` following merge of #226.

nbusseneau:
Following rebase on `master` after conflicts with #226, created #323 to fix issue introduced in #226.
This PR is now dependent on #323 to be merged.

ti-mo (approved):
:shipit:

michi-covalent (approved):

christarazi (approved):

PR #311, opened 2021-06-14
draft: use follow

michi-covalent:


michi-covalent:
will use https://github.com/cilium/cilium-cli/pull/319 instead

PR #315, opened 2021-06-14
gke: Bump up gke timeout to 30 minutes

michi-covalent:
We do `kubectl wait` for 15 minutes, and we don't get sysdump if the
workflow gets cancelled.

Ref: https://github.com/cilium/cilium-cli/runs/2816183763?check_suite_focus=true

### DIFF
--- a/.github/workflows/gke.yaml
+++ b/.github/workflows/gke.yaml
@@ -18,7 +18,7 @@ env:
 jobs:
   installation-and-connectivity:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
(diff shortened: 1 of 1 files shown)


nbusseneau (approved):

michi-covalent:
looks like the change doesn't take effect until it gets merged?

pchaigno (approved):

pchaigno:
> looks like the change doesn't take effect until it gets merged?

Yep, that's expected for all `pull_request_target` triggers.

nebril (approved):

Issue #276, opened 2021-06-03
Pods in `cilium-test` don't respond to interrupt signals


ti-mo:
This is a classic mistake, but the echo and client pods currently don't respond to SIGTERM and need to be SIGKILLed (after k8s's termination grace period). This makes the pods take forever to terminate instead of what should take microseconds.

A brief explanation is here: https://petermalmgren.com/signal-handling-docker.

Using `tini` or `dumb-init` is not necessary, but e.g. the `/bin/sh -c "sleep 100000"` should be replaced by `/bin/sh -c "exec sleep 100000"`. pid1 doesn't receive default signal handlers, and shells don't register default signal handlers (although a process like `sleep` does). Making `sleep` pid1 by replacing the shell's process image using `exec` should fix this.

The echo container might need a fix to the entrypoint script if it doesn't contain an `exec` of the node.js process. (we should probably roll our own, we need something more flexible)

Issue #275, opened 2021-06-03
Create pool of test namespaces to enable cross-namespace Scenarios


ti-mo:
Currently, we're creating a fixed `cilium-test` Namespace that contains a single stack of client/echo Pods. In order to support cross-namespace tests, we should move away from this model in order to make implementing new cross-namespace Scenarios less painful.

I propose to:
- Make testing Namespaces randomly-generated
- Create a `sync.Pool` (or similar) construct that allocates Namespaces on-demand (cattle, not pets)
- Scenarios request one or more Namespaces at the start. If there are Namespaces in the pool that were `Put()` by finished Scenarios, they are re-used.
- Make the harness take care of applying CNPs/KNPs to all Namespaces automatically, and to mutate the policies in a predictable way, like overwriting the `Namespace` field or overwriting certain annotations/labels. We could opt for placeholder values or (my preference) empty values (e.g. requiring `metadata.namespace` or certain label values or matchers to be empty).

This would also imply test Namespaces get removed when the connectivity test suite finishes. Remove `cilium-test` deletion behaviour from `cilium uninstall` as well.

Also, this brings us a step closer to being able to schedule Tests across Namespaces and being able to run them concurrently.

LarsBingBong:
Any news on this? We have to delete the cilium-test namespace after the test ... a little sad when the test is part of some automation actually verifying the key core services of X Kubernetes cluster are running.

Thank you for the great product all in all - Cilium really kicks some tires.

PR #330, opened 2021-06-17
connectivity: use the right test namespace in cnps

bmcustodio:
Previously, the CNPs created during the connectivity test referenced `cilium-test` in `matchLabels` as opposed to the user-specified namespace. This commit addresses that by looking up potential matches and doing the proper replacement.

### DIFF
--- a/connectivity/check/test.go
+++ b/connectivity/check/test.go
@@ -11,7 +11,17 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cilium/cilium-cli/defaults"
+
+	k8sConst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+)
+
+const (
+	// kubernetesSourcedLabelPrefix is the optional prefix used in labels to
+	// indicate they are sourced from Kubernetes.
+	// NOTE: For some reason, ':' gets replaced by '.' in keys so we use that instead.
+	kubernetesSourcedLabelPrefix = "k8s."
 )
 
 type Test struct {
@@ -182,6 +192,30 @@ func (t *Test) WithPolicy(policy string) *Test {
 	pl, err := parsePolicyYAML(policy)
 	if err != nil {
 		t.Fatalf("Parsing policy YAML: %s", err)
+	}
+
+	// Change the default test namespace as required.
+	for i := range pl {
+		pl[i].Namespace = t.ctx.params.TestNamespace
+		if pl[i].Spec != nil {
+			// Check both 'io.kubernetes.pod.namespace' and 'k8s:io.kubernetes.pod.namespace'.
+			for _, k := range []string{k8sConst.PodNamespaceLabel, kubernetesSourcedLabelPrefix + k8sConst.PodNamespaceLabel} {
+				for _, e := range pl[i].Spec.Egress {
+					for _, es := range e.ToEndpoints {
+						if n, ok := es.MatchLabels[k]; ok && n == defaults.ConnectivityCheckNamespace {
+							es.MatchLabels[k] = t.ctx.params.TestNamespace
+						}
+					}
+				}
+				for _, e := range pl[i].Spec.Ingress {
+					for _, es := range e.FromEndpoints {
+						if n, ok := es.MatchLabels[k]; ok && n == defaults.ConnectivityCheckNamespace {
+							es.MatchLabels[k] = t.ctx.params.TestNamespace
+						}
+					}
+				}
+			}
+		}
 	}
 
 	if err := t.addCNPs(pl...); err != nil {
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -49,7 +49,7 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 		return err
 	}
 
-	k.Log("🔥 Deleting cilium-test namespace...")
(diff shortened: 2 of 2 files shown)


connectivity/check/policy.go
```diff
@@ -32,6 +32,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// testNamespacePlaceholder will get replaced with the test namespace in use when parsing CPNs from YAML.
+	testNamespacePlaceholder = "${TEST_NAMESPACE}"
```
ti-mo, line 37: I'm not a fan of the placeholder idea, that will create complexity sprawl in the long run.

My idea was to have policies in yaml form with e.g. the namespace field empty and populate them in Go during execution. (empty fields to make sure we don't inadvertently clobber them) This would mean these policies can no longer be applied using `kubectl`, but that's not the case with this approach either.

ti-mo (changes requested):
Related: https://github.com/cilium/cilium-cli/issues/275. Also, see https://github.com/cilium/cilium-cli/issues/276, which makes spinning down test namespaces quite time-consuming at the moment. Should be an easy fix, though. Let's hold off on generating random namespaces for now.

I think a solution to this problem might require some more thought. Populating the namespace field is trivial, but the target field of matchLabel annotations looks a bit trickier. Any thoughts?

install/uninstall.go
```diff
@@ -58,7 +58,7 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 		return err
 	}
 
-	k.Log("🔥 Deleting cilium-test namespace...")
+	k.Log("🔥 Deleting the %q test namespace...", k.params.TestNamespace)
```
ti-mo, line 61: This would require passing the correct test namespace to `cilium uninstall` as well, which a user would quite easily overlook. Not a comment on this PR, but we've discussed before that this code shouldn't be here, only `cilium test` should be responsible for managing the lifecycle of the test namespace.

connectivity/check/policy.go
bmcustodio: > I think a solution to this problem might require some more thought.

That's why I was proposing this as a quick, interim fix until we figure it out. 🙂

> This would mean these policies can no longer be applied using kubectl, but that's not the case with this approach either.

`cat connectivity/manifests/client-egress-to-echo.yaml| TEST_NAMESPACE=foo envsubst | k apply -f-` works 🙂 Of course it isn't as simple as just `kubectl`ing, but it does actually work.

bmcustodio:
> I think a solution to this problem might require some more thought.

@ti-mo that's why I was proposing this as a quick, interim fix until we figure it out. 🙂

nbusseneau (approved):
Same comments as Timo, I'll approve because I do think the quick fix is sound. I would suggest tracking in an issue if we end up merging this.

bmcustodio:
@tklauser nice one — I have based the new approach on that, with slight changes (cc @ti-mo @jrajahalme).

tklauser (approved):
:rocket:

jrajahalme (approved):
Thanks for fixing this :-)

jrajahalme (approved):

ti-mo (approved):

connectivity/check/test.go
```diff
@@ -11,9 +11,18 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cilium/cilium-cli/defaults"
+	k8sConst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
```
christarazi, line 16: Since `cilium/cilium` is considered 3rd party in this repo, there should be a newline separating the local imports (`cilium/ciluim-cli` pkgs)

```suggestion
	"github.com/cilium/cilium-cli/defaults"
	
	k8sConst "github.com/cilium/cilium/pkg/k8s/apis/cilium.io"
	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
```

connectivity/check/test.go
```diff
@@ -189,6 +198,30 @@ func (t *Test) WithPolicy(policy string) *Test {
...
+				for _, e := range pl[i].Spec.Ingress {
+					for _, es := range e.FromEndpoints {
+						if n, ok := es.MatchLabels[k]; ok && n == defaults.ConnectivityCheckNamespace {
+							es.MatchLabels[k] = t.ctx.params.TestNamespace
+						}
+					}
+				}
+			}
+		}
+	}
```
christarazi, line 223: Just at a glance, this doesn't seem to be the exhaustive list of resources that might need a namespace change or is that not the case?

christarazi (changes requested):

connectivity/check/test.go
bmcustodio: Yes, this just affects the CNPs, but those are the only ones that need the namespace change (at least for now) since they're the only ones sourced from YAML files. Everything else (deployments, etc...) is built in Go code AFAIK.

tklauser:
Removed the `ready-to-merge` label to give @christarazi a chance to review again re. https://github.com/cilium/cilium-cli/pull/330#discussion_r699603373

michi-covalent (approved):

christarazi (approved):
Let's squash the last commit into the first

tklauser:
> Let's squash the last commit into the first

Done and force pushed.

PR #332, opened 2021-06-17
hubble: fix opening Hubble UI URL on platforms other than macOS

### LINKED ISSUE #287: Support opening hubble UI on non-mac platforms
This line uses `open` to open the Hubble UI:

https://github.com/cilium/cilium-cli/blob/309c2a34257bb08e21fed9ef87b499e2d7016121/hubble/ui.go#L338

This works on OSX, but not on other platforms.

On Linux, I believe that we could alternatively try `xdg-open`.

tklauser:
Currently, opening the Hubble UI URL only works on macOS. Use the
github.com/pkg/browser package which provides a more portable way of
opening an URL in a browser. Tested on Linux and macOS.

Fixes #287


rolinh (approved):

joestringer (approved):
Thanks, LGTM.

PR #336, opened 2021-06-18
misc: use hubble command rather than port-forwards

nbusseneau:
Seems like we had missed some of these.

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -82,9 +82,9 @@ jobs:
 
       - name: Relay Port Forward
         run: |
-          pkill -f "kubectl port-forward.*hubble-relay" || true
+          pkill -f "cilium hubble port-forward|kubectl port-forward.*hubble-relay"
           sleep 1s
-          kubectl port-forward -n kube-system deployment/hubble-relay 4245:4245&
+          cilium hubble port-forward&
           sleep 5s
 
       - name: Connectivity test
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ To install Cilium while automatically detected:
 
 ### Hubble
 
-    ./cilium hubble enable
+    cilium hubble enable
     🔑 Generating certificates for Relay...
     2021/01/06 14:40:21 [INFO] generate received request
     2021/01/06 14:40:21 [INFO] received CSR
@@ -132,7 +132,7 @@ To install Cilium while automatically detected:
 
 #### With Flow Validation
 
-    kubectl port-forward -n kube-system deployment/hubble-relay 4245:4245&
+    cilium hubble port-forward&
     cilium connectivity test --single-node
     ⌛ Waiting for deployments to become ready
     🔭 Enabling Hubble telescope...
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -330,7 +330,10 @@ func (ct *ConnectivityTest) enableHubbleClient(ctx context.Context) error {
 	status, err := ct.hubbleClient.ServerStatus(ctx, &observer.ServerStatusRequest{})
 	if err != nil {
 		ct.Warn("Unable to contact Hubble Relay, disabling Hubble telescope and flow validation:", err)
-		ct.Info("Expose Relay locally with: kubectl port-forward -n kube-system deployment/hubble-relay 4245:4245")
+		ct.Info(`Expose Relay locally with:
+   cilium hubble enable
+   cilium status --wait
+   cilium hubble port-forward&`)
 		ct.hubbleClient = nil
 		ct.params.Hubble = false
 


.github/workflows/kind.yaml
```diff
@@ -82,9 +82,9 @@ jobs:
 
       - name: Relay Port Forward
         run: |
-          pkill -f "kubectl port-forward.*hubble-relay" || true
+          pkill -f "cilium hubble port-forward" || true
```
nbusseneau, line 85: I think the `pkill` has failed 🤔 (even though it's not reported as failure due to `|| true`)
```
Error: Unable to port forward: unable to execute "kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80": exit status 1
❌ Unable to execute "kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80":
 Unable to listen on port 4245: Listeners failed to create with the following errors: [unable to create listener: Error listen tcp4 0.0.0.0:4245: bind: address already in use unable to create listener: Error listen tcp6 [::]:4245: bind: address already in use]
error: unable to listen on any of the requested ports: [{4245 4245}]
```

nbusseneau: Actually, why do we even a `|| true`. We should never expect it fail to kill the previous `port-forward`.

nbusseneau: Fixed in latest version.

nbusseneau:
AKS testing https://github.com/cilium/cilium-cli/actions/runs/950610632 failed with:
```
Error: Connectivity test failed: timeout reached waiting for ***&Service***ObjectMeta:***echo-other-node  cilium-test /api/v1/namespaces/cilium-test/services/echo-other-node fd2c3c95-c8b4-48be-b159-1bdfb2041050 1725 0 2021-06-18 17:38:45 +0000 UTC <nil> <nil> map[kind:echo] map[] [] []  [***cilium Update v1 2021-06-18 17:38:45 +0000 UTC FieldsV1 ***"f:metadata":***"f:labels":***".":***,"f:kind":***,"f:spec":***"f:externalTrafficPolicy":***,"f:ports":***".":***,"k:***\"port\":8080,\"protocol\":\"TCP\"***":***".":***,"f:name":***,"f:port":***,"f:protocol":***,"f:targetPort":***,"f:selector":***".":***,"f:name":***,"f:sessionAffinity":***,"f:type":***]***,Spec:ServiceSpec***Ports:[]ServicePort***ServicePort***Name:echo-other-node,Protocol:TCP,Port:8080,TargetPort:***0 8080 ***,NodePort:30597,AppProtocol:nil,***,***,Selector:map[string]string***name: echo-other-node,***,ClusterIP:10.0.171.109,Type:NodePort,ExternalIPs:[],SessionAffinity:None,LoadBalancerIP:,LoadBalancerSourceRanges:[],ExternalName:,ExternalTrafficPolicy:Cluster,HealthCheckNodePort:0,PublishNotReadyAddresses:false,SessionAffinityConfig:nil,IPFamily:nil,TopologyKeys:[],***,Status:ServiceStatus***LoadBalancer:LoadBalancerStatus***Ingress:[]LoadBalancerIngress***,***,***,***
```
I have never seen this one before 🤔

michi-covalent (approved):

nebril (approved):

jibi (approved):

.github/workflows/kind.yaml
pchaigno: > Actually, why do we even have a || true. We should never expect it fail to kill the previous port-forward.

Can't it fail to setup port forwarding the first time, in which case we would fail to kill the process?

christarazi (approved):

michi-covalent:
aks failed with this flake: https://github.com/cilium/cilium-cli/issues/342

.github/workflows/kind.yaml
pchaigno: Related: https://github.com/cilium/cilium/issues/16636.

gandro: I was unaware of this change here - I guess this might conflict with https://github.com/cilium/cilium/pull/16637

gandro: I think `|| true` (or `|| echo "nothing killed`) is needed, because `pkill` will return a non-zero value of the process already exited by itself due to `cilium uninstall` already killing the pod.

gandro: I just discovered that `cilium uninstall` will not cause `kubectl port-forward` to terminate. So I think _not_ using `|| true` might actually be correct here, because if there is no running `port-forward` process, then something went wrong.

pchaigno: I agree, but we should then surface the failure to run `kubectl port-forward` in the first place.

tklauser (approved):

.github/workflows/kind.yaml
nbusseneau: > Can't it fail to setup port forwarding the first time, in which case we would fail to kill the process?

In that case the job will fail at the port-forward step and should not reach the second port-forward step.

> I agree, but we should then surface the failure to run `kubectl port-forward` in the first place.

Can you clarify, I'm not sure what you meant? 🤔

gandro: FWIW, the `kubectl port-forward` failures I observed in `cilium/cilium` didn't surface because we are using `kubectl port-forward &` to run the command in the background. So there is never a non-zero exit code.
And because we are using `flow-validation=warning`, even the tests using the port-forward didn't fail as any Hubble Relay errors were ignored. The failure only finally surfaced when `pkill`/`pgrep` never found any process.

nbusseneau: Ah, if that is the issue then we can just retrieve the `&` exit code and expose it as shell return code.

nbusseneau: I've opened #390 to fix this. It relies on this PR, so I propose we merge it and continue discussion in #390.

pchaigno: I'm not sure I understand the need for a separate PR, but I'm fine with that approach if it helps :shrug:

pchaigno (approved):

pchaigno:
All team review requests are covered and tests are passing. Merging.

.github/workflows/kind.yaml
nbusseneau: I just didn't want to clutter this PR with something unrelated ^^

PR #343, opened 2021-06-18
Release v0.8.1

michi-covalent:


michi-covalent:
filed https://github.com/cilium/cilium-cli/issues/344 to track https://github.com/cilium/cilium-cli/pull/343/checks?check_run_id=2861768000

joestringer (approved):

joestringer:
@michi-covalent and I have triaged the GKE issue, Michi will follow up with the details. Seems like we need better robustness in the CLI handling of nodeport service initialization.

GKE failure is not new, master is failing regularly with about 40% probability:
https://github.com/cilium/cilium-cli/actions/workflows/gke.yaml?query=branch%3Amaster

New CI approach says that CI failures should block merging so that we can create pressure to coordinate and address the issue. However, at the same time the release manager cannot be responsible to fix every single flake, so whether a release PR gets blocked by CI failures becomes a value judgement. If it's important to get the release out, and we have assurance that we are not observing obvious regressions, then it may be better to get the release out to fix the known issues. In this case, minikube install with v0.8.0 is completely broken so we need a new release to fix that.

@michi-covalent and I decided that in this case, we should give until late Monday PT to investigate and mitigate the issue, and if we cannot fix it by then, we will bypass this CI failure to merge & release v0.8.1 to address the outstanding issue.

tklauser (approved):

tklauser:
> @michi-covalent and I decided that in this case, we should give until late Monday PT to investigate and mitigate the issue, and if we cannot fix it by then, we will bypass this CI failure to merge & release v0.8.1 to address the outstanding issue.

@michi-covalent and I discussed this out-of-band this morning CEST and reasoned that #346 should be sufficient to mitigate the flake. Given that that PR was merge and this PR was rebased and now also passed, we should be good to go for the release.

jibi (approved):

Issue #344, opened 2021-06-18
flake: curl to nodeport fails with exit code 7


michi-covalent:
# flake instances

- https://github.com/cilium/cilium-cli/runs/2861768000?check_suite_focus=true
- https://github.com/cilium/cilium-cli/pull/319/checks?check_run_id=2861173337

# symptoms

curl to nodeport (can be either a local or remote nodeport) fails with exit code 7 (Connect failed):
```
  ❌ command "curl -w %***local_ip***:%***local_port*** -> %***remote_ip***:%***remote_port*** = %***response_code*** --silent --fail --show-error --connect-timeout 5 --output /dev/null http://10.168.0.24:32193" failed: command terminated with exit code 7
```
and the subsequent flow validation fails.

nbusseneau:
Hit in #339: https://github.com/cilium/cilium-cli/actions/runs/950721302

PR #346, opened 2021-06-19
check: Wait for nodeports

michi-covalent:
Wait until Cilium implements nodeports on all the nodes before starting
the connectivity test. For now it uses `nc` in a client pod to check if
the nodeports are open.

Fixes: #344


tklauser (approved):
:shipit:

Issue #342, opened 2021-06-18
flake: timeout reached waiting for service (echo-same-node or echo-other-node)


michi-covalent:
# flake instances

- https://github.com/cilium/cilium-cli/pull/331/checks?check_run_id=2861168613
- https://github.com/cilium/cilium-cli/runs/2842079516?check_suite_focus=true#341 
- https://github.com/cilium/cilium-cli/pull/336/checks?check_run_id=2860556736

# symptoms

`cilium connectivity test` times out waiting for echo-same-node or echo-other-node service.

```
⌛ [cilium-cilium-cli-950610632] Waiting for deployments [client client2 echo-same-node] to become ready...
⌛ [cilium-cilium-cli-950610632] Waiting for deployments [echo-other-node] to become ready...
⌛ [cilium-cilium-cli-950610632] Waiting for CiliumEndpoint for pod cilium-test/client-7b7bf54b85-blm9q to appear...
⌛ [cilium-cilium-cli-950610632] Waiting for CiliumEndpoint for pod cilium-test/client2-666976c95b-6zhqj to appear...
⌛ [cilium-cilium-cli-950610632] Waiting for CiliumEndpoint for pod cilium-test/echo-other-node-697d5d69b7-2nnln to appear...
⌛ [cilium-cilium-cli-950610632] Waiting for CiliumEndpoint for pod cilium-test/echo-same-node-7967996674-rt6hg to appear...
⌛ [cilium-cilium-cli-950610632] Waiting for Service cilium-test/echo-other-node to become ready...

[...]

PR #348, opened 2021-06-22
check: Add extra info

michi-covalent:
Print the number of times the validation command got executed before
timeout was reached. I just want to confirm that the command is not
getting stuck for a long time.

Ref: #342


michi-covalent:
```
Error: Connectivity test failed: timeout reached after 8 tries waiting for NodePort 192.168.150.66:31122 (cilium-test/echo-other-node)
```
ok so at least we are trying multiple times...

PR #349, opened 2021-06-22
connectivity: remove Warn(f) from Test and Action

ti-mo:
This proved a bit difficult to use in practice. With output of successful test runs now being hidden in CI, things need to fail to pop up. We cannot have things failing and throwing warnings without anyone noticing.

Failed finalizers will now cause an error. Finalizers are not supposed to fail, at least not silently.

Errors polling the Hubble buffer now also fail the test, but this behaviour is already being removed in the open Hubble streaming PR. (#319)

Note: `ConnectivityTest` still has `Warn(f)`, but it's purely cosmetical and not integrated into the rest of the test suite.

This is an alternative proposal to #327.

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -74,9 +74,6 @@ type Action struct {
 
 	// failed is true when Fail was called on the Action
 	failed bool
-
-	// warned is true when Warn was called on the Action
-	warned bool
 }
 
 func newAction(t *Test, name string, s Scenario, src *Pod, dst TestPeer) *Action {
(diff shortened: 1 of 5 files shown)


twpayne (approved):

michi-covalent:
rebasing to pick up https://github.com/cilium/cilium-cli/pull/356

michi-covalent (approved):

michi-covalent:
hey @jrajahalme could you take a look

tklauser (approved):

michi-covalent:
resolved conflicts

michi-covalent:
ok i'm merging this

PR #350, opened 2021-06-22
ci: fix the go vendoring check

kaworu:
Before this patch, git diff was used to ensure that Go modules were properly vendored. While it could catch a failure to update a vendored module, it would not fail when a module was not vendored (untracked).

This patch uses git status instead of git diff, effectively catching untracked vendored modules as well.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -49,7 +49,7 @@ jobs:
         go mod tidy
         go mod vendor
         go mod verify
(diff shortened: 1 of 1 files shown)


rolinh (approved):

pchaigno (approved):

tklauser:
Removed the "ready to merge" label in order to be able to analyze failing tests before merge.

PR #352, opened 2021-06-22
install: re-enable version validation for non-default images

tklauser:
After commit 03bbc81da1c1 ("install: set the default images at the cli
flag level") the command `cilium install --version foobar` would no
longer hit the version check, while previously such a version would be
rejected (unless a custom agent/operator image was specified):

```
invalid parameters: invalid syntax "foobar" for image tag
```

Re-enabled the version tag check for the default image.

Fixes: 03bbc81da1c1 ("install: set the default images at the cli flag level")

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -1070,7 +1070,7 @@ func (p *Parameters) validate() error {
 
 		p.configOverwrites[t[0]] = t[1]
 	}
-	if p.AgentImage != "" || p.OperatorImage != "" {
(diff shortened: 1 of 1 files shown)


christarazi (approved):

tklauser:
03bbc81 was reverted in b6b99e9f978d2c80f680fa62d9b9aff9b4d3290e. Changing this to draft status.

PR #354, opened 2021-06-22
Revert "install: set the default images at the cli flag level"

michi-covalent:
This reverts commit 03bbc81da1c161b045ed195d0f62c362d142d1b6.

fqOperatorImage() sets the default image based on k.params.DatapathMode
so that utils.BuildImagePath() uses the correct default operator image.

With the commit being reverted, however, the install command never uses
the default operator image because k.params.OperatorImage is always set
to quay.io/cilium/operator-generic.

Ref: https://github.com/cilium/cilium-cli/blob/03bbc81da1c161b045ed195d0f62c362d142d1b6/install/install.go#L1115


michi-covalent:
https://github.com/cilium/cilium-cli/pull/354/checks?check_run_id=2889294353 failed because of this flake https://github.com/cilium/cilium-cli/issues/355

would like to get this merged so that at least the CI job deploys the right operator image.

joestringer (approved):

joestringer:
Looking at the original PR, I don't know why it was merged in the first place with three failing jobs. Happy to merge to get the tree more green. I understand that @michi-covalent is investigating the failure here which was the original issue he was trying to chase down before he found this issue.

PR #359, opened 2021-06-23
install: rename `DeploymentIsReady` to `CheckDeploymentStatus`

tklauser:
From the method name alone one might assume it to return a `bool` rather
than an `error`. Rename the method to avoid this confusion.

This is also in line with the `CheckDaemonSetStatus` method which serves
the same purpose for `DaemonSet`s.


tklauser:
[EKS (ENI) failure](https://github.com/cilium/cilium-cli/pull/359/checks?check_run_id=2894053515) is #355

ldelossa (approved):
LGTM.

michi-covalent (approved):
let's roll

PR #360, opened 2021-06-23
install/aks: set 'enable-ipv4-masquerade' as false

aanm:
This new flag was introduced in Cilium 1.10 and it deprecates the
'masquerade' option. With https://github.com/cilium/cilium/pull/16502
being merged, Cilium-cli was still using the 'masquerade' option which
caused Cilium installation to fail on AKS.

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -27,6 +27,7 @@ import (
 	"github.com/cilium/cilium-cli/internal/k8s"
 	"github.com/cilium/cilium-cli/internal/utils"
 	"github.com/cilium/cilium-cli/status"
+	"github.com/cilium/cilium/pkg/versioncheck"
 
 	"github.com/cilium/cilium/api/v1/models"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
@@ -1316,7 +1317,19 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 		m.Data["enable-endpoint-routes"] = "true"
 		m.Data["auto-create-cilium-node-resource"] = "true"
 		m.Data["enable-local-node-route"] = "false"
-		m.Data["masquerade"] = "false"
+
+		ersion := strings.TrimPrefix(k.params.Version, "v")
+		v, err := versioncheck.Version(ersion)
+		if err != nil {
+			k.Log("Unable to parse the provided version %q, assuming it's >= 1.10.0", k.params.Version)
+			v = versioncheck.MustVersion("1.10.0")
+		}
+		if v.GTE(versioncheck.MustVersion("1.10.0")) {
+			m.Data["enable-ipv4-masquerade"] = "false"
+		} else {
+			// Deprecated. Remove once we stop supporting Cilium 1.10
+			m.Data["masquerade"] = "false"
+		}
 		m.Data["enable-bpf-masquerade"] = "false"
 	}
 


pchaigno (approved):

tklauser (approved):

christarazi (approved):

pchaigno:
:scream: Why did MLH add the `ready-to-merge` label?? Several Required jobs are failing.

michi-covalent:
not sure why CI jobs are not getting triggered. rebasing

install/install.go
```diff
@@ -1316,7 +1317,17 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 		m.Data["enable-endpoint-routes"] = "true"
 		m.Data["auto-create-cilium-node-resource"] = "true"
 		m.Data["enable-local-node-route"] = "false"
-		m.Data["masquerade"] = "false"
+
+		if v, err := semver.Parse(k.params.Version); err != nil {
+			return nil, fmt.Errorf("unable to parse version %q: %w", k.params.Version, err)
+		} else {
+			if v.GTE(semver.MustParse("v1.10.0")) {
```
michi-covalent, line 1324: doesn't this panic? `v1.10.0` is not a semver 🤔 
```
panic: semver: Parse(v1.10.0): Invalid character(s) found in major number "v1" [recovered]
	panic: semver: Parse(v1.10.0): Invalid character(s) found in major number "v1"
```

michi-covalent (approved):

michi-covalent:
lint complaining
```
  Warning: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (revive)
```

install/install.go
```diff
@@ -1316,7 +1317,18 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 		m.Data["enable-endpoint-routes"] = "true"
 		m.Data["auto-create-cilium-node-resource"] = "true"
 		m.Data["enable-local-node-route"] = "false"
-		m.Data["masquerade"] = "false"
+
+		ersion := strings.TrimPrefix(k.params.Version, "v")
+		if v, err := semver.Parse(ersion); err != nil {
+			return nil, fmt.Errorf("unable to parse version %q: %w", k.params.Version, err)
```
tklauser, line 1323: I think this will error out in case users provides their own custom version, e.g. a tag during developent. Could we just assume >=1.10.0 in this case and log a warning?

install/install.go
```diff
@@ -1316,7 +1317,18 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 		m.Data["enable-endpoint-routes"] = "true"
 		m.Data["auto-create-cilium-node-resource"] = "true"
 		m.Data["enable-local-node-route"] = "false"
-		m.Data["masquerade"] = "false"
+
+		ersion := strings.TrimPrefix(k.params.Version, "v")
+		if v, err := semver.Parse(ersion); err != nil {
```
tklauser, line 1322: Other code in this package uses `github.com/cilium/cilium/pkg/versioncheck` for these kind of checks. Would that be appropriate here as well?

If not, we could at least use `github.com/blang/semver/v4` instead of `github.com/blang/semver` which is already vendored as a transitive dependency (once #358 is merged, that is).

michi-covalent: 👍 for github.com/cilium/cilium/pkg/versioncheck

install/install.go
michi-covalent: i'm a bit torn about this one. warning might not be sufficient in this case, cilium might not function properly if we can't configure for a specific version 🤔

michi-covalent:
relay stayed in pending state: https://github.com/cilium/cilium-cli/pull/360/checks?check_run_id=2901439071
```
            "status": {
                "conditions": [
                    {
                        "lastProbeTime": null,
                        "lastTransitionTime": "2021-06-24T04:21:15Z",
                        "message": "0/2 nodes are available: 2 node(s) had taint {node.kubernetes.io/network-unavailable: }, that the pod didn't tolerate.",
                        "reason": "Unschedulable",
                        "status": "False",
                        "type": "PodScheduled"
                    }
                ],
                "phase": "Pending",
                "qosClass": "BestEffort"
            }
```

install/install.go
tklauser: Maybe we could just perform the check if the user didn't specifiy a custom agent and/or operator image, like in other places in the code. And assume >= 1.10 in case a custom agent/operator image was specified.

Otherwise I think something like `cilium install --version latest --agent-image quay.io/cilium/cilium-ci --operator-image quay.io/cilium/operator-azure-ci` would no longer work to test the lastest `master` build?

install/install.go
```diff
@@ -1316,7 +1317,19 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 		m.Data["enable-endpoint-routes"] = "true"
 		m.Data["auto-create-cilium-node-resource"] = "true"
 		m.Data["enable-local-node-route"] = "false"
-		m.Data["masquerade"] = "false"
+
+		ersion := strings.TrimPrefix(k.params.Version, "v")
```
jibi, line 1321: (shouldn't this be `version`?)

tklauser (approved):

PR #335, opened 2021-06-18
hubble: Wait for Cilium to start before deploying relay/ui

michi-covalent:
Hubble Relay and UI do not function properly if they are not managed by
Cilium. This commit modifies `cilium hubble enable` command to wait for
Cilium pods to become ready after the restart before deploying them.

Fixes #334

Signed-off-by: Michi Mutsuzaki <michi@isovalent.com>

PR #362, opened 2021-06-24
.github: enable Hubble relay in external workloads test

tklauser:
Currently the external workloads test doesn't use Hubble. Enable it
during installation, which will help to debug failures happening during
the test.

### DIFF
--- a/.github/in-cluster-test-scripts/external-workloads-install.sh
+++ b/.github/in-cluster-test-scripts/external-workloads-install.sh
@@ -11,6 +11,9 @@ cilium install \
   --kube-proxy-replacement=strict \
   --native-routing-cidr="${CLUSTER_CIDR}"
 
+# Enable Relay
+cilium hubble enable
+
 # Wait for Cilium status to be ready
 cilium status --wait
 
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![GKE](https://github.com/cilium/cilium-cli/workflows/GKE/badge.svg)](https://github.com/cilium/cilium-cli/actions?query=workflow%3AGKE)
 [![AKS](https://github.com/cilium/cilium-cli/workflows/AKS/badge.svg)](https://github.com/cilium/cilium-cli/actions?query=workflow%3AAKS)
 [![Multicluster](https://github.com/cilium/cilium-cli/workflows/Multicluster/badge.svg)](https://github.com/cilium/cilium-cli/actions?query=workflow%3AMulticluster)
+[![External Workloads](https://github.com/cilium/cilium-cli/actions/workflows/externalworkloads.yaml/badge.svg)](https://github.com/cilium/cilium-cli/actions/workflows/externalworkloads.yaml)
 
 ## Installation
 


tklauser:
This should also help to prevent issues like these where apparently Hubble relay didn't come up in time during connectivity tests: https://github.com/cilium/cilium-cli/pull/336/checks?check_run_id=2902370072

nebril (approved):

.github/in-cluster-test-scripts/external-workloads-install.sh
```diff
@@ -11,6 +11,9 @@ cilium install \
   --kube-proxy-replacement=strict \
   --native-routing-cidr="${CLUSTER_CIDR}"
 
+# Enable Relay
+cilium hubble enable
```
nebril, line 15: Shouldn't we move this below `cilium status --wait`? We had an issue caused by Cilium pods not coming up fast enough when pods created by `cilium clustermesh enable` were not managed by Cilium.

I am not sure whether Relay pod needs to be managed by Cilium, but I assume that that's the case.

nebril (changes requested):

.github/in-cluster-test-scripts/external-workloads-install.sh
tklauser: We seem to do it in this order (i.e. `cilium hubble enable`, then `cilium status --wait`) in most other test jobs. Presumably because we want to wait for Hubble relay to become available and managed by Cilium. I'm not sure I understand the issue with `cilium clustermesh enable`. I see that `cilium status --wait` appears before we enable ClusterMesh, so we should be safe in that regard. Or am I missing something?

tklauser: Regarding having `cilium status --wait` really wait for Hubble relay to be available also see #365 which should help in cases where `cilium status --wait` would return even though Hubble relay wasn't fully available, i.e. still reporting warnings.

gandro: I think we might need to have two `cilium status --wait`. As @nebril pointed out correctly, it is important that `clustermesh-apiserver` and `hubble-relay` are managed by Cilium, meaning the CNI needs to be installed properly before the deployment is started. So I think it does make sense to first have a `cilium status --wait` just for Cilium, then do a `cilium hubble/clustermesh enable`, and then do another `cilium status --wait` (with #365 applied) to ensure the restarted cilium pods, clustermesh-apiserver and hubble-relay are all ready to go.

tklauser: Thanks, I'ill add a second `cilium status --wait` before `cilium hubble enable` in that case. I think the `clustermesh-apiserver` readiness check will be taken care of by `cilium clustermesh status --wait` further down, also see commit c171080ef434.

michi-covalent: fyi `cilium hubble enable` restarts cilium and wait for it to become ready before deploying hubble relay / ui: https://github.com/cilium/cilium-cli/pull/335 so the first `cilium status --wait` is probably not strictly required.

michi-covalent (approved):

.github/in-cluster-test-scripts/external-workloads-install.sh
tklauser: Thanks. I've checked that `cilium hubble enable` will indeed wait for Cilium to become fully ready before proceeding:

https://github.com/cilium/cilium-cli/blob/619f9d5a7b799e3a43ec6af42ac7a3f3e1d83c8d/hubble/hubble.go#L277-L282

So the first `cilium status --wait` is indeed not necessary. I've dropped it again.

michi-covalent (approved):

gandro (approved):

PR #365, opened 2021-06-24
status: Do not ignore warnings by default

gandro:
This is an attempt to fix cilium/cilium#16636 - in the linked issue, `cilium status --wait` did continue even though there were still Hubble related warnings as the server was still initializing:

```
Run cilium status --wait
    /¯¯\
 /¯¯\__/¯¯\    Cilium:         1 warnings
 \__/¯¯\__/    Operator:       OK
 /¯¯\__/¯¯\    Hubble:         1 warnings
 \__/¯¯\__/    ClusterMesh:    disabled
    \__/

DaemonSet         cilium             Desired: 2, Ready: 2/2, Available: 2/2
Deployment        cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
Deployment        hubble-relay       Unavailable: 1/0
Containers:       cilium             Running: 2
                  cilium-operator    Running: 1
                  hubble-relay       
Image versions    cilium             quay.io/cilium/cilium-ci:7f20b3abd6837913f0deff225c301bcc11d3da3f: 2
                  cilium-operator    quay.io/cilium/operator-generic-ci:7f20b3abd6837913f0deff225c301bcc11d3da3f: 1
Warnings:         cilium             cilium-gnjq2    Hubble: Server not initialized
                  hubble-relay       hubble-relay    1 pods of DaemonSet hubble-relay are not available
```

This caused issues later on in the CI run, as Hubble Relay was clearly not ready.

This PR adds therefore adds a new **disabled-by-default** CLI flags to `cilium status`:

- `--ignore-warnings` which requires that there are no warnings before `cilium status --wait` returns.

This flags should help us to not continue a CI run if there are warnings (such as e.g. Hubble not being ready yet).

If users want to restore the old behavior, they can use `cilium status --wait --ignore-warnings`, which will return even if there are warnings. 

There is also a drive by fix to avoid calling the `hubble-relay` Deployment wrongly a DaemonSet.

### DIFF
--- a/internal/cli/cmd/status.go
+++ b/internal/cli/cmd/status.go
@@ -48,10 +48,10 @@ func newCmdStatus() *cobra.Command {
 			return err
 		},
 	}
-
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
-	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait for status to report success (no errors)")
+	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait for status to report success (no errors and warnings)")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
+	cmd.Flags().BoolVar(&params.IgnoreWarnings, "ignore-warnings", false, "Ignore warnings when waiting for status to report success")
 	cmd.Flags().StringVar(&contextName, "context", "", "Kubernetes configuration context")
 
 	return cmd
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -33,10 +33,18 @@ import (
 var retryInterval = 2 * time.Second
 
 type K8sStatusParameters struct {
-	Namespace       string
-	Wait            bool
-	WaitDuration    time.Duration
+	Namespace    string
+	Wait         bool
+	WaitDuration time.Duration
+	// WarningFreePods specifies a list of pods which are required to be
+	// warning free. This takes precedence over IgnoreWarnings and is only
+	// used if Wait is true.
 	WarningFreePods []string
+	// IgnoreWarnings will, if set to true, ignore any warnings on pods to
+	// determine the readiness. This is only used if Wait is true and
+	// WarningFreePods is empty. If WarningFreePods is non-empty, the value
+	// of this flag is meaningless.
+	IgnoreWarnings bool
 }
 
 type K8sStatusCollector struct {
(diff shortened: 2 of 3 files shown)


tklauser (commented):
Looking at how we currently use `cilium status --wait` in CI (i.e. usually after we ran `cilium hubble enable`, indicating that we actually want for the Hubble relay deployment to be ready), we might even want to make sure Hubble is warning-free in `cilium status --wait` by default.

Or are there legitimate cases where the Hubble relay deployment has warnings and we still consider it to be usable?

gandro:
> Or are there legitimate cases where the Hubble relay deployment has warnings and we still consider it to be usable?

Honestly, I don't think so. I also just added a new `--no-warnings` flags that requires no warnings for all pods (instead of requiring the user to specifying a list of pods). We could consider making `--no-warnings=true` by default, as I agree there is probably no reason to ignore warnings. What do you think?

tklauser:
> Honestly, I don't think so. I also just added a new --no-warnings flags that requires no warnings for all pods (instead of requiring the user to specifying a list of warning free pods). We could consider making --no-warnings=true by default, as I agree there is probably no reason to ignore warnings. What do you think?

Thanks. Introducing `--no-warnings` instead of a list of warning free pods and setting it to `true` by default sounds like a good solution to me. I imagine that for a user, having to compile a list of warning free pods might be a bit tedious.

gandro:
> Thanks. Introducing `--no-warnings` instead of a list of warning free pods and setting it to `true` by default sounds like a good solution to me. I imagine that for a user, having to compile a list of warning free pods might be a bit tedious.

Fully agreed. I went for the `warn-free-pods` solution first because it was easier to implement. But is clearly has worse UX. I think I'll remove it again. I'll rework the PR to make `--no-warnings` (or maybe invert it, and call it `--ignore-warnings=false`) the default.

gandro:
As discussed above, reworked the PR to contain a `--ignore-warnings` flag which is **enabled by default**.

tklauser (approved):
Thanks!

tklauser:
Not sure why the [GKE job](https://github.com/cilium/cilium-cli/pull/365/checks?check_run_id=2904654177) failed, logs show nothing indicating an error :thinking:

ldelossa (approved):
lgtm

gandro:
Did some more investigation of the GKE timeout. It's really not clear why it timed out, there is nothing in the logs that indicates a broken state. At ~2021-06-24T12:25Z there was a 12 minute pause where simply nothing happened for no apparent reason.

Rebasing to pull in d8bfe029c52f9726d7cac56bde75f81f9e6b1c3a - which might help in debugging this further.

PR #366, opened 2021-06-24
workflows: Add `--timestamps` to `kubectl logs`

gandro:
Timestamps are very useful when investigating a unknown CI failure. An
example where timestamps would yield us more information is for example
when a job runs into a timeout, such as e.g.
https://github.com/cilium/cilium-cli/runs/2904654177 where we suspect
that it might have run a bit longer than usual, but it's hard to tell.

Signed-off-by: Sebastian Wicki <sebastian@isovalent.com>

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -110,8 +110,8 @@ jobs:
       - name: Post-test information gathering
         if: ${{ failure() }}
         run: |
-          kubectl logs -n kube-system job/cilium-cli-install
-          kubectl logs -n kube-system job/cilium-cli
+          kubectl logs --timestamps -n kube-system job/cilium-cli-install
+          kubectl logs --timestamps -n kube-system job/cilium-cli
           kubectl get pods --all-namespaces -o wide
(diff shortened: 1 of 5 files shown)


pchaigno (approved):
:shipit:

tklauser (approved):

tklauser:
The [external workloads failure](https://github.com/cilium/cilium-cli/pull/366/checks?check_run_id=2905226413) is a known flake: #329.

nebril (approved):

gandro:
EKS seems to have hit a new flake: #367

gandro:
Kind _also_ seems to have hit a variant of flake #367. https://github.com/cilium/cilium-cli/pull/366/checks?check_run_id=2905226507

 In both cases, we see UDP response packets for the DNS request, but no outgoing TCP connection. Because we don't collect L7 logs, it's not clear what the DNS request was.

tklauser:
Rebased to pick up c171080ef434d7a65b0cc1ddcd7a2ee82f2bd34e hopefully helping to avoid some flakes.

tklauser:
The [Kind failure](https://github.com/cilium/cilium-cli/pull/366/checks?check_run_id=2912968953) looks like another instance of #367.

Issue #367, opened 2021-06-24
flake: `allow-all/pod-to-world/http(s)-to-cilium-io` fails


gandro:
Observed in https://github.com/cilium/cilium-cli/pull/366 on EKS:

https://github.com/cilium/cilium-cli/pull/366/checks?check_run_id=2905226287

The flow logs indicate that a DNS request was made, but no TCP connection was ever established:

```
   [.] Action [allow-all/pod-to-world/http-to-www-google: cilium-test/client-7b7bf54b85-h6qvt (10.0.1.205) -> www-google-http (www.google.com:80)]
  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --connect-timeout 5 --output /dev/null http://www.google.com:80" failed: command terminated with exit code 28
  📄 Matching flows for pod cilium-test/client-7b7bf54b85-h6qvt
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ℹ️  SYN and(ip(src=10.0.1.205),tcp(dstPort=80),tcpflags(syn)) not found
  ⌛ Waiting (5s) for flows: Required flows not found yet
[...]

PR #369, opened 2021-06-25
test: Add DNS visibility annotations to client deployments

michi-covalent:
Bring back the DNS visibility annotation (466e8f1) that got reverted in
PR #356. I reverted it because it depended on 07a161d which was causing
issue #355. I updated the commit so that it no longer depends on 07a161d.
Having DNS visibility might give us some additional info to debug #367.

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -49,6 +49,7 @@ type deploymentParameters struct {
 	Affinity       *corev1.Affinity
 	ReadinessProbe *corev1.Probe
 	Labels         map[string]string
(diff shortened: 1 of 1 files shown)


michi-covalent:
guess this doesn't work for test cases that apply cilium network policies to client pods 😞 

https://docs.cilium.io/en/v1.10/policy/visibility/#limitations
```
Visibility annotations do not apply if rules are imported which select the pod which is annotated.
```

michi-covalent:
didn't work as expected, this only works when there is no other policy. will revisit.

PR #371, opened 2021-06-25
Add helpful success message to select subcommands

Mierdin:
This PR fixes a fairly minor aesthetic issue when running some `cilium` sub-commands; there is no explicit confirmation that an operation finished successfully. In some cases, one may safely assume a successful operation in the absence of any error messages, but I think this makes it more obvious, and gives the opportunity to indicate a next step, especially for the more common operations like `install`.

I perused the `cilium` help output and identified a few candidates where a message like this might be helpful. I left some out because I felt it wasn't necessary (those that shell out to `kubectl port-forward` for example) but if a reviewer feels this list is incomplete, I'm happy to add to it. Same applies to any call to actions I may have missed an opportunity to include.

- [x] `cilium install/uninstall`
- [x] `cilium hubble enable/disable`
- [x] `cilium clustermesh enable/disable/connect/disconnect`
- [x] `cilium clustermesh external-workload create/delete`

I also fixed a typo in `newCmdExternalWorkloadInstall()` because I was there, and I saw it. :smile:

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -544,6 +544,8 @@ func (k *K8sClusterMesh) Disable(ctx context.Context) error {
 
 	k.deleteCertificates(ctx)
 
+	k.Log("✅ ClusterMesh disabled.")
+
 	return nil
 }
 
@@ -613,6 +615,8 @@ func (k *K8sClusterMesh) Enable(ctx context.Context) error {
 	if _, err := k.client.CreateService(ctx, k.params.Namespace, svc, metav1.CreateOptions{}); err != nil {
 		return err
 	}
+
+	k.Log("✅ ClusterMesh enabled!")
 
 	return nil
 }
@@ -909,7 +913,13 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
 	}
 
 	k.Log("✨ Connecting cluster %s -> %s...", remoteCluster.ClusterName(), k.client.ClusterName())
-	return k.patchConfig(ctx, remoteCluster, aiLocal)
+	if err := k.patchConfig(ctx, remoteCluster, aiLocal); err != nil {
+		return err
+	}
+
+	k.Log("✅ Connected cluster %s and %s!", k.client.ClusterName(), remoteCluster.ClusterName())
+
+	return nil
 }
 
 func (k *K8sClusterMesh) disconnectCluster(ctx context.Context, src, dst k8sClusterMeshImplementation) error {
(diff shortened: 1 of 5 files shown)


michi-covalent:
@nbusseneau i guess the aks job still doesn't work from forked repos?

Mierdin:
The WF trigger would likely need to be changed to `pull_request_target` instead of `pull_request` - but looking at https://github.com/cilium/cilium-cli/pull/212 it appears this was intentionally not updated for reasons related to secrets usage.

nbusseneau:
> @nbusseneau i guess the aks job still doesn't work from forked repos?

Nope, haven't had the time to work on that yet... so this failure is expected (no secret available):
```
Run azure/login@77f1b2e3fb80c0e8645114159d17008b8a2e475a
Error: Az CLI Login failed. Please check the credentials. For more information refer https://aka.ms/create-secrets-for-GitHub-workflows
Error: Error: Input required and not supplied: creds
```

michi-covalent:
thanks @nbusseneau 

@Mierdin i guess this is ready for review then? or are you planning on adding success messages to other sub-commands you listed in this PR?

Mierdin:
@michi-covalent I am happy to add as many as I listed in the description if that's desired. I came up with that list after a short perusal of the help output, but happy to add/remove as necessary.

I'll take a stab at the list above and mark ready for review.

michi-covalent (approved):

ldelossa (approved):
IMO - it would be better to have the logging package have specific methods which attach those emojis to a plan string. 

For example..

```
k.LogSuccess("here's a success message") 
``` 
and the LogSuccess call adds the check mark emoji. 

Approving this, but the use of specific emojis for symbols may drift as unrelated pieces of code start utilizing different emojis unintentionally. Thoughts @michi-covalent ?

Mierdin:
I like that approach, but confused on the reference to a `logging` package. The `Log` methods I used as inspiration for my change are implemented in each subcommand's package:

https://github.com/cilium/cilium-cli/blob/54ab891e0f80c38a0bb166cd7b483fbee2941f19/install/install.go#L1153-L1155

https://github.com/cilium/cilium-cli/blob/54ab891e0f80c38a0bb166cd7b483fbee2941f19/clustermesh/clustermesh.go#L488-L490

Would you like me to add a `LogSuccess` method to each of these types which does what you describe, and then I'll use that instead? Or if I'm missing a more centralized approach let me know.

ldelossa:
@Mierdin yes I see, I was under the impression a logging package was being used across the board. I this concern branches from the point of this PR, but is a conversation worth having as this code base grows.

Mierdin:
@ldelossa Sounds good to me - I took the liberty of creating https://github.com/cilium/cilium-cli/issues/388 to track that; I agree it's worth taking on now before things grow too much more.

Looks like I can't merge myself, so I'll leave it to you or @michi-covalent - cheers!

Issue #370, opened 2021-06-25
Connectivity test failed: services "echo-same-node" is forbidden


aditighag:
Hit this issue on GKE. 

```
Run cilium connectivity test --flow-validation=warning
✨ [gke_***_us-west2-a_cilium-cilium-971904670] Deploying echo-same-node service...

Error: Connectivity test failed: services "echo-same-node" is forbidden: unable to create new content in namespace cilium-test because it is being terminated
Error: Process completed with exit code 1.

```
Also, reported by @michi-covalent .

https://github.com/cilium/cilium/runs/2915516477?check_suite_focus=true

[cilium-sysdump-out.zip (5).zip](https://github.com/cilium/cilium-cli/files/6718024/cilium-sysdump-out.zip.5.zip)

PR #372, opened 2021-06-25
uninstall: Wait for the test namespace to be terminated

michi-covalent:
Wait for the test namespace to be terminated when the `--wait` flag is
specified to avoid the situation where the next connectivity check fails
to deploy resources because the test namespace is in Terminating state.

Ref: #370

### DIFF
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -115,6 +115,16 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 			time.Sleep(retryInterval)
 			goto retry
 		}
+
+		k.Log("⌛ Waiting for %s namespace to be terminated...", k.params.TestNamespace)
+	retryNamespace:
+		// Wait for the test namespace to be terminated. Subsequent connectivity checks would fail
+		// if the test namespace is in Terminating state.
+		_, err = k.client.GetNamespace(ctx, k.params.TestNamespace, metav1.GetOptions{})
+		if err == nil {
+			time.Sleep(retryInterval)
+			goto retryNamespace
+		}
 	}
 
 	return nil


install/uninstall.go
```diff
@@ -115,6 +115,16 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 			time.Sleep(retryInterval)
 			goto retry
 		}
+
+		k.Log("⌛ Waiting for %s namespace to be terminated...", k.params.TestNamespace)
+	retryNamespace:
+		// Wait for the test namespace to be terminated. Subsequent connectivity checks would fail
+		// if the test namespace is in Terminating state.
+		_, err = k.client.GetNamespace(ctx, k.params.TestNamespace, metav1.GetOptions{})
```
aditighag, line 123: Hmm... Instead of the busy wait loop, I wonder if we can use the `wait` functionality exposed by kubectl - https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait.

>Alternatively, the command can wait for the given set of resources to be deleted by providing the "delete" keyword as the value to the --for flag.

aditighag (changes requested):
Fastest turnaround time 🚀

install/uninstall.go
michi-covalent: agreed it's better to use watch or something instead. do you want to tackle that in a separate PR? right now we do busy wait everywhere: https://github.com/cilium/cilium-cli/blob/master/connectivity/check/deployment.go would be better to get all fixed in a dedicated PR.

aditighag: Ok, I've filed - https://github.com/cilium/cilium-cli/issues/374.

aditighag (approved):

michi-covalent:
rebasing to pick up https://github.com/cilium/cilium-cli/pull/319

PR #375, opened 2021-06-28
clustermesh: remove always true `(*K8sClusterMesh).Status log bool` parameter

tklauser:
`(*K8sClusterMesh).Status` is only called from `newCmdClusterMeshStatus`
which sets `log bool` to `true`. Remove the parameter, also from the called
methods where appropriate.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -985,11 +985,9 @@ retry:
 	return ai, err
 }
 
-func (k *K8sClusterMesh) statusService(ctx context.Context, log bool) (*corev1.Service, error) {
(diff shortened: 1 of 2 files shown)


ldelossa (approved):
This looks better.

The conditional logging should be implemented via the logging package, based on the severity of the log, is that occuring?

tklauser:
> The conditional logging should be implemented via the logging package, based on the severity of the log, is that occuring?

Currently, we don't use a logging library, but this will be the logical next step.

ldelossa:
> > The conditional logging should be implemented via the logging package, based on the severity of the log, is that occuring?
> 
> Currently, we don't use a logging library, but this will be the logical next step.

Makes sense to me.

PR #380, opened 2021-06-28
Set the default code owner to @cilium/cli

michi-covalent:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,4 @@
 # Code owners groups and a brief description of their areas:
-# @cilium/janitors           Catch-all for code not otherwise owned
 # @cilium/azure              Integration with Azure
 # @cilium/ci-structure       Continuous integration, testing
 # @cilium/cli                Commandline interfaces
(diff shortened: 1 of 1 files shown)


aditighag (approved):

CODEOWNERS
```diff
@@ -12,7 +12,7 @@
 # The following filepaths should be sorted so that more specific paths occur
 # after the less specific paths, otherwise the ownership for the specific paths
 # is not properly picked up in Github.
-* @cilium/janitors
+* @cilium/cli
```
tklauser, line 15: Agree with the change itself. It looks like we can also remove the comment describing `@cilium/janitors` at the top of the file, given that this group is no longer referenced in `CODEOWNERS` (I cannot comment on the actual line, it seems).

michi-covalent: 👍 will update. let's see if the CI passes again 😄

tklauser: Let's continue your lucky streak with 🟢  CI 😄

tklauser (approved):

PR #391, opened 2021-07-01
workflows: enhance `aws-node` DaemonSet testing

nbusseneau:
In 619f9d5a7b799e3a43ec6af42ac7a3f3e1d83c8d, we improved the way the CLI handles the `aws-node` DaemonSet by using a node selector to evict its pods, allowing us to simply remove the selector on uninstall for a transparent user experience.

We build upon the existing workflow testing:
- Move up `aws-node` number of pods scheduled check right after install and node group have been created, which should be the earliest point at which we can verify if patching the node selector worked.
- Add an additional test at the end to uninstall Cilium and check that backtracking worked on the node selector patching.
- Add same testing to the `eks-tunnel` workflow. Because why not.

This is a follow-up to #208.

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -92,6 +92,10 @@ jobs:
             --spot \
             --node-private-networking
 
+      - name: Make sure the 'aws-node' DaemonSet exists but has no scheduled pods
+        run: |
+          [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 0 ]]
+
       - name: Load cilium cli script in configmap
         run: |
           kubectl create configmap cilium-cli-test-script -n kube-system --from-file=in-cluster-test-script.sh=.github/in-cluster-test-scripts/eks-tunnel.sh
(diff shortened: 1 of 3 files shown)


.github/workflows/eks-tunnel.yaml
```diff
@@ -117,6 +121,12 @@ jobs:
           python cilium-sysdump.zip --output cilium-sysdump-out
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
+      - name: Uninstall and make sure the 'aws-node' DaemonSet has been rescheduled
+        if: ${{ success() }}
+        run: |
+          cilium uninstall --wait
+          [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 2 ]]
```
bmcustodio, line 128: I'm afraid that (i) this may be flaky as the `aws-node` pods may take variable amount of time to be scheduled and (ii) one day we may bump the number of nodes and forget this. Would it make sense for us to (i) wait until this equals (ii) the number of nodes?

.github/workflows/eks-tunnel.yaml
```diff
@@ -92,6 +92,10 @@ jobs:
             --spot \
             --node-private-networking
 
+      - name: Make sure the 'aws-node' DaemonSet exists but has no scheduled pods
+        run: |
+          [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 0 ]]
```
bmcustodio, line 97: :+1:

.github/workflows/eks.yaml
```diff
@@ -121,6 +121,12 @@ jobs:
           python cilium-sysdump.zip --output cilium-sysdump-out
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
+      - name: Uninstall and make sure the 'aws-node' DaemonSet has been rescheduled
+        if: ${{ success() }}
+        run: |
+          cilium uninstall --wait
+          [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 2 ]]
```
bmcustodio, line 128: Same remark as above.

aanm (approved):

.github/workflows/eks-tunnel.yaml
nbusseneau: Replaced with a new version where we look for the annotation on the DaemonSet and check that is has been properly removed.

nbusseneau:
`Uninstall and make sure the 'aws-node' DaemonSet blocking nodeSelector was removed` step in EKS tests failed with:
```
/home/runner/work/_temp/e83fa01f-d794-4e5a-9e8b-d4b2b33cbfad.sh: line 1: cilium: command not found
Error: Process completed with exit code 127.
```
I forgot that `cilium` is only available in-cluster 🤦🏻
Will change approach.

nbusseneau:
New approach worked on both EKS tests:
- https://github.com/cilium/cilium-cli/actions/runs/1001572367
- https://github.com/cilium/cilium-cli/actions/runs/1001572368

Removing temporary commit.

bmcustodio (approved):
LGTM!

tklauser (approved):

PR #397, opened 2021-07-05
install: set `enable-ipv4-masquerade` by default on ≥ Cilium 1.10.0

tklauser:
On Cilium ≥ 1.10.0, use the `enable-ipv4-masquerade` option rather than
`masquerade` which was used on 1.9.x and earlier.


aanm (approved):

tklauser:
Multicluster workflow failed on #361 https://github.com/cilium/cilium-cli/runs/2990063487?check_suite_focus=true

twpayne (approved):

PR #398, opened 2021-07-05
hubble: Auto-detect Relay version 

gandro:
This PR auto-detects the Relay image version when invoking `cilium hubble enable`.
It uses the exact same mechanism as `cilium clustermesh enable`, as both
 `clustermesh-apiserver` and `hubble-relay` share the version number with Cilium.

With this change, users who install a non-default Cilium version via
`cilium install --version v1.x.y` will not have to pass
`--relay-version` to `cilium hubble enable` anymore, as the recommended
version is auto-detected.

### DIFF
--- a/internal/cli/cmd/hubble.go
+++ b/internal/cli/cmd/hubble.go
@@ -19,7 +19,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/hubble"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


internal/cli/cmd/hubble.go
```diff
@@ -62,8 +61,8 @@ func newCmdHubbleEnable() *cobra.Command {
 
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
 	cmd.Flags().BoolVar(&params.Relay, "relay", true, "Deploy Hubble Relay")
-	cmd.Flags().StringVar(&params.RelayImage, "relay-image", defaults.RelayImage, "Image path to use for Relay")
```
gandro, line 65: Reviewer note: The `defaults.RelayImage` is still passed to `BuildImagePath` above and therefore will be used if this flag is absent.

rolinh (approved):
Thanks, you beat me to it :rocket:

gandro:
golangci-lint was not happy with `GetRunningCiliumVersion` an unconditional return for a `for` loop, replaced it with an if-statement.

tklauser (approved):

PR #400, opened 2021-07-05
.github: create clusters with taints

aanm:
As recommended to users, cilium-cli should also set the clusters with
taints. This guarantees that Cilium is properly configured and
application pods can start successfully.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -68,21 +68,45 @@ jobs:
           echo ::set-output name=owner::${OWNER}
 
       - name: Create AKS cluster
+        id: cluster-creation
         run: |
+          # Create group
           az group create \
             --name ${{ env.name }} \
             --location ${{ env.location }} \
             --tags usage=${{ github.repository_owner }}-${{ github.event.repository.name }} owner=${{ steps.vars.outputs.owner }}
+
+          # Create cluster with a 1 node-count (we will remove this node pool
+          # afterwards)
+          # Details: Basic load balancers are not supported with multiple node
+          # pools. Create a cluster with standard load balancer selected to use
+          # multiple node pools, learn more at https://aka.ms/aks/nodepools.
           az aks create \
             --resource-group ${{ env.name }} \
             --name ${{ env.name }} \
             --location ${{ env.location }} \
             --network-plugin azure \
+            --node-count 1 \
+            --load-balancer-sku standard \
+            --node-vm-size Standard_B2s \
+            --node-osdisk-size 30 \
+            --generate-ssh-keys
+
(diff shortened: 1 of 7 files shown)


joestringer (approved):
Looks fine for @cilium/github-sec . I'll let folks with more understanding of Azure comment on those bits.

.github/workflows/eks.yaml
```diff
@@ -57,10 +57,34 @@ jobs:
 
       - name: Create EKS cluster without nodegroup
```
nbusseneau, line 58: Likewise.

.github/workflows/aks.yaml
```diff
@@ -68,21 +68,43 @@ jobs:
...
+            --name nodepool2 \
+            --cluster-name ${{ env.name }} \
+            --resource-group ${{ env.name }} \
             --node-count 2 \
             --node-vm-size Standard_B2s \
             --node-osdisk-size 30 \
-            --load-balancer-sku basic \
-            --generate-ssh-keys
+            --mode system \
+            --node-taints node.cilium.io/agent-not-ready=true:NoSchedule
```
nbusseneau, line 107:
```suggestion
            --priority Spot \
            --eviction-policy Delete \
            --spot-max-price -1 \
            --mode system \
            --node-taints node.cilium.io/agent-not-ready=true:NoSchedule
```
Can we try adding these in? Now that we are using the `standard` load balancer and an additional nodepool anyway, let's try to use spot instances.

defaults/defaults.go
```diff
@@ -76,7 +76,7 @@ const (
 	ConnectivityCheckJSONMockImage   = "quay.io/cilium/json-mock:v1.3.0@sha256:2729064827fa9dbfface8d3df424feb6c792a0ba07117b844349635c93c06d2b"
 
 	ConfigMapName = "cilium-config"
-	Version       = "v1.10.2"
+	Version       = "v1.10.3"
```
nebril, line 79: nitpick - this change should probably be in separate commit in case we need to revert it?

nebril (approved):

defaults/defaults.go
joestringer: I think this has to go together, otherwise you may end up with clusters with taints that will never be removed.

twpayne (approved):

defaults/defaults.go
aanm: Correct

.github/workflows/aks.yaml
aanm: System node pool can not use spot VM

tklauser (approved):

PR #410, opened 2021-07-07
Revert "k8s: Use TTY to cancel execution"

### LINKED ISSUE #408: sysdump command is broken
after commit https://github.com/cilium/cilium-cli/commit/8d0139bb8af6b885da71c4c1d0e14af41f901274, output from cilium-bugtool command is going to stdout instead of stderr for some reason: 

cilium sysdump fails with this error:
```
⚠️ The sysdump may be incomplete — 1 error occurred:
        * failed to collect 'cilium-bugtool' output for "cilium-tdvgp" in namespace "kube-system": output doesn't contain archive name
```

bmcustodio:
This reverts commit `8d0139bb8af6b885da71c4c1d0e14af41f901274`. The current implementation breaks 'cilium sysdump', so I think we'll need to rethink that one through.

Fixes #408


tklauser (approved):
Thanks

PR #413, opened 2021-07-08
k8s: Check observedGeneration in status

gandro:
When a DaemonSet or Deployment is updated, it might take a while for the
status field to reflect the status of the updated version. If we check
the status too soon, there is a high chance that we are checking an
outdated status.

Therefore, check the observedGeneration field first, before reading out
the status flag. This ensures that the reported status actually refers
to the most recent desired state.

Recommended reading: https://alenkacz.medium.com/-250728868792

FWIW: I'm not sure we actually have hit this problem in practice. I just noticed it while debugging something else.

### DIFF
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -208,6 +208,11 @@ func (c *Client) CheckDeploymentStatus(ctx context.Context, namespace, deploymen
 		return fmt.Errorf("deployment is not available")
 	}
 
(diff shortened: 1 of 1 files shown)


tklauser (approved):
Nice catch and thanks for the link providing additional context.

nathanjsweet (approved):
Very cool.

nathanjsweet:
test-me-please

PR #416, opened 2021-07-12
CI workflows: Build cilium binary

michi-covalent:
We need the binary for the sysdump command in case of failure.

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -24,6 +24,19 @@ jobs:
       - name: Checkout
         uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
 
+      - name: Set up Go
+        uses: actions/setup-go@37335c7bb261b353407cff977110895fa0b4f7d8
+        with:
+          go-version: 1.16.5
+
+      - name: Set up Go for root
+        run: |
+          sudo ln -sf `which go` `sudo which go` || true
+          sudo go version
+
+      - name: Build and install cilium CLI binary
+        run: sudo make install
+
       - name: Install eksctl CLI
         run: |
           curl -LO "https://github.com/weaveworks/eksctl/releases/download/0.54.0/eksctl_$(uname -s)_amd64.tar.gz"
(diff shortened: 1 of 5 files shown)


tklauser (approved):

tklauser:
* External workloads failed with:

```
 ERROR: (gcloud.compute.instances.delete) Could not fetch resource:
 - The resource 'projects/***/zones/us-west2-a/instances/cilium-cilium-cli-1023755226-vm' was not found
```

https://github.com/cilium/cilium-cli/actions/runs/1023755226

* AKS failed due to missing privileges, likely because AKS secrets were changed at around the same time? /cc @nbusseneau 

https://github.com/cilium/cilium-cli/actions/runs/1023755252

Merging this PR despite the above failures as we currently don't get sysdumps for some failed tests without building the `cilium` binary.

nbusseneau (changes requested):
Sorry but I have to veto that PR :/

The whole point of having the PR-built `cilium` run in-cluster is to forbid attackers to edit `cilium-cli` code and then submit PR and execute `cilium` in the context of the workflow, which would allow them to leak our secrets due to the `pull_request_target` triggers running in a privileged context. This is why these were removed in #212.

We cannot have the workflows themselves use code coming from the PR.

nbusseneau:
> 
> 
>     * External workloads failed with:
> 
> 
> ```
>  ERROR: (gcloud.compute.instances.delete) Could not fetch resource:
>  - The resource 'projects/***/zones/us-west2-a/instances/cilium-cilium-cli-1023755226-vm' was not found
> ```
> 
> https://github.com/cilium/cilium-cli/actions/runs/1023755226
> 
>     * AKS failed due to missing privileges, likely because AKS secrets were changed at around the same time? /cc @nbusseneau
> 
> 
> https://github.com/cilium/cilium-cli/actions/runs/1023755252
> 
> Merging this PR despite the above failures as we currently don't get sysdumps for some failed tests without building the `cilium` binary.

Unrelated to my comment above: in this instance I would advocate for re-running the workflows:
- The GCP VM boot issue is a known infrastructure issue, and is one of the cases where re-running is allowed (and should actually be automated, it's just not done yet, in case anyone's up for it :D).
- Likewise, I would also consider the AKS failure due to missing privileges to be infrastructure issue, hence re-running.

tklauser:
Superseded by #423

PR #431, opened 2021-07-14
skip Succeeded pods

### LINKED ISSUE #430: cilium should not delete succeeded pods
Pod with succeeded phase means all containers exit with code of 0, and the system will not restart any of these containers.  Delete these pods will also delete all the logs of the pods. So it makes sense not to delete the succeeded pods.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2543-L2545

xyz-li:
Fixes #430


ldelossa (approved):
LGTM

tklauser (approved):
Thanks

PR #439, opened 2021-07-16
.github/workflows: wait for install job before checking aws-node DaemonSet

tklauser:
Wait until the cilium-cli-install job completed and Cilium is installed
before checking the status of the aws-node DaemonSet. Otherwise, we
might fail the check if the aws-node DaemonSet is still terminating.

### DIFF
--- a/.github/in-cluster-test-scripts/eks-install.sh
+++ b/.github/in-cluster-test-scripts/eks-install.sh
@@ -6,5 +6,4 @@ set -e
 # Install Cilium
 cilium install \
   --cluster-name "${CLUSTER_NAME}" \
-  --wait=false \
   --config monitor-aggregation=none
--- a/.github/in-cluster-test-scripts/eks-tunnel-install.sh
+++ b/.github/in-cluster-test-scripts/eks-tunnel-install.sh
@@ -6,7 +6,6 @@ set -e
 # Install Cilium
 cilium install \
   --cluster-name "${CLUSTER_NAME}" \
-  --wait=false \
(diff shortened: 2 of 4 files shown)


tklauser:
Both EKS workflows passed:

* EKS (ENI) https://github.com/cilium/cilium-cli/runs/3101902415?check_suite_focus=true
* EKS (tunnel) https://github.com/cilium/cilium-cli/runs/3101902514?check_suite_focus=true

Removing test commit and marking as ready to review.

.github/in-cluster-test-scripts/eks-install.sh
```diff
@@ -6,5 +6,4 @@ set -e
 # Install Cilium
 cilium install \
   --cluster-name "${CLUSTER_NAME}" \
-  --wait=false \
```
nbusseneau, line 9: We already actively `wait` immediately after enabling Hubble in the next script. Initially, the whole point of splitting the EKS scripts into 2 separate `install`/`test` steps was to take advantage of the fact we install into clusters without any nodes, with nodes being added later.

Now that we don't have that and instead create clusters with nodes available right away, I think the split is not warranted anymore and we should switch back to a single script that works like the others.

nbusseneau (changes requested):

.github/in-cluster-test-scripts/eks-install.sh
nbusseneau: (proposal in #442)

tklauser:
Let's close this in favor of #442.

PR #440, opened 2021-07-19
adds custom mount cmd for k3d

vniche:


brb (changes requested):
Thanks for your contribution!

Could you please extend your commit message explaining why the change is needed.

vniche:
> Thanks for your contribution!
> 
> Could you please extend your commit message explaining why the change is needed.

Hello there @brb, my pleasure to contribute! Just updated the PR message and also signed-off the commit.

aditighag:
Hi @vniche 

>raw commands doesn't work, there for mount went /bin/busybox mount

We've removed the dependency on having `mount` installed on the underlying distro in Cilium OSS daemonset yaml [1]. I'll be making similar changes to the cilium-cli crafted yaml. 

>the second command mount --make-shared /sys/fs/bpf is the result of an error message that was stating that this mount point was not shared (got the command from here).

See this commit on cilium repo that uses bidirectional mount in lieu of the running the mount command to mount BPF fs - https://github.com/cilium/cilium/pull/16656. /cc @aanm 


[1] https://github.com/cilium/cilium/pull/16815/commits/c11ca032e1d60ac2f927462490acc49830d78bcb

brb:
Considering the @aditighag comment above, marking the PR as draft.

vniche:
I don't mind if this gets closed to help on cleaning of open PRs, no use but reference for a workaround until the actual is merged.

PR #450, opened 2021-07-21
workflows: add test exceptions to EKS tunnel workflow

nbusseneau:
Following merge of #251, testing on EKS in tunnel mode is now reliably broken due to an issue in upstream: cilium/cilium#16975

This type of failure was already happening before #251, but the PR made it very evident as before we were only running checks from a few random pods (which would sometimes work) and are now running checks from all pods, drastically increasing the chance of hitting at least one failure.

Since the test setup is not to blame and this is an actual issue with Cilium in tunnel mode, we disable the failing tests until upstream issue is fixed.

### DIFF
--- a/.github/in-cluster-test-scripts/eks-tunnel.sh
+++ b/.github/in-cluster-test-scripts/eks-tunnel.sh
@@ -26,4 +26,10 @@ sleep 10s
 [[ $(pgrep -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay" | wc -l) == 2 ]]
 
 # Run connectivity test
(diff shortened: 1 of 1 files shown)


nbusseneau:
Link to working test run of the changes in the EKS tunnel workflow: https://github.com/cilium/cilium-cli/actions/runs/1056560809

ti-mo (approved):
Thanks, should provide some breathing room in the mean time. :+1:

michi-covalent (approved):

christarazi (approved):

michi-covalent:
ok let's ship it

PR #451, opened 2021-07-21
workflows: fix in-cluster job kubectl wait

nbusseneau:
`kubectl wait --for=condition=complete --timeout=X` behaviour is a bit counterintuitive: it waits until either the job succeeds or timeout is hit. When the job fails, it does not stop waiting: it will continue waiting until timeout is hit.

For watching for failures, `--for=condition=failed` should be used. However, this will likewise wait until either the job fails or timeout is hit, and will not stop waiting if the job succeeds.

`kubectl wait` unfortunately does not allow waiting for multiple conditions. To work around this, we set up two concurrent background waits for both conditions, and actively wait for the first one to end.

This will ensure we do not wait for the whole allocated timeout everytime there is an error during the in-cluster script execution.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -160,8 +160,20 @@ jobs:
             --set azure.client_secret=${{ steps.az.outputs.client-secret }}
 
       - name: Wait for install job
+        env:
+          timeout: 5m
         run: |
-          kubectl -n kube-system wait job/cilium-cli-install --for=condition=complete --timeout=5m
+          # Background wait for job to complete or timeout
+          kubectl -n kube-system wait job/cilium-cli-install --for=condition=complete --timeout=${{ env.timeout }} &
+          complete_pid=$!
+
+          # Background wait for job to fail
+          (kubectl -n kube-system wait job/cilium-cli-install --for=condition=failed --timeout=${{ env.timeout }} && exit 1) &
+          failed_pid=$!
+
+          # Active wait for whichever background process ends first
+          wait -n $complete_pid $failed_pid
+          exit $?
 
       - name: Delete the first node pool
         run: |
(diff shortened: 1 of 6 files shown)


nbusseneau:
Links to test runs of workflow changes:
- EKS: https://github.com/cilium/cilium-cli/actions/runs/1053975796
- EKS (tunnel): https://github.com/cilium/cilium-cli/actions/runs/1053975803
- GKE: https://github.com/cilium/cilium-cli/actions/runs/1053975793
- Multicluster: https://github.com/cilium/cilium-cli/actions/runs/1053975794
- External workloads: https://github.com/cilium/cilium-cli/actions/runs/1053975802

michi-covalent (approved):

christarazi (approved):
Nice find

nbusseneau:
All test runs passed except EKS (tunnel), but that is because of the consistent failure discussed on Slack [here](https://cilium.slack.com/archives/C7PE7V806/p1626877817422600?thread_ts=1626877621.422500&cid=C7PE7V806).

Since:
- The changes in this PR have been validated against other workflows and are exactly the same in EKS (tunnel).
- The PR improves CI testing on `cilium-cli` (faster runs).
- The PR will help to debug the EKS (tunnel) issue itself, since the workflow is now more reliable and does not hit the "hard job timeout" (as seen in other runs of the workflow) which causes unreliable information gathering / cluster cleanup.
=> We are in one of the exception cases to the zero-flakes strategy and do not need to wait to rebase on top of the future (hopefully soon) fix for the EKS (tunnel) workflow.

This PR can be marked as `ready-to-merge` once reviews are in.

aanm (changes requested):
Needs similar changes in the *-v1.10 GH workflows

nbusseneau:
> Needs similar changes in the *-v1.10 GH workflows
Sir, this is `cilium-cli`. We don't do that here.

aanm (approved):

Issue #455, opened 2021-07-28
Run connectivity tests on services with a restrictive policy


pchaigno:
In https://github.com/cilium/cilium/pull/17000, we fixed a bug where policy enforcement was failing when using IPsec + endpoint routes and connecting to a ClusterIP service.

IPsec + endpoint routes is tested on CI 3.0 since endpoint routes are enabled by default on GKE, EKS, and AKS. `cilium connectivity test` also covers [service handling](https://github.com/cilium/cilium-cli/blob/ecaaeb8ddac02aaf744e0db79a9d0fee43752f5c/connectivity/tests/service.go), including with a policy enforced at the same time: https://github.com/cilium/cilium-cli/blob/ecaaeb8ddac02aaf744e0db79a9d0fee43752f5c/connectivity/suite.go#L71-L82

The allow-all policy is however not enough to detect issues in security identities. Since it allows all, it will also allow a connection from an incorrect `world` identity as in the above bug. So to detect that bug, we would need to replace the allow-all policy by a policy that covers the actual endpoints in use (maybe with `to/fromEndpoints`).

PR #462, opened 2021-08-04
connectivity: Replace allow-all test with allow-all-except-world

christarazi:
In the Cilium datapath, the identity "world" is a special case. If
traffic cannot be identified, then the datapath falls back to assigning
it as "world". Having only "allow-all" in the connectivity test will
mask failures in which we have datapath bugs that incorrectly assign
traffic as "world", but the traffic is still allowed. One such case is
https://github.com/cilium/cilium/pull/17000.

This commit replaces the "allow-all" test with "allow-all-except-world"
(and unmanaged), thereby covering the datapath special case. We don't
want to allow unmanaged traffic either because it could also lead mark
underlying datapath bugs.

Fixes: https://github.com/cilium/cilium-cli/issues/455


connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -2,12 +2,20 @@ apiVersion: cilium.io/v2
...
 spec:
   endpointSelector: {}
   egress:
   - toEntities:
-    - all
+    - host
+    - remote-node
+    - cluster
+    - init
+    - health
```
jrajahalme, line 14: This looks like a different test to me. The point of `allow-all` test was to test with a policy that does not drop anything. This makes it possible to rule out policy implementation as the cause for a consistent failure one might see in a deployment.

Also, I'm not sure these listed entities will cover any CIDRs, such as generated from toFQDN policies.

jrajahalme (changes requested):

connectivity/manifests/allow-all-except-world.yaml
pchaigno: > The point of allow-all test was to test with a policy that does not drop anything.

Doesn't having no policies loaded at all implement the same? We will also go through the policy engine and allow all.

> This makes it possible to rule out policy implementation as the cause for a consistent failure one might see in a deployment.

Why doesn't `no-policies` already allow us to rule out policy enforcement as the cause of failures?

aditighag: >Doesn't having no policies loaded at all implement the same? We will also go through the policy engine and allow all.

If an endpoint isn't selected by any policies, we won't enter the policy enforcement block at all, no?

aditighag: >If traffic cannot be identified, then the datapath falls back to assigning it as "world".

This behavior will try to mask failures where packets are not taking the expected path, no? Why don't we distinguish between "world" and "unknown"?

>Also, I'm not sure these listed entities will cover any CIDRs, such as generated from toFQDN policies.

+1

pchaigno: > If an endpoint isn't selected by any policies, we won't enter the policy enforcement block at all, no?

We will. That's why we load two rules in that case to allow everything on egress and ingress (+ allow all from hostns on ingress):
```
$ vagrant ssh k8s1 -- sudo cilium bpf policy get --all
/sys/fs/bpf/tc/globals/cilium_policy_00980:

POLICY   DIRECTION   LABELS (source:key[=value])   PORT/PROTO   PROXY PORT   BYTES   PACKETS   
Allow    Ingress     reserved:unknown              ANY          NONE         0       0         
Allow    Ingress     reserved:host                 ANY          NONE         0       0         
Allow    Egress      reserved:unknown              ANY          NONE         0       0
[...]
```

jrajahalme: > Why doesn't no-policies already allow us to rule out policy enforcement as the cause of failures?

It does, and the actual policy in the datapath should be the same :-) So yes it seems we do not need the `allow-all` test, so it can be morphed into a more restrictive test.

christarazi:
Kind test is blocked on: https://github.com/cilium/cilium/issues/17144?

christarazi:
@reviewers, please see the new commit. This should unblock this PR.

connectivity/suite.go
```diff
@@ -63,8 +63,6 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 			tests.PodToPod(""),
 			tests.ClientToClient(""),
 			tests.PodToService(""),
-			tests.PodToRemoteNodePort(""),
```
aditighag, line 66: I wonder if we should comment these out with the corresponding context so that we don't try to add these back in future until https://github.com/cilium/cilium/issues/17144 is resolved.

The commit message doesn't quite capture the context for why we are disabling the test. At the very least, mention about how nodeport is typically used for N-S traffic, and it's okay to skip the pod-nodeport case as a trade-off.

aditighag (changes requested):
Changes look fine to me.

connectivity/suite.go
christarazi: Good point, updated.

pchaigno (approved):

aditighag (approved):

jrajahalme (approved):

connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -0,0 +1,27 @@
...
+  egress:
+  - toEntities:
+    - host
+    - remote-node
+    - cluster
+    - init
+    - health
+    - kube-apiserver
+  - toEndpoints:
+    - {}
```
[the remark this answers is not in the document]
christarazi: Because it doesn't cover other clusters in the clustermesh. I know, I also had the same thought when I discovered the multicluster Ci was failing reliably. Took me some time to realize the mistake. The key was checking the `cilium policy selectors` output.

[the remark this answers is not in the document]
pchaigno: Oh, wow! I had never considered this. Does that mean we'd need a new entity for `clusters` (with a less ambiguous name)?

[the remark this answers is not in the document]
christarazi: Oh good point. I considered that yesterday, but more of should we expand the definition of `cluster`. I came to the conclusion that that would be confusing, but a new entity of `clusters` is actually an idea worth considering!

twpayne (approved):
LGTM, thanks!

connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -0,0 +1,27 @@
...
+spec:
+  endpointSelector: {}
+  egress:
+  - toEntities:
+    - host
+    - remote-node
+    - cluster
+    - init
+    - health
+    - kube-apiserver
```
jrajahalme, line 15: This line breaks the tests in Cilium releases < 1.11:
```
  🔥 Running test allow-all-except-world: setting up test: applying network policies: policy application failed: CiliumNetworkPolicy.cilium.io "allow-all-except-world" is invalid: [spec.egress.toEntities: Unsupported value: "kube-apiserver": supported values: "all", "world", "cluster", "host", "init", "unmanaged", "remote-node", "health", "none", spec.ingress.fromEntities: Unsupported value: "kube-apiserver": supported values: "all", "world", "cluster", "host", "init", "unmanaged", "remote-node", "health", "none"]
```

connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - toEndpoints:
+    - {}
+  ingress:
+  - fromEntities:
+    - host
+    - remote-node
+    - cluster
+    - init
+    - health
+    - kube-apiserver
```
jrajahalme, line 25: This too.

connectivity/manifests/allow-all-except-world.yaml
pchaigno: Might be best to open an issue for that.

tklauser: #719

christarazi: PR #720 is out to fix

PR #471, opened 2021-08-06
workflows: display logs in job execution step

nbusseneau:
Doing this has two advantages:

- Behavior is similar to what we currently have on `cilium/cilium` and what we had on `cilium/cilium-cli` before #212, with the logs to the various `cilium` commands displayed immediately on execution.
- When a job fails, the failing step is automatically expanded in GitHub Actions workflow run view. This will avoid the need to manually open the post-test gathering step to know why a job failed.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -173,7 +173,11 @@ jobs:
 
           # Active wait for whichever background process ends first
           wait -n $complete_pid $failed_pid
-          exit $?
+          EXIT_CODE=$?
+
+          # Retrieve job logs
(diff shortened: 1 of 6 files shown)


michi-covalent (approved):
🚢

nbusseneau:
Link to test run of workflow changes: https://github.com/cilium/cilium-cli/actions/runs/1105522366
Output is present in the `Wait for test job` step as expected.

Removing temporary commit.

christarazi (approved):

aanm (approved):

PR #473, opened 2021-08-09
external workload: Increase the VM creation retry count

michi-covalent:
Increase the retry count from 3 to 10. I have seen 3 consecutive
failures a couple of times 😐

Ref: https://github.com/cilium/cilium-cli/actions/runs/1113449756

### DIFF
--- a/.github/workflows/externalworkloads.yaml
+++ b/.github/workflows/externalworkloads.yaml
@@ -74,7 +74,7 @@ jobs:
         with:
           retry_on: error
           timeout_minutes: 1
-          max_attempts: 3
+          max_attempts: 10
           command: |
             gcloud compute instances create ${{ env.vmName }} \
               --labels "usage=${{ github.repository_owner }}-${{ github.event.repository.name }},owner=${{ steps.vars.outputs.owner }}" \


nbusseneau (approved):
:/

michi-covalent:
rebased to pick up https://github.com/cilium/cilium-cli/pull/474

.github/workflows/externalworkloads.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
         with:
           retry_on: error
           timeout_minutes: 1
-          max_attempts: 3
+          max_attempts: 10
```
pchaigno, line 77: If 3+ retries are needed from time to time only, 10 seems a bit excessive no? Also, won't we hit the timeout of 20 minutes anyway before reaching 10 retries?

michi-covalent: timeout is per attempt https://github.com/nick-invision/retry#timeout_minutes

not sure what the right value is tbh.

pchaigno: I was referring to the timeout of the job, on line 38 of the workflow.

michi-covalent: ah yes of course. personally i'd say we might as well retry until the 20 minute timeout until we come up with a better way to handle this.

pchaigno (approved):
LGTM once second commit is removed of course.

michi-covalent:
ignoring https://github.com/cilium/cilium-cli/pull/473/checks?check_run_id=3285027333, after this pr gets merged, the retry count will get bumped up.

PR #474, opened 2021-08-09
workflows: fix logs not being displayed on failure

nbusseneau:
In #471, we refactored in-cluster jobs log retrieval to be displayed immediately in the same step as the one waiting for job execution to end.

Unfortunately, this only works when the job succeeds, as if the job does not then a non-0 exit code is received and the step immediately stops due to fail-fast being enabled by default in workflows.

Since we are manually handling exit codes, we disable fail-fast behaviour.

Also, we fix exit codes not being properly handled for EKS uninstall steps.

Fixes: e20c76976fde976e52603608810c1c034bf2c016

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -178,6 +178,7 @@ jobs:
           # Retrieve job logs
           kubectl logs --timestamps -n kube-system job/cilium-cli-install
           exit ${EXIT_CODE}
+        shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
       - name: Delete the first node pool
         run: |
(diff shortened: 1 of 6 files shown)


michi-covalent (approved):

nbusseneau:
It worked -- uuuuh I mean, we got a failure, but the logs are displayed! https://github.com/cilium/cilium-cli/actions/runs/1113613560

Removing temporary commit.

michi-covalent:
ok i'm merging this. would be difficult to investigate test failures without cilium-cli logs.

pchaigno (commented):
LGTM!

PR #480, opened 2021-08-11
internal/cli/cmd, install: Add --dev flag

christarazi:
This flag allows developers to use the `install` command, passing
a `--version` that's not accepted by the version validation regex. This
can be a SHA or any other string.

The intended use for developers is to pass this flag to the `install`
command when specifying a non-typical version string:

```
cilium-cli install --dev --version $(git describe) ...
```

Signed-off-by: Chris Tarazi <chris@isovalent.com>


install/install.go
```diff
@@ -1172,7 +1173,10 @@ func (k *K8sInstaller) getCiliumVersion() semver.Version {
 		ersion = strings.TrimPrefix(k.params.Version, "v")
 	}
 	v, err := versioncheck.Version(ersion)
-	if err != nil {
+	switch {
+	case k.params.DevMode:
+		return v
+	case err != nil:
```
tklauser, line 1179: I think this part of the change will break installation (i.e. creation of the ConfigMap wrt. `enable-ipv4-masquerade` vs. `masquerade`) when running development versions build from Cilium `v1.9` branch or earlier, see here:

https://github.com/cilium/cilium-cli/blob/3b177845c18812045d92e6bc398b9d6be6f9d986/install/install.go#L1319-L1324

In this case, I think we should instead rely on the developer specifying the correct base version using the `--base-version` flag and if not specified assume >=1.10.0 even in `--dev` mode.

tklauser (changes requested):

tklauser:
FWIW, with #568 merged it's now possible to install a particular SHA using:

```
$ cilium install --version "-ci:$(git describe)"
```

Same for `:latest`:

```
# latest release or local build when testing e.g. in kind
$ cilium install --version :latest
# latest master build
$ cilium install --version "-ci:latest"
```

christarazi:
Sounds like we have no need for this PR then :rocket:

ldelossa:
@christarazi @tklauser the follow does not work when testing with cli code at head `bcc992b93cce627c4505de25cbb207164b90390c`

```
 ./cilium install \
--agent-image localhost:5000/cilium/cilium-dev --version ":podCidrs2" \
--config="bgp-announce-lb-ip=true" \
--config="bgp-announce-pod-cidr=true" \
--config="debug=true"  
```

The image deployed is missing the version tag:
```
Image
localhost:5000/cilium/cilium-dev
```

PR #482, opened 2021-08-12
Add support for displaying the running Cilium version

### LINKED ISSUE #131: Show version of cilium (as well as of cilium-cli)
`cilium version` reports the version of the CLI agent. It would be useful if it also showed the running version of cilium (there's already a [TODO comment](https://github.com/cilium/cilium-cli/blob/f05d883cedea527377fac6a23a6849ba39b2a8ea/internal/cli/cmd/version.go#L47) to this effect)

zhaojizhuang:
Show the running version of cilium

Fixes #131

### DIFF
--- a/internal/cli/cmd/version.go
+++ b/internal/cli/cmd/version.go
@@ -15,6 +15,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
@@ -48,16 +49,26 @@ func getLatestStableVersion() string {
 }
 
 func newCmdVersion() *cobra.Command {
-	return &cobra.Command{
+	var namespace string
+	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Display detailed version information",
 		Long:  `Displays information about the version of this software.`,
-		Run: func(cmd *cobra.Command, _ []string) {
-			// TODO: add support for reporting the Cilium version running in
-			// the cluster, if any. See https://github.com/cilium/cilium-cli/issues/131
+		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Printf("cilium-cli: %s compiled with %v on %v/%v\n", Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 			fmt.Printf("cilium image (default): %s\n", defaults.Version)
 			fmt.Printf("cilium image (stable): %s\n", getLatestStableVersion())
+			version, err := k8sClient.GetRunningCiliumVersion(context.Background(), namespace)
+			if version == "" || err != nil {
+				fmt.Printf("cilium image (running): unknown. Unable to obtain cilium version, no cilium pods found in namespace %q\n", namespace)
+			} else {
+				fmt.Printf("cilium image (running): %s\n", version)
+			}
+			return nil
 		},
 	}
+
+	cmd.Flags().StringVar(&contextName, "context", "", "Kubernetes configuration context")
+	cmd.Flags().StringVarP(&namespace, "namespace", "n", "kube-system", "Namespace Cilium is running in")
+	return cmd
 }
--- a/internal/cli/cmd/cmd.go
+++ b/internal/cli/cmd/cmd.go
@@ -36,7 +36,7 @@ func NewDefaultCiliumCommand() *cobra.Command {
 				return nil
 			}
 			switch cmd.Name() {
-			case "completion", "help", "version":
+			case "completion", "help":
 				return nil
 			}
 


internal/cli/cmd/version.go
```diff
@@ -32,6 +37,61 @@ var (
...
+			if err != nil {
+				return "", err
+			}
+		}
+	}
+
+	return "", errors.New("unable to obtain cilium version: no cilium pods found")
+}
+
+func (c *CiliumVersion) getCiliumVersionOfNamespace(ctx context.Context, namespace string) (string, error) {
```
gandro, line 75: Why are you duplicating
 https://github.com/cilium/cilium-cli/blob/6032e9e660af7da127163ad3e731e62ffd274120/internal/k8s/client.go#L676

here? Is there a reason we cannot use it?

zhaojizhuang: yeah, we can use it ,but have to encapsulate it ( for we don't know which namespace cilium-agent is running in)

internal/cli/cmd/version.go
```diff
@@ -32,6 +37,38 @@ var (
...
+			version, _ := c.client.GetRunningCiliumVersion(ctx, namespace.Name)
+			if version != "" {
+				return version, nil
+			}
+		}
+	}
+
+	return "", errors.New("unable to obtain cilium version: no cilium pods found")
+}
+
```
tklauser, line 71: Instead of first trying `kube-system` and potentially looping over multiple namespaces, how about just adding a `--namespace` flag to `cilium version` with a default set to `kube-system`, so it will work in the same way as other commands taking a namespace?

zhaojizhuang: Yeah,that sounds like a good suggestion

kaworu:
@zhaojizhuang thanks patch LGTM besides @tklauser comment about the cilium namespace. I think this PR closes #131 as well :tada:

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,45 @@
...
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
 	"runtime"
 	"strings"
 
-	"github.com/cilium/cilium-cli/defaults"
```
tklauser, line 24: Please keep the `github.com/cilium/cilium-cli` import in a separate import group before any other 3rd party imports, i.e.

```go
import (
	"context"
	"fmt"
	"io"
	"net/http"
	"runtime"
	"strings"
	
	"github.com/cilium/cilium-cli/internal/k8s"
	
	"github.com/spf13/cobra"
)
```

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,45 @@
...
+	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/internal/k8s"
 )
 
 // The following variables are set at compile time via LDFLAGS.
 var (
 	// Version is the software version.
 	Version string
+	// namespace is the ns where cilium  exists
+	namespace string
```
tklauser, line 36: The comment at the top of this `var` section does not apply to `namespace`. Please move this into a separate section.

Also, I think you can also drop the comment as it doesn't add any additional information. The purpose is obvious from the variable name.

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,45 @@
...
+type CiliumVersion struct {
+	client *k8s.Client
+}
+
+func NewCiliumVersion() *CiliumVersion {
+	return &CiliumVersion{client: k8sClient}
+}
+
+func (c *CiliumVersion) GetRunningCiliumVersionForCluster(ctx context.Context) (string, error) {
+
```
tklauser, line 48: Nit: delete this superfluous empty line

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,45 @@
...
+	version, _ := c.client.GetRunningCiliumVersion(ctx, namespace)
+	if version != "" {
+		return version, nil
+	}
+
+	return "", fmt.Errorf("unable to obtain cilium version: no cilium pods found in ns <%s>", namespace)
```
tklauser, line 54:
```suggestion
	return "", fmt.Errorf("unable to obtain cilium version: no cilium pods found in namespce %q", namespace)
```

internal/cli/cmd/version.go
```diff
@@ -48,16 +70,26 @@ func getLatestStableVersion() string {
...
-			// TODO: add support for reporting the Cilium version running in
-			// the cluster, if any. See https://github.com/cilium/cilium-cli/issues/131
+		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Printf("cilium-cli: %s compiled with %v on %v/%v\n", Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 			fmt.Printf("cilium image (default): %s\n", defaults.Version)
 			fmt.Printf("cilium image (stable): %s\n", getLatestStableVersion())
+			ciliumVersion := NewCiliumVersion()
+			version, err := ciliumVersion.GetRunningCiliumVersionForCluster(context.Background())
+			if err != nil {
+				fmt.Printf("cilium image (running): unknown. Error: %s \n", err.Error())
```
tklauser, line 84:
```suggestion
				fmt.Printf("cilium image (running): unknown. Error: %v\n", err)
```

internal/cli/cmd/version.go
```diff
@@ -48,16 +68,26 @@ func getLatestStableVersion() string {
...
-			// TODO: add support for reporting the Cilium version running in
-			// the cluster, if any. See https://github.com/cilium/cilium-cli/issues/131
+		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Printf("cilium-cli: %s compiled with %v on %v/%v\n", Version, runtime.Version(), runtime.GOOS, runtime.GOARCH)
 			fmt.Printf("cilium image (default): %s\n", defaults.Version)
 			fmt.Printf("cilium image (stable): %s\n", getLatestStableVersion())
+			ciliumVersion := NewCiliumVersion()
+			version, err := ciliumVersion.GetRunningCiliumVersionForCluster(context.Background())
+			if err != nil {
+				fmt.Printf("cilium image (running): unknown. Error: %v \n", err)
```
tklauser, line 82: Nit: drop the trailing space after the error message

```suggestion
				fmt.Printf("cilium image (running): unknown. Error: %v\n", err)
```

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,43 @@
...
+	return &CiliumVersion{client: k8sClient}
+}
+
+func (c *CiliumVersion) GetRunningCiliumVersionForCluster(ctx context.Context) (string, error) {
+	version, _ := c.client.GetRunningCiliumVersion(ctx, namespace)
+	if version != "" {
+		return version, nil
+	}
+
+	return "", fmt.Errorf("unable to obtain cilium version: no cilium pods found in ns %q", namespace)
```
tklauser, line 52:
```suggestion
	return "", fmt.Errorf("unable to obtain cilium version: no cilium pods found in namespace %q", namespace)
```

internal/cli/cmd/version.go
```diff
@@ -15,23 +15,43 @@
...
 
+type CiliumVersion struct {
+	client *k8s.Client
+}
+
+func NewCiliumVersion() *CiliumVersion {
+	return &CiliumVersion{client: k8sClient}
+}
+
+func (c *CiliumVersion) GetRunningCiliumVersionForCluster(ctx context.Context) (string, error) {
```
tklauser, line 46: Now that the implementation here merely consits of calling `c.client.GetRunningCiliumVersion`, let's just drop the `CiliumVersion` type and this method and inline this function into `newCmdVersion`.

internal/cli/cmd/version.go
```diff
@@ -15,21 +15,22 @@
...
-
 	"github.com/spf13/cobra"
 )
 
-// The following variables are set at compile time via LDFLAGS.
 var (
 	// Version is the software version.
 	Version string
+	// namespace is the ns where cilium  exists
+	namespace string
```
tklauser, line 33: Move this declaration into `newCmdVersion` and drop the comment.

tklauser (changes requested):

tklauser (approved):

gandro (approved):
Thanks!

PR #481, opened 2021-08-12
sysdump: Remove the Hidden flag

michi-covalent:
We've been using the sysdump command for a while in both cilium and
cilium-cli CI workflows and it's been stable.

Signed-off-by: Michi Mutsuzaki <michi@isovalent.com>

PR #483, opened 2021-08-12
internal/cli/cmd: fix formatting linter issue

tklauser:
The change in PR #481 wasn't properly formatted, leading to a
golangci-lint failure on master:
https://github.com/cilium/cilium-cli/runs/3310670546

Fix it by running `goimports -w internal/cli/cmd/sysdump.go`.

Also fix the `golangci-lint` GH action configuration to disable the `only-new-issues` which seems to lead to static checker failures correctly failing tests on PRs. See #484 for an example run.

### DIFF
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -34,9 +34,9 @@ var (
 
 func newCmdSysdump() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "sysdump",
-		Short:  "Collects information required to troubleshoot issues with Cilium and Hubble",
-		Long:   ``,
+		Use:   "sysdump",
+		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
+		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Silence klog to avoid displaying "throttling" messages - those are expected.
 			klog.SetOutput(io.Discard)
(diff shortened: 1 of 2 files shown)


michi-covalent:
not sure why the workflow is not starting. will re-push

internal/cli/cmd/sysdump.go
```diff
@@ -34,9 +34,9 @@ var (
 
 func newCmdSysdump() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "sysdump",
-		Short:  "Collects information required to troubleshoot issues with Cilium and Hubble",
-		Long:   ``,
+		Use:   "sysdump",
+		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
+		Long:  ``,
```
pchaigno, line 39: It's really weird that the test passed in the original PR. Could it be due to a bug in golangci-lint when setting this to true? https://github.com/cilium/cilium-cli/blob/6a1699d6f34f4b65d333b41b01999ee70db35f47/.github/workflows/go.yml#L35-L36

pchaigno: Ah, just saw the second commit and seems you had the same though as I did :smiley:

pchaigno (approved):

nebril (approved):

tklauser:
AKS job failed on #367: https://github.com/cilium/cilium-cli/pull/483/checks?check_run_id=3314687814
Multicluster job failed on #361: https://github.com/cilium/cilium-cli/pull/483/checks?check_run_id=3314687986

Both are known flakes and this PR only affects code formatting and CI. And it fixes failing static checks on current `master`, thus merging.

PR #499, opened 2021-08-26
Add managed endpoint count to status command

### LINKED ISSUE #495: Add managed endpoint count to status command
Occasionally there can be problems come up in clusters where the initial Cilium installation was unsuccessful, so when users create pods, Cilium does not actually manage the endpoints but they are managed by some other plugin (kubenet or another CNI).

The `cilium status` command currently prints a bunch of useful statistics around Cilium pods and whether they are ready or not, maybe we could add another line something like:

```
Cluster Pods:  10/10 managed by Cilium
```

The first number can be derived from `$ kubectl get ciliumendpoints --all-namespaces -o json | jq '.items | length'`.
The second number can be derived from `$ kubectl get pods --all-namespaces -o json | jq '[.items[] | select(.spec.hostNetwork |.==null)] | length'`.

If the counts are not equal, then we should consider that to be a warning sign.

zhaojizhuang:
Fix #495 

Add managed endpoint count to status command
Signed-off-by: zhaojizhuang <571130360@qq.com>

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -416,6 +416,7 @@ type k8sClusterMeshImplementation interface {
 	CreateCiliumExternalWorkload(ctx context.Context, cew *ciliumv2.CiliumExternalWorkload, opts metav1.CreateOptions) (*ciliumv2.CiliumExternalWorkload, error)
 	DeleteCiliumExternalWorkload(ctx context.Context, name string, opts metav1.DeleteOptions) error
 	GetRunningCiliumVersion(ctx context.Context, namespace string) (string, error)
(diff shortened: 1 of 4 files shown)


joestringer (approved):
:tada: LGTM, thanks for the submission!

tklauser (approved):

tklauser:
Thanks @zhaojizhuang 

PR approved by all requested teams. Azure tests (https://github.com/cilium/cilium-cli/runs/3476088449?check_suite_focus=true) and kind tests (https://github.com/cilium/cilium-cli/runs/3461510199?check_suite_focus=true) failed because of #367. This is an external contributor's PR, so the zero flake policy doesn't apply. Merging.

PR #501, opened 2021-08-30
go.mod, vendor: bump github.com/go-openapi/strfmt to v0.20.2

tklauser:
This will also bump the transitive dependency
go.mongodb.org/mongo-driver to v1.5.1 in order to fix CVE-2021-20329

Ref. https://github.com/advisories/GHSA-f6mq-5m25-4r72


rolinh (approved):

tklauser:
AKS is the only failing test and it's failing presumably because of an AKS infrastructure issue. Given this is a security fix, I'll thus merge this PR.

PR #503, opened 2021-08-30
Add a flag to disable the ipcache check

michi-covalent:
Add `--skip-ip-cache-check` flag with the default set to true. This is
meant to be a temporary flag while we investigate what's causing the flake.

Ref: #361

### DIFF
--- a/internal/cli/cmd/connectivity.go
+++ b/internal/cli/cmd/connectivity.go
@@ -119,6 +119,7 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.Verbose, "verbose", "v", false, "Show informational messages and don't buffer any lines")
 	cmd.Flags().BoolVarP(&params.Debug, "debug", "d", false, "Show debug messages")
 	cmd.Flags().BoolVarP(&params.PauseOnFail, "pause-on-fail", "p", false, "Pause execution on test failure")
+	cmd.Flags().BoolVar(&params.SkipIPCacheCheck, "skip-ip-cache-check", true, "Skip IPCache check")
 
 	return cmd
 }
--- a/connectivity/check/check.go
+++ b/connectivity/check/check.go
@@ -34,6 +34,7 @@ type Parameters struct {
 	Verbose               bool
 	Debug                 bool
 	PauseOnFail           bool
+	SkipIPCacheCheck      bool
 }
 
 func (p Parameters) ciliumEndpointTimeout() time.Duration {
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -477,13 +477,17 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
-	// Set the timeout for all IP cache lookup retries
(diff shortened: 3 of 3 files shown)


michi-covalent:
aks cluster creation failed with this error. re-triggering.

https://github.com/cilium/cilium-cli/actions/runs/1184098825

> ERROR: (ReconcileStandardLoadBalancerError) AKS encountered an internal error while attempting the requested Creating operation. AKS will continuously retry the requested operation until successful or a retry timeout is hit. Check back to see if the operation requires resubmission. Correlation ID: cc4b16c3-e207-4ec0-935b-172e2abdbb8b, Operation ID: 597493aa-980d-41f3-8c4c-5a418e13f412, Timestamp: 2021-08-30T23:27:08Z.

michi-covalent:
hitting the same issue again. not sure how to proceed 😞

internal/cli/cmd/connectivity.go
```diff
@@ -119,6 +119,7 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.Verbose, "verbose", "v", false, "Show informational messages and don't buffer any lines")
 	cmd.Flags().BoolVarP(&params.Debug, "debug", "d", false, "Show debug messages")
 	cmd.Flags().BoolVarP(&params.PauseOnFail, "pause-on-fail", "p", false, "Pause execution on test failure")
+	cmd.Flags().BoolVar(&params.SkipIPCacheCheck, "skip-ip-cache-check", true, "Skip IPCache check")
```
tklauser, line 122: Should we mark this flag as hidden so we can easily remove it again? Given that we intend to fix the ipcache flake #361 eventually,

tklauser (approved):

bmcustodio (approved):

internal/cli/cmd/connectivity.go
bmcustodio: Hmmm this makes sense, but at the same time I think that if users face the issue and the flag is hidden they won't know they can use it to proceed.

tklauser: Given that we default the flag to `true` (i.e. skip the ipcache check by default), I think users should not face the ipcache issue anymore. They would need to figure out to run with `--skip-ip-cache-check=false` at which point they already figured out the flag :wink:

bmcustodio: Oh right, makes total sense 💯 — I hadn't looked at the default.

twpayne (approved):

nbusseneau:
> hitting the same issue again. not sure how to proceed disappointed

This is an infrastructure issue, waiting for AKS support atm: https://github.com/cilium/cilium/issues/17260

michi-covalent:
re-triggered to see if aks happy now 👀

internal/cli/cmd/connectivity.go
tklauser: Looks like this hasn't been addressed before merging. I've opened another PR: #508

michi-covalent: thanks @tklauser , i just saw the green merge button and pushed it 😅

PR #504, opened 2021-08-31
install: set base version to current default version

### LINKED ISSUE #491: Can cilium-cli support the install of the image tag of "latest"?
Now I am doing some modification to the Cilium code and verification testing then, but I don't find the support for the built images with the tag "latest", so I would like ask if the cilium-cli can support that tag in a certain way.

I did some change to the current cilium-cli code to support this tag in my local environment, if it's useful I would like to check it.

tklauser:
Always set the base version to the current default version and rework
the parsing in `(*K8sInstaller).getCiliumVersion` to fall back to the base
version in case of a non-parseable version. This avoids the confusing log
message, e.g. when using `cilium install --version latest`:

    Unable to parse the provided version latest, assuming it's =1.10.0

Also reword the log message to hopefully be a bit less confusing.

Fixes #491

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -1154,17 +1154,15 @@ func (k *K8sInstaller) Exec(command string, args ...string) ([]byte, error) {
 }
 
 func (k *K8sInstaller) getCiliumVersion() semver.Version {
-	var ersion string
-	if k.params.BaseVersion != "" {
-		ersion = strings.TrimPrefix(k.params.BaseVersion, "v")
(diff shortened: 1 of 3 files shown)


michi-covalent (approved):
let's do it

tklauser:
Re-triggered AKS tests as it looks like they are working again: https://github.com/cilium/cilium/issues/17260#issuecomment-909356649. All green now, merging.

PR #507, opened 2021-08-31
ci: increase hubble events buffer

ldelossa:
currently the hubble event buffer is too small and events from the
begining of the test are overwritten before a read of the buffer takes
place.

this patch increases the hubble event buffer to its largest possible
value to ensure all events are captured.

Signed-off-by: Louis DeLosSantos <louis.delos@isovalent.com>

### DIFF
--- a/.github/in-cluster-test-scripts/aks-install.sh
+++ b/.github/in-cluster-test-scripts/aks-install.sh
@@ -12,4 +12,5 @@ cilium install \
   --azure-client-id "${AZURE_CLIENT_ID}" \
   --azure-client-secret "${AZURE_CLIENT_SECRET}" \
   --wait=false \
+  --config hubble-event-buffer-capacity=65535 \
   --config monitor-aggregation=none
--- a/.github/in-cluster-test-scripts/eks-tunnel.sh
+++ b/.github/in-cluster-test-scripts/eks-tunnel.sh
@@ -8,6 +8,7 @@ cilium install \
   --cluster-name "${CLUSTER_NAME}" \
   --wait=false \
   --config monitor-aggregation=none \
+  --config hubble-event-buffer-capacity=65535 \
   --datapath-mode=tunnel \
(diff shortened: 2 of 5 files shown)


nbusseneau (commented):
Oh wow that's a bummer. Before approving I'm wondering: do you think it would make sense to increase the default value for `hubble-event-buffer-capacity` rather than specify it manually? 

My goal would be first to prevent users hitting this in the wild, and second also avoid the need to specify it manually in every CI test file (here and over at `cilium/cilium`).

kaworu:
deferring to @twpayne here as he's working on a ring buffer refactoring and might have more pointers.

ldelossa:
@twpayne thanks. I'm not really sure if we want to bump this up as a "default". My assumption is the "default" case, a small hubble buffer is fine, as things will be reading off it real time and dumping to a file. 

We can more about expanding this buffer in sysdump situations where a point-in-time process extracts what ever data is available in the buffer.

michi-covalent:
i'm not against the change, but even 65k flows is pretty small since we run CI jobs with monitor aggregation disabled.

we already collect flows as a part of connectivity check, maybe it makes sense to add a flag to dump these flows in JSON format to a file instead?

ldelossa:
@michi-covalent absolutely. 

This pr was a "hold over" until I imagined a way for there to be some background reading and dumping of the buffer during syslog. 

I was not aware that we do currently capture the flows *during* the test. It would be better by all accounts to just tee logs to a file and copy those logs into the sysdump directory instead of doing the "point-in-time" read of the buffer at syscall creation time.

twpayne (approved):
I'm not against merging this, but would prefer the more robust solution proposed.

ldelossa:
@twpayne I agree.

ldelossa:
Closing to explore dumping real time flows to files and copying them into sysdump.

PR #511, opened 2021-09-01
Switch pod to world testing from `cilium.io` to `one.one.one.one`

nbusseneau:
Tentative fix for #367.

We have two main hypothesis around why connections to external domains are flaky:

- The external domain itself is unreliable.
- The external domain works fine, but CoreDNS is unreliable.

`cilium.io` is hosted on single non-HA EC2 instance, which is definitely the most robus thing out there. We propose switching to
`one.one.one.one` as a quick fix to check if that helps with the reliability.

If yes, we will evaluate moving to a HA system that we control ourselves (e.g. a DNS zone hosted at a major provider).

If not, we will investigate the CoreDNS hypothesis.

### DIFF
--- a/connectivity/tests/world.go
+++ b/connectivity/tests/world.go
@@ -10,7 +10,7 @@ import (
 	"github.com/cilium/cilium-cli/connectivity/check"
 )
 
-// PodToWorld sends multiple HTTP(S) requests to cilium.io
+// PodToWorld sends multiple HTTP(S) requests to one.one.one.one
 // from random client Pods.
 func PodToWorld(name string) check.Scenario {
 	return &podToWorld{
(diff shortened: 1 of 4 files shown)


connectivity/tests/world.go
```diff
@@ -32,10 +32,9 @@ func (s *podToWorld) Name() string {
 }
 
 func (s *podToWorld) Run(ctx context.Context, t *check.Test) {
-	chttp := check.HTTPEndpoint("cilium-io-http", "http://cilium.io")
-	chttps := check.HTTPEndpoint("cilium-io-https", "https://cilium.io")
-	chttpindex := check.HTTPEndpoint("cilium-io-http-index", "http://cilium.io/index.html")
-	jhttp := check.HTTPEndpoint("jenkins-cilium-io-http", "http://jenkins.cilium.io")
+	http := check.HTTPEndpoint("one-one-one-one-http", "http://one.one.one.one")
+	https := check.HTTPEndpoint("one-one-one-one-https", "https://one.one.one.one")
+	httpsindex := check.HTTPEndpoint("one-one-one-one-https-index", "https://one.one.one.one/index.html")
```
nbusseneau, line 37: I am probably missing context but is that check really needed?

michi-covalent (approved):

ldelossa:
Will "client-egress-to-fqdns-cilium-io.yaml" need to be updated?

```
client-egress-to-fqdns-cilium-io.yaml
1:apiVersion: cilium.io/v2
5:  name: client-egress-to-fqdns-cilium-io
20:    - matchName: "cilium.io"
```

tklauser (approved):

connectivity/tests/world.go
tklauser: Do you mean the check over HTTPS specifically or the check for `/index.html`?

If it's the latter, I think this is needed for L7 policy testing, see commit e54d39bc43e6884dd66fd2ae8ea7c674da10fedc /cc @jrajahalme

ldelossa (approved):
I do not believe this will fix our flakes as you're aware from my recent update to the flake issue proper, but I do think this is a good idea none-the-less.

connectivity/tests/world.go
nbusseneau: The check for `index.html`, yes. OK!

PR #533, opened 2021-09-13
Remove stable.txt and update release docs

tklauser:
First commit removes `stable.txt`, see commit message:

    Remove stable.txt
    
    This was added to allow CI in the Cilium repo to always get the latest
    stable release version. However, this lead to problems on certain
    updates, so that Cilium CI was switched back to pin the CLI version
    explicitly.
    
    Also, we forgot to bump stable.txt for the latest release v0.9.0.
    
    As we currently don't use release branches, we can also always refer to
    the latest release using the GitHub URL
    https://github.com/cilium/cilium-cli/releases/latest, e.g. in the Cilium
    docs.

Second commit updates `RELEASE.md` to include an optional step to update the `cilium-cli` Homebrew formula.


michi-covalent (approved):
sounds reasonable

PR #539, opened 2021-09-14
Ensure defaultMode=0400 for projected volumes containing secrets

tklauser:
Follow https://github.com/cilium/cilium/pull/17367:

It seems that these volumes were wrongly given permission 420 instead of
0400, giving write permission to the group. In all instances, the actual
volumeMounts were set to readonly so this change should have no effect
but better be more cautious about permissions for secrets.

Reported-by: Robin Hahling <robin.hahling@gw-computing.net>
Signed-off-by: Tobias Klauser <tobias@cilium.io>


rolinh (approved):
:rocket:

kaworu (approved):

PR #547, opened 2021-09-15
cmd: Add --extra-label-selectors log collector

gandro:
This adds an optional flag which allows users to specify additional pod label selectors, which will be used to collect collect logs of additional pods matching that selector in the Cilium namespace.

The flag allows for multiple selectors, e.g. `--extra-label-selectors key=value --extra-label-selectors foo=bar,baz=bat` will select any pods with the `key=value` label OR any pods with both the `foo=bar` AND `baz=bat` labels.

Example usage:
```console
cilium sysdump --extra-label-selectors k8s-app=kube-dns
```

### DIFF
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -55,6 +55,9 @@ func newCmdSysdump() *cobra.Command {
 	cmd.Flags().BoolVar(&sysdumpOptions.Debug,
 		"debug", sysdump.DefaultDebug,
 		"Whether to enable debug logging")
+	cmd.Flags().StringArrayVar(&sysdumpOptions.ExtraLabelSelectors,
+		"extra-label-selectors", nil,
+		"Optional set of labels selectors used to target additional pods for log collection.")
 	cmd.Flags().StringVar(&sysdumpOptions.HubbleLabelSelector,
 		"hubble-label-selector", sysdump.DefaultHubbleLabelSelector,
 		"The labels used to target Hubble pods")
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -36,6 +36,8 @@ type Options struct {
 	ClustermeshApiserverLabelSelector string
 	// Whether to enable debug logging.
 	Debug bool
+	// The labels used to target additional pods
+	ExtraLabelSelectors []string
 	// The labels used to target Hubble pods.
 	HubbleLabelSelector string
 	// Number of Hubble flows to collect.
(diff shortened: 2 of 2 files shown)


gandro:
This PR originally used the newly added flag to collect kube-dns logs to troubleshoot #342 - but I reverted the collection of kube-dns logsfor two reasons:

1. The kube-dns logs collected so far have not been very useful, as they don't contain debug logs
2. We are using the stable cilium-cli to collect the sysdumps on many workflows, so we can't use flags that haven't been released yet.

In any case, I believe the flag itself is still useful, as it allows the inclusion of additional logs without having to change the code. Thus I'm proposing the flag for inclusion, even if I don't have an urgent use-case for it right now.

internal/cli/cmd/sysdump.go
```diff
@@ -55,6 +55,9 @@ func newCmdSysdump() *cobra.Command {
 	cmd.Flags().BoolVar(&sysdumpOptions.Debug,
 		"debug", sysdump.DefaultDebug,
 		"Whether to enable debug logging")
+	cmd.Flags().StringArrayVar(&sysdumpOptions.ExtraLabelSelectors,
+		"extra-label-selector", nil,
```
tklauser, line 59: Given that this flag can potentially take multiple selectors, should it be called `extra-label-selectors`?

```suggestion
		"extra-label-selectors", nil,
```

tklauser: Looking a bit closer, it seems we're rather inconsistent with using singular `label` and `labels` in existing option names already.

gandro: Good idea! Addressed!

gandro: > Looking a bit closer, it seems we're rather inconsistent with using singular `label` and `labels` in existing option names already.

Only saw this after already pushing the change. 

As you pointed out, `extra-label-selectors` is slightly different than the others, in that it allows multiple selectors, e.g. `--extra-label-selectors k8s-app=kube-dns --extra-label-selectors foo=bar,baz=bat` will select any pods with the `k8s-app=kube-dns` pods OR any pods with the `foo=bar` AND `baz=bat` labels.

We can also consider renaming `extra-label-selectors` to something a bit different to make that more explicit. I don't have a strong opinion here.

tklauser: Sorry, I've edited my additional comment again while you replied. The existing option names seem to stem from `cilium-sysdump`, so I guess we want to keep them as is for backwards compatibility.

As for this PR:

> As you pointed out, extra-label-selectors is slightly different than the others, in that it allows multiple selectors, e.g. --extra-label-selectors k8s-app=kube-dns --extra-label-selectors foo=bar,baz=bat will select any pods with the k8s-app=kube-dns pods OR any pods with the foo=bar AND baz=bat labels.

Thanks for providing that example. Maybe we could extend the PR description and/or commit message to include this?

> We can also consider renaming extra-label-selectors to something a bit different to make that more explicit. I don't have a strong opinion here.

I think `extra-label-selectors` is fine. The name together with the usage message makes the intended purpose pretty clear IMO.

tklauser (approved):

internal/cli/cmd/sysdump.go
gandro: Kept the name, but updated the usage message and commit description to be a bit more helpful (hopefully).

gandro:
- Kind (https://github.com/cilium/cilium-cli/runs/3620500041?check_suite_focus=true) hit #367 
- External Workloads (https://github.com/cilium/cilium-cli/runs/3620541145?check_suite_focus=true) hit an infrastructure problem `ERROR: gcloud crashed (TransportError): HTTPSConnectionPool(host='oauth2.googleapis.com', port=443): Max retries exceeded with url: /token (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:1131)')))`
- GKE (https://github.com/cilium/cilium-cli/runs/3620499730?check_suite_focus=true) also hit an infrastructure problem (`net/http: TLS handshake timeout` while accessing the GKE K8s API)

Restarting the latter two.

michi-covalent (approved):

ldelossa (approved):

PR #555, opened 2021-09-20
.github/workflows: install Go for `go run` in multicluster job

### LINKED ISSUE #554: CI: multicluster fails when creating gcloud-free kubeconfig for cluster 1
Observed on https://github.com/cilium/cilium-cli/pull/553, see https://github.com/cilium/cilium-cli/runs/3650339141?check_suite_focus=true

```
 Creating the Kubernetes Service Account with minimal RBAC permissions.
Warning: resource namespaces/kube-system is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
namespace/kube-system configured
serviceaccount/kubectl created
clusterrolebinding.rbac.authorization.k8s.io/kubectl-test-crb created
Writing kubeconfig.
---
Done!

Copy the generated kubeconfig file to your Teleport Proxy server, and set the
kubeconfig_file parameter in your teleport.yaml config file to point to this
kubeconfig file.

Note: Kubernetes RBAC rules for Teleport were created, you won't need to create them manually.
vendor/k8s.io/client-go/plugin/pkg/client/auth/exec/metrics.go:21:2: cannot find package "." in:
	/home/runner/work/cilium-cli/cilium-cli/vendor/io/fs
Error: Process completed with exit code 1.
```

`io/fs` was added in Go 1.17, so it looks like that job is using somehow an older Go version (but not consistently, it seems).

It looks like that step is using `go run` but doesn't explicitly install it in a previous step like other jobs using `go`:

https://github.com/cilium/cilium-cli/blob/8e3f099579eeea351cfab514a58c7260ad6a223d/.github/workflows/multicluster.yaml#L126

tklauser:
The multicluster workflow uses `go run` but doesn't explicitly install a
specific Go version. Thus, in some cases an older Go version might be
used which leads to failures such as:
    
```
vendor/k8s.io/client-go/plugin/pkg/client/auth/exec/metrics.go:21:2: cannot find package "." in:
        /home/runner/work/cilium-cli/cilium-cli/vendor/io/fs
```

Fixes #554

### DIFF
--- a/.github/workflows/multicluster.yaml
+++ b/.github/workflows/multicluster.yaml
@@ -51,6 +51,11 @@ jobs:
         run: |
           gcloud info
 
+      - name: Set up Go
+        uses: actions/setup-go@331ce1d993939866bb63c32c6cbbfd48fa76fc57
+        with:
+          go-version: 1.17.1
+
       - name: Set up job variables
         id: vars
         run: |


nebril (approved):

tklauser:
Multicluster test suceeded with this change: https://github.com/cilium/cilium-cli/actions/runs/1253179656

pchaigno (approved):

PR #558, opened 2021-09-23
test: Reorder tests to mitigate DNS resolution flake

pchaigno:
The first commit ensures tests are executed in the order they are registered. The second commit implements a mitigation for https://github.com/cilium/cilium-cli/issues/367. See commits for details.

I filed https://github.com/cilium/cilium/issues/17459 to keep track of this bug in Cilium.

Fixes: https://github.com/cilium/cilium-cli/issues/367.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -43,7 +43,8 @@ type ConnectivityTest struct {
 	echoServices      map[string]Service
 	externalWorkloads map[string]ExternalWorkload
 
-	tests map[string]*Test
+	tests     []*Test
+	testNames map[string]struct{}
 
 	lastFlowTimestamps map[string]time.Time
 }
@@ -137,7 +138,8 @@ func NewConnectivityTest(client *k8s.Client, p Parameters) (*ConnectivityTest, e
 		clientPods:         make(map[string]Pod),
 		echoServices:       make(map[string]Service),
 		externalWorkloads:  make(map[string]ExternalWorkload),
-		tests:              make(map[string]*Test),
+		tests:              []*Test{},
+		testNames:          make(map[string]struct{}),
 		lastFlowTimestamps: make(map[string]time.Time),
 	}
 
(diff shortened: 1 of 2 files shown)


nbusseneau (approved):
Thanks!

> This is of course not a long term fix since the issue can still happen
in production when users often change their DNS policies. To fix it
properly, we would probably need to implement a grace period in the DNS
proxy during which incorrectly redirected packets are simply forwarded.

Do we already have an issue tracking this?

pchaigno:
> Do we already have an issue tracking this?

No, it's why this PR is still in draft ;-)

connectivity/check/context.go
```diff
@@ -43,7 +43,8 @@ type ConnectivityTest struct {
 	echoServices      map[string]Service
 	externalWorkloads map[string]ExternalWorkload
 
-	tests map[string]*Test
+	tests     []*Test
```
ldelossa, line 46: nit: 

Creating a structure of this maybe a bit nicer to express they work together. 

```
struct TestSet {
  tests []*Test
  set    map[string]struct{}
}
```

Appropriate methods can be written.

ldelossa (changes requested):
Early review since I think this PR will probably stick if we see tests continue to pass. 

The CT flush at the end of the test may become problemsome if we expect users of cilium to run the connectivity test. I think we'd need to communicate that this occurs in some way. 

Also, we need docs which now inform an author of new tests that they need to be added in a specific order and with the work-around in mind.

pchaigno:
> The CT flush at the end of the test may become problemsome if we expect users of cilium to run the connectivity test.

There is no CT flush in this PR.

> Also, we need docs which now inform an author of new tests that they need to be added in a specific order and with the work-around in mind.

I'll had a comment to the suite.go file.

connectivity/check/context.go
pchaigno: I'm not sure this is worth the additional code. There's only one place where we need to edit both at the same time and we're very unlikely to need to change the set in more places.

ldelossa: fair enough.

ldelossa (approved):

tklauser (approved):

tklauser:
Thanks @pchaigno!

PR #564, opened 2021-10-01
Wait correctly for Cilium to be ready before deploying Hubble Relay

### LINKED ISSUE #164: Add --wait support for cilium hubble enable
Similar to `cilium install`, `cilium hubble enable` should block until hubble is enabled successfully. The same `Status()` command can be called as is done for `install.`

### LINKED ISSUE #389: `cilium hubble enable` does not wait correctly for Cilium to be ready
PR https://github.com/cilium/cilium-cli/pull/335 introduced logic to ensure the restarted Cilium pod is ready before Hubble Relay is deployed. However, it seems that this logic does not always work correctly. While using the CLI locally with Kind, I found it immediately declared Cilium to be ready, when it was in fact not (as evident by the `ks get pods` I executed right after).

```console
$ cilium hubble enable
🔑 Found existing CA in secret cilium-ca
✨ Patching ConfigMap cilium-config to enable Hubble...
♻️  Restarted Cilium pods
⌛ Waiting for Cilium to become ready before deploying other Hubble component(s)...
🔑 Generating certificates for Relay...
2021/07/01 11:14:20 [INFO] generate received request
2021/07/01 11:14:20 [INFO] received CSR
2021/07/01 11:14:20 [INFO] generating key: ecdsa-256
2021/07/01 11:14:20 [INFO] encoded CSR
2021/07/01 11:14:20 [INFO] signed certificate with serial number 239843794291536723391962613067911988441539490153
2021/07/01 11:14:20 [INFO] generate received request
2021/07/01 11:14:20 [INFO] received CSR
2021/07/01 11:14:20 [INFO] generating key: ecdsa-256
2021/07/01 11:14:20 [INFO] encoded CSR
2021/07/01 11:14:20 [INFO] signed certificate with serial number 464405006462523752890851327326871415875074304609
✨ Deploying Relay from quay.io/cilium/hubble-relay:v1.9.8...
$ ks get pods -A
NAMESPACE            NAME                                         READY   STATUS              RESTARTS   AGE
kube-system          cilium-5dp2j                                 0/1     Running             0          8s
kube-system          cilium-kwsrq                                 0/1     Running             0          4s
kube-system          cilium-operator-69dcbc4df4-xz7k6             1/1     Running             0          75s
kube-system          cilium-td4f4                                 0/1     Running             0          8s
kube-system          cilium-xgsfd                                 0/1     Running             0          8s
kube-system          coredns-558bd4d5db-84ddw                     1/1     Running             0          111s
kube-system          coredns-558bd4d5db-ggc24                     1/1     Running             0          111s
kube-system          etcd-kind-control-plane                      1/1     Running             0          115s
kube-system          hubble-relay-5b7dff4749-2494t                0/1     ContainerCreating   0          10s
kube-system          kube-apiserver-kind-control-plane            1/1     Running             0          115s
kube-system          kube-controller-manager-kind-control-plane   1/1     Running             0          115s
kube-system          kube-proxy-5nqkn                             1/1     Running             0          94s
kube-system          kube-proxy-bcglx                             1/1     Running             0          94s
kube-system          kube-proxy-gns7h                             1/1     Running             0          111s
kube-system          kube-proxy-tqxfk                             1/1     Running             0          94s
kube-system          kube-scheduler-kind-control-plane            1/1     Running             0          115s
local-path-storage   local-path-provisioner-547f784dff-7q5v5      1/1     Running             0          111s
```

Cilium CLI version:
```
cilium-cli: v0.8.3 compiled with go1.16.5 on linux/amd64
```

tklauser:
Like on install and upgrade, use a `K8sStatusCollector` to wait for Cilium
to be fully ready before deploying Hubble Relay. Moreover, report the
status in case Cilium didn't become ready within the given timeout.
    
The first commit refactors some common code in preparation for the second commit, which implements the actual functionality. The third commit changes `cilium hubble enable` to - by default - wait for the Hubble deployments to be ready (this behavior can be disabled using `--wait=false` as with other commands). The fourth commit drops the now superfluous `cilium status --wait` commands after `cilium hubble enable` from CI.

See individual commit messages for details.

Fixes #164
Fixes #389

### DIFF
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -82,9 +82,6 @@ type Parameters struct {
 	Context          string // Only for 'kubectl' pass-through commands
 	Wait             bool
 	WaitDuration     time.Duration
-	// CiliumReadyTimeout defines the wait timeout for Cilium to become ready after enabling
-	// Hubble before deploying Hubble Relay / UI.
-	CiliumReadyTimeout time.Duration
 }
 
 func (p *Parameters) Log(format string, a ...interface{}) {
@@ -273,17 +270,20 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		return err
 	}
 
+	var dur time.Duration
 	if k.params.Relay || k.params.UI {
+		start := time.Now()
 		k.Log("⌛ Waiting for Cilium to become ready before deploying other Hubble component(s)...")
 		collector, err := status.NewK8sStatusCollector(ctx, k.client, status.K8sStatusParameters{
 			Namespace:       k.params.Namespace,
 			Wait:            true,
-			WaitDuration:    k.params.CiliumReadyTimeout,
+			WaitDuration:    k.params.WaitDuration,
 			WarningFreePods: []string{defaults.AgentDaemonSetName, defaults.OperatorDeploymentName},
 		})
 		if err != nil {
 			return err
 		}
+		dur = time.Since(start)
 
 		s, err := collector.Status(ctx)
 		if err != nil {
@@ -318,7 +318,7 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		collector, err := status.NewK8sStatusCollector(ctx, k.client, status.K8sStatusParameters{
 			Namespace:       k.params.Namespace,
 			Wait:            true,
-			WaitDuration:    k.params.WaitDuration,
+			WaitDuration:    k.params.WaitDuration - dur,
 			WarningFreePods: pods,
 		})
 		if err != nil {
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -274,7 +274,7 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 	if k.params.Relay || k.params.UI {
 		start := time.Now()
 		k.Log("⌛ Waiting for Cilium to become ready before deploying other Hubble component(s)...")
-		collector, err := status.NewK8sStatusCollector(ctx, k.client, status.K8sStatusParameters{
+		collector, err := status.NewK8sStatusCollector(k.client, status.K8sStatusParameters{
 			Namespace:       k.params.Namespace,
 			Wait:            true,
 			WaitDuration:    k.params.WaitDuration,
@@ -315,7 +315,7 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		}
 
 		k.Log("⌛ Waiting for Hubble to be installed...")
-		collector, err := status.NewK8sStatusCollector(ctx, k.client, status.K8sStatusParameters{
+		collector, err := status.NewK8sStatusCollector(k.client, status.K8sStatusParameters{
 			Namespace:       k.params.Namespace,
 			Wait:            true,
 			WaitDuration:    k.params.WaitDuration - dur,
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -13,7 +13,10 @@ import (
 	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/internal/certs"
 	"github.com/cilium/cilium-cli/internal/utils"
+	"github.com/cilium/cilium-cli/status"
 
+	"github.com/cilium/cilium/api/v1/models"
+	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -50,6 +53,10 @@ type k8sHubbleImplementation interface {
 	CreateService(ctx context.Context, namespace string, service *corev1.Service, opts metav1.CreateOptions) (*corev1.Service, error)
 	DeleteService(ctx context.Context, namespace, name string, opts metav1.DeleteOptions) error
 	DeletePodCollection(ctx context.Context, namespace string, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error
+	ListPods(ctx context.Context, namespace string, options metav1.ListOptions) (*corev1.PodList, error)
+	GetDaemonSet(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*appsv1.DaemonSet, error)
+	CiliumStatus(ctx context.Context, namespace, pod string) (*models.StatusResponse, error)
+	ListCiliumEndpoints(ctx context.Context, namespace string, opts metav1.ListOptions) (*ciliumv2.CiliumEndpointList, error)
 	CheckDaemonSetStatus(ctx context.Context, namespace, daemonset string) error
 	GetRunningCiliumVersion(ctx context.Context, namespace string) (string, error)
 }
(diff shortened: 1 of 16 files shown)


tklauser:
> Like on install and upgrade, use a `K8sStatusCollector` to wait for Cilium to be fully ready before deploying Hubble Relay. Moreover, report the status in case Cilium didn't become ready within the given timeout.
> 
> The first commit refactors some common code in preparation for the second commit, which implements the actual functionality. The third commit drops the now superfluous `cilium status --wait` commands after `cilium hubble enable` from CI.

This is not yet enough to drop `cilium status --wait` from CI (3rd commit). We also need to wait for the Hubble Relay Deployment to be ready after `cilium hubble enable`. I'll add that check.

tklauser:
> > Like on install and upgrade, use a `K8sStatusCollector` to wait for Cilium to be fully ready before deploying Hubble Relay. Moreover, report the status in case Cilium didn't become ready within the given timeout.
> > The first commit refactors some common code in preparation for the second commit, which implements the actual functionality. The third commit drops the now superfluous `cilium status --wait` commands after `cilium hubble enable` from CI.
> 
> This is not yet enough to drop `cilium status --wait` from CI (3rd commit). We also need to wait for the Hubble Relay Deployment to be ready after `cilium hubble enable`. I'll add that check.

Fixed in what now is the third commit:

    hubble: wait for deployments to be ready after enabling them
    
    By default, wait for the Relay (and UI, if `--ui` is specified)
    deployments to be ready as part of enabling Hubble. This behavior can be
    disabled by setting `--wait=false`.
    
    In combination with the previous commit this allows to avoid invoking
    `cilium status --wait` after enabling Hubble.

internal/cli/cmd/hubble.go
```diff
@@ -56,6 +56,8 @@ func newCmdHubbleEnable() *cobra.Command {
 	cmd.Flags().BoolVar(&params.UI, "ui", false, "Enable Hubble UI")
 	cmd.Flags().BoolVar(&params.CreateCA, "create-ca", false, "Automatically create CA if needed")
 	cmd.Flags().StringVar(&contextName, "context", "", "Kubernetes configuration context")
+	cmd.Flags().BoolVar(&params.Wait, "wait", true, "Wait for status to report success (no errors)")
+	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
```
joestringer, line 60: Do you have any thoughts on why this needs to be a dedicated CLI flag separate from `cilium-ready-timeout` and also why 15 minutes? I think it's easy from development side to say "well let's just add another flag here" but over time, there's more and more complicated different flags specific to particular aspects of the implementation that make it increasingly difficult for users to understand or use properly. 

Broadly what I would expect from this command is a generic "spend X minutes on this and fail out if the entire operation takes that long". This makes the behaviour reliable from a user perspective, and also means that if something gets stuck in CI then we're not randomly waiting 15-20 minutes for the CLI to decide that it really won't make any progress.

joestringer (changes requested):
LGTM overall, just one specific nit around user-facing flags / wait timers below.

install/install.go
```diff
@@ -1688,9 +1688,7 @@ func (k *K8sInstaller) Install(ctx context.Context) error {
 
 		s, err := collector.Status(ctx)
 		if err != nil {
-			if s != nil {
-				fmt.Println(s.Format())
-			}
+			fmt.Println(s.Format())
```
gandro, line 1691: This will now print an empty line instead of nothing at all. Is that intended?

internal/cli/cmd/hubble.go
tklauser: Good point. With this change, I don't think we need a dedicated `cilium-ready-timeout` flag anymore. 15 minutes for `wait` is probably too much, I agree. I think I just blindly copied this from another command. Seems like all other commands exposing `wait-duration` set it to 15 minutes though:

```
% git grep "\"wait-duration\""
internal/cli/cmd/clustermesh.go:165:    cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
internal/cli/cmd/clustermesh.go:300:    cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
internal/cli/cmd/install.go:64: cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
internal/cli/cmd/install.go:148:        cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
internal/cli/cmd/status.go:42:  cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait for status")
```

My suggestion would be to deprecate `cilium-ready-timeout` (i.e. mark it as hidden for 0.9.2 and remove it for 0.9.3 or 0.9.4) and introduce `wait` and `wait-timeout` with a default of 15 minutes (to be constistent with other commands). That timeout would then include waiting for Cilium to become ready and waiting for the Hubble deployment. The hidden/deprecated `cilium-ready-timeout` would just be an alias for `wait`.

Regarding the 15 minutes timeout, I'd propose to adjust this for all commands in a separate PR in case we decide that this is too long to wait.

How does that sound?

gandro (approved):
Overall looks good to me! Thanks! One minor question

install/install.go
tklauser: Well spotted! That is not intended. Will change to `fmt.Print` and add an additional newline to `s.Format()`s return value.

internal/cli/cmd/hubble.go
tklauser: I've added a separate commit implementing this. Commit message:

    hubble: deprecate and hide `cilium-ready-timeout` flag

    With the introduction of the `--wait-duration` flag in the previous
    commit, we now have two flags specifying a timeout to wait for a
    particular aspect of the `hubble enable` command. As Joe points out,
    this makes it increasingly difficult for users to understand or use
    properly. To simplify the user experience, deprecate
    `cilium-ready-timeout` and make `wait-duration` the total timeout to
    wait for Hubble to be enabled. For backwards compatibility
    `cilium-ready-timeout` will be retained until release 0.9.3 as an alias
    for `wait-duration`.

tklauser:
All CI passed:

GKE: https://github.com/cilium/cilium-cli/actions/runs/1303339705
EKS (ENI): https://github.com/cilium/cilium-cli/actions/runs/1303339703
EKS (tunnel): https://github.com/cilium/cilium-cli/actions/runs/1303339712
Multicluster: https://github.com/cilium/cilium-cli/actions/runs/1303339602
External workloads: https://github.com/cilium/cilium-cli/actions/runs/1303339707

internal/cli/cmd/hubble.go
joestringer: Personally I think we should revisit the 15m but maybe there's some case we're trying to account for with a longer time like that that I'm just missing. :+1:

joestringer (approved):

internal/cli/cmd/hubble.go
tklauser: Agree, let's revisit in a separate PR. What do you think would be a reasonable timeout? 5m? 10m?

hubble/hubble.go
```diff
@@ -318,7 +318,7 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		collector, err := status.NewK8sStatusCollector(ctx, k.client, status.K8sStatusParameters{
 			Namespace:       k.params.Namespace,
 			Wait:            true,
-			WaitDuration:    k.params.WaitDuration,
+			WaitDuration:    k.params.WaitDuration - dur,
```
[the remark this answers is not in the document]
tklauser: Good point. It looks like the `(*K8sStatusCollector).Status` configures the timeout using `context.WithTimeout`, so we would probably need to refactor a bit such that the context with timeout is passed at each call site and omit the `WaitDuration` parameter altogether.

I'll look into that and send a follow-up PR.

internal/cli/cmd/hubble.go
joestringer: 5m seems plenty to me, typically in my test clusters Cilium is up & ready within ~1m or so. I wonder whether we could get some additional datapoints maybe from CI or from others in the community.

PR #571, opened 2021-10-06
sysdump: Implement AddTasks method

michi-covalent:
probably easier to review commit by commit.


sysdump/sysdump.go
```diff
@@ -162,7 +165,7 @@ func (c *Collector) Run() error {
 			Description: "Collect Kubernetes nodes",
 			Quick:       true,
 			Task: func(_ context.Context) error {
-				if err := writeYaml(absoluteTempPath(kubernetesNodesFileName), n); err != nil {
+				if err := writeYaml(c.AbsoluteTempPath(kubernetesNodesFileName), n); err != nil {
```
tklauser: While at it, instead of having all these `writeYaml(c.AbsoluteTempPath(...))` calls could we define `(*Collector).writeYAML` as follows and then call that?

```go
func (c *Collector) writeYAML(p path, o runtime.Object) error {
	return writeYaml(c.AbsoluteTempPath(p), o)
}
```

sysdump/sysdump.go
```diff
@@ -24,7 +24,7 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
-// Options groups together the set of options required to collect a sysdump.
+// Options groups together the set of Options required to collect a sysdump.
```
tklauser, line 27: Nit: I think we can keep that comment as is.

sysdump/sysdump.go
```diff
@@ -98,14 +99,15 @@ type Collector struct {
 	// allNodes is a list of all the node names in the cluster.
 	allNodes *corev1.NodeList
 	// NodeList is a list of nodes to collect sysdump information from.
-	NodeList []string
+	NodeList        []string
```
tklauser, line 102: Nit: maybe move that up to the other exported fields, e.g. below `Pool`?

sysdump/sysdump.go
```diff
@@ -808,7 +813,7 @@ func (c *Collector) Run() error {
 	}
 
 	if loggedStart {
-		c.logWarn("Please note that depending on your Cilium version and installation options, this may be expected")
+		c.logWarn("Please note that depending on your Cilium version and installation Options, this may be expected")
```
tklauser, line 816: Nit: I don't think we want to change `s/options/Options` here and in other error and log messages.

sysdump/sysdump.go
```diff
@@ -772,27 +777,27 @@ func (c *Collector) Run() error {
...
+		if err := c.Pool.Submit(fmt.Sprintf("[%d] %s", i, t.Description), func(ctx context.Context) error {
 			if t.CreatesSubtasks {
 				defer c.subtasksWg.Done()
 			}
 			c.logTask(t.Description)
 			defer c.logDebug("Finished %q", t.Description)
 			return t.Task(ctx)
 		}); err != nil {
-			return fmt.Errorf("failed to submit task to the worker pool: %w", err)
+			return fmt.Errorf("failed to submit task to the worker Pool: %w", err)
```
tklauser, line 788: Nit: I don't think we want to change `s/pool/Pool` here and in other error and log messages.

sysdump/sysdump.go
```diff
@@ -756,13 +761,13 @@ func (c *Collector) Run() error {
...
+	if wc < c.Options.WorkerCount {
+		wc = c.Options.WorkerCount
 	}
-	c.pool = workerpool.New(wc)
-	c.logDebug("Using %d workers (requested: %d)", wc, c.options.WorkerCount)
+	c.Pool = workerpool.New(wc)
+	c.logDebug("Using %d workers (requested: %d)", wc, c.Options.WorkerCount)
 
-	// Add the tasks to the worker pool.
+	// Add the tasks to the worker Pool.
```
tklauser, line 770: Nit: I don't think we want to change s/pool/Pool here.

sysdump/sysdump_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/cilium/cilium-cli/internal/k8s"
+	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	ciliumv2alpha1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
+	"gopkg.in/check.v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
tklauser, line 23: Nit: split `cilium-cli` imports from other 3rd party imports in a separate import group:

```suggestion
	"github.com/cilium/cilium-cli/internal/k8s"

	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
	ciliumv2alpha1 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2alpha1"
	"gopkg.in/check.v1"
	appsv1 "k8s.io/api/apps/v1"
	corev1 "k8s.io/api/core/v1"
	networkingv1 "k8s.io/api/networking/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/runtime/schema"
```

tklauser (commented):
LGTM apart from a few nits. I'd also suggest to splt up the last commit (if possible) even further into one commit exporting all necessary funcs, struct members etc. and one commit adding the `AddTasks` method. This would make it a bit easier to review.

It also seems like the exporting of members was done using search/replace and led to some comments, error messages and log messages being changed as well (e.g. `s/options/Options`). I pointed out some of those but might have missed some. Would be nice to revert these IMO unnecessary changes.

Issue #569, opened 2021-10-05
Service load-balancing not working in external workloads


wazir-ahmed:
### Problem
The external workload script generated by Cilium CLI (`cilium clustermesh vm install <file-name>`), does not handle the fact that recent versions of docker creates a new cgroup namespace for the containers. Because of this, bpf programs responsible for service load-balancing are getting attached in the container's cgroup, instead of host's root cgroup and the service load-balancing in external workload is not working.

### Solution
We can change the external workload script in one of the following ways.

1. In the host, we could mount the `cgroup2` in `/run/cilium/cgroupv2` and bind mount the directory into the container using docker's `--mount` option. Similar to https://github.com/cilium/cilium/pull/16259
2. We could use docker's `--cgroupns=host` option with docker run to avoid creation of a new cgroup namespace for the container.

**Cilium CLI version - v0.9.0**

aditighag:
Per the Slack thread, this seems to be a dup of https://github.com/cilium/cilium-cli/issues/256 as you mentioned that the clustermesh script uses cilium CLI to install cilium underneath.

aditighag:
@wazir-ahmed brought this issue in the community meeting, and he said he'll be contributing a PR. Sorry for the confusion as I misinterpreted your comment when you said that the script uses the CLI to install cilium.

wazir-ahmed:
[current-cew-script.txt](https://github.com/cilium/cilium-cli/files/7295545/current-cew-script.txt)

This is a sample script file generated by `cilium clustermesh vm install <file-name>`.

wazir-ahmed:
The script generated by the CLI needs to incorporate the following changes.

```
--- current-cew-script.sh       2021-10-06 21:25:00.145321359 +0530
+++ patched-cew-script.sh       2021-10-06 21:25:05.719563566 +0530
@@ -17,6 +17,11 @@
         echo "Shutting down running Cilium agent"
         ${SUDO} docker rm -f cilium || true
     fi
+    if [ -d /run/cilium/cgroupv2 ] ; then
+        echo "Removing mountpoint /run/cilium/cgroupv2"
+        ${SUDO} umount /run/cilium/cgroupv2
+        ${SUDO} rm -d /run/cilium/cgroupv2
+    fi
     if [ -f /usr/bin/cilium ] ; then
         echo "Removing /usr/bin/cilium"
         ${SUDO} rm /usr/bin/cilium
@@ -129,6 +134,9 @@
[...]

PR #572, opened 2021-10-07
external workload: Run cilium-agent in host cgroup namespace

wazir-ahmed:
Recent versions of Docker create a new cgroup namespace for each container.
Add Docker CLI option `--cgroupns=host` to the external workload installation
script (generated by `cilium clustermesh vm install`) so that the cilium
agent will run in host cgroup namespace and service load-balancing will work
as expected in external workloads.

Fixes: #569

Signed-off-by: Wazir Ahmed <wazir@accuknox.com>

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -1420,6 +1420,10 @@ fi
 
 DOCKER_OPTS=" -d --log-driver local --restart always"
 DOCKER_OPTS+=" --privileged --network host --cap-add NET_ADMIN --cap-add SYS_MODULE"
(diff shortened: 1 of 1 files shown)


aditighag:
External workloads workflow has successfully completed so marking it ready for merge.

ldelossa:
@wazir-ahmed

if I'm understanding this change correctly, you are mounting the host's cgroup into the agent container, so the container can see the entire tree of cgroup namespaces created by docker? 

sanity check me: what occurs when this script runs on a host without cgroup_v2 enabled?

PR #584, opened 2021-10-15
workflows: update AKS workflows with new taints

nbusseneau:
Re-impacted from: https://github.com/cilium/cilium/pull/17529

Context: we recommend users taint all nodepools with `node.cilium.io/agent-not-ready=true:NoSchedule` to prevent application pods from being managed by the default AKS CNI plugin.

To this end, the proposed workflow users should follow when installing Cilium into AKS was to replace the initial AKS node pool with a new tainted system node pool, as it is not possible to taint the initial AKS node pool, cf. Azure/AKS#1402.

AKS recently pushed a change on the API side that forbids setting up custom taints on system node pools, cf. Azure/AKS#2578.

It is not possible anymore for us to recommend users taint all nodepools with `node.cilium.io/agent-not-ready=true:NoSchedule` to prevent application pods from being managed by the default AKS CNI plugin.

To work around this new limitation, we propose the following workflow instead:

- Replace the initial node pool with a system node pool tainted with `CriticalAddonsOnly=true:NoSchedule`, preventing application pods from being scheduled on it.
- Create a secondary user node pool tainted with `node.cilium.io/agent-not-ready=true:NoSchedule` to prevent application pods from being scheduled on the user node pool until Cilium is ready to manage them.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -28,6 +28,7 @@ concurrency:
 env:
   name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
   location: westeurope
+  cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
 
 jobs:
   installation-and-connectivity:
@@ -69,7 +70,6 @@ jobs:
           echo ::set-output name=owner::${OWNER}
 
       - name: Create AKS cluster
-        id: cluster-creation
         run: |
           # Create group
           az group create \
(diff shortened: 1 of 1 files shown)


nbusseneau:
@tklauser Weird issue while trying to run `cilium install` -- seems consistent:

```
2021-10-15T14:44:56.791125839Z 🔑 Generating certificates for Hubble...
2021-10-15T14:44:56.815733603Z 
2021-10-15T14:44:56.815775605Z Error: Unable to install Cilium: unable to create secret kube-system/hubble-server-certs: secrets "hubble-server-certs" already exists
2021-10-15T14:44:56.815802205Z ↩️ Rolling back installation...
```

This does not ring a bell to me: why would there already be a `hubble-server-certs` before install?

nbusseneau:
> Should we push this change to cilium/cilium as well for consistency?

I would personally prefer not to:

- It is not necessary on `cilium/cilium` because we don't use in-cluster scripts, so it would only slow down `cilium/cilium` AKS testing (not by much, but we probably gain something like 30 seconds out of it because the Cilium components get installed earlier).
- Regarding consistency, I think keeping the test code consistent with the way users are supposed to install Cilium whenever possible is more important, and this would be with `--no-wait` as it stands right now.

> Probably worth a comment in the code, since reading the rest, one would expect a `--no-wait` in the delete command as well.

PR #605, opened 2021-11-04
add arm64 suffix to etcd image when CPU arch is arm64.

dereknex:
Fixes: cilium/cilium#17779

Signed-off-by: Derek Chen derek_chen@live.com


clustermesh/clustermesh.go
```diff
@@ -156,6 +157,39 @@ func (k *K8sClusterMesh) apiserverImage() string {
 	return utils.BuildImagePath(k.params.ApiserverImage, defaults.ClusterMeshApiserverImage, k.params.ApiserverVersion, k.imageVersion)
 }
 
+func (k *K8sClusterMesh) etcdImage() string {
+	etcdVersion := "v3.4.13"
+	arch := runtime.GOARCH
```
tklauser, line 162: This will give us the `GOARCH` for the host the Cilium CLI is running. I think we instead should to check the `GOARCH` for the cluster where Cilium is deployed.

clustermesh/clustermesh.go
```diff
@@ -156,6 +157,39 @@ func (k *K8sClusterMesh) apiserverImage() string {
...
+		{
+			Name: "HOSTNAME_IP",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "status.podIP",
+				},
+			},
+		},
+	}
+	if runtime.GOARCH == "arm64" {
```
tklauser, line 184: Same here, this should check the `GOARCH` for the cluster where Cilium is deployed.

tklauser (changes requested):

clustermesh/clustermesh.go
dereknex: Because we use the `quay.io/coreos/etcd: 3.4.13` that it isn't multi-arch image. So k8s will pull `amd64` image. It can't run on  arm64 server. And i have got your point. Do you think adding param to the `K8sClusterMesh.params` is a good idea?

tklauser: I don't think this needs a dedicated parameter. Instead we should be able to get the target architecture from the k8s server version information (`apimachinery/pkg/version.Info` in the `Platform` field which contains the `GOOS`/`GOARCH` pair). This is already queried here:

https://github.com/cilium/cilium-cli/blob/f277e6540e7607d67d4616b90f19eed6e22fb018/internal/k8s/client.go#L555-L561

so I think you could adapt that method (and all its callers) to return the `version.Info` struct instead and then extract the arch from there.

dereknex: Okay, thank you. I will fix it later.

tklauser (approved):

tklauser:
Thanks @dereknex!

anthr76:
Shouldn't a "fat" manifest exist for the etcd image? This seems like it's a little backwards..

dereknex:
> Shouldn't a "fat" manifest exist for the etcd image? This seems like it's a little backwards..

The `etcd:v3.5.1` will solve this issue, but we need upgrade etcd to v3.5.

nathanjsweet (approved):

michi-covalent (approved):

Issue #600, opened 2021-11-02
CI: Multicluster: flow validation failures in no-policies/pod-to-service/curl-0


michi-covalent:
## CI failure

this is similar to https://github.com/cilium/cilium-cli/issues/419, but in this case, only the initial SYN from the client to echo-other-node is observed, and nothing else.

- Copy-paste the output the test failure.

```
2021-11-02T15:17:52.510288664Z   ℹ️  SYN and(ip(src=10.116.0.84),or(tcp(dstPort=8080),tcp(dstPort=8080)),tcpflags(syn)) not found
2021-11-02T15:17:52.510295476Z   ❌ Aborting flow matching: context deadline exceeded
2021-11-02T15:17:52.510330371Z   ❌ Flow validation failed for peer cilium-test/client-6488dcf5d4-g4bfp: 1 failures (first: 0, last: 8, matched: 2)
2021-11-02T15:17:52.510346171Z   
2021-11-02T15:17:52.510353504Z   📄 Flow logs for peer cilium-test/client-6488dcf5d4-g4bfp:
2021-11-02T15:17:52.510370224Z   ❓ [0] Nov  2 15:17:37.501: cilium-test/client-6488dcf5d4-g4bfp:34203 -> kube-system/kube-dns-b4f5c58c7-b5nx8:53 from-stack FORWARDED (UDP)
2021-11-02T15:17:52.510408826Z   ❓ [1] Nov  2 15:17:37.501: cilium-test/client-6488dcf5d4-g4bfp:34203 -> kube-system/kube-dns-b4f5c58c7-b5nx8:53 to-endpoint FORWARDED (UDP)
2021-11-02T15:17:52.510425980Z   ❓ [2] Nov  2 15:17:37.501: cilium-test/client-6488dcf5d4-g4bfp:34203 -> kube-system/kube-dns-b4f5c58c7-b5nx8:53 from-stack FORWARDED (UDP)
2021-11-02T15:17:52.510524477Z   ❓ [3] Nov  2 15:17:37.501: cilium-test/client-6488dcf5d4-g4bfp:34203 -> kube-system/kube-dns-b4f5c58c7-b5nx8:53 to-endpoint FORWARDED (UDP)
2021-11-02T15:17:52.510541266Z   ❓ [4] Nov  2 15:17:37.502: cilium-test/client-6488dcf5d4-g4bfp:34203 -> kube-system/kube-dns:53 from-endpoint FORWARDED (UDP)
[...]

PR #609, opened 2021-11-08
install: Disable BPF masquerading by default

pchaigno:
In commit bf2c1f99, we switched `kube-proxy-replacement` to disabled by default. We didn't change the default setting for BPF masquerading, even though it requires `kube-proxy-replacement` to be enabled. As a result, default installation result in a warning when BPF masquerading is forcefully disabled.

This pull request disables BPF masquerading by default to fix it. We could enable BPF masquerading when `kube-proxy-replacement` is enabled by the user, but it seems more confusing than anything.

Fixes: https://github.com/cilium/cilium-cli/pull/136.

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -1264,7 +1264,7 @@ func (k *K8sInstaller) generateConfigMap() (*corev1.ConfigMap, error) {
 			// wait-bpf-mount makes init container wait until bpf filesystem is mounted
 			"wait-bpf-mount": "true",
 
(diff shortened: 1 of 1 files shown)


tklauser (approved):

pchaigno:
Multicluster hit known flake https://github.com/cilium/cilium-cli/issues/600. Other tests are passing and Tobias reviewed. Marking ready to merge.

PR #632, opened 2021-11-26
install: replace --native-routing-cidr by --ipv4-native-routing-cidr

kaworu:
Cilium v1.11 [has deprecated --native-routing-cidr in favor of a new --ipv4-native-routing-cidr flag][1]. This patch bring the same change in the Cilium CLI.

[1]: https://github.com/cilium/cilium/pull/16646/commits/a66b2f2824522c3129a27552cee435ac7ce11135

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
@@ -17,6 +18,9 @@ import (
 
 func newCmdInstall() *cobra.Command {
 	var params = install.Parameters{Writer: os.Stdout}
+	var deprecated struct {
+		NativeRoutingCIDR string
+	}
 
 	cmd := &cobra.Command{
 		Use:   "install",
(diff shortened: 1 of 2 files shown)


internal/cli/cmd/install.go
```diff
@@ -55,7 +55,9 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().MarkHidden("base-version")
 	cmd.Flags().StringVar(&params.DatapathMode, "datapath-mode", "", "Datapath mode to use")
 	cmd.Flags().StringVar(&params.IPAM, "ipam", "", "IP Address Management (IPAM) mode")
-	cmd.Flags().StringVar(&params.NativeRoutingCIDR, "native-routing-cidr", "", "CIDR within which native routing is possible")
+	cmd.Flags().StringVar(&params.IPv4NativeRoutingCIDR, "ipv4-native-routing-cidr", "", "IPv4 CIDR within which native routing is possible")
+	cmd.Flags().StringVar(&params.NativeRoutingCIDR, "native-routing-cidr", "", "IPv4 CIDR within which native routing is possible. Deprecated in favor of --ipv4-native-routing-cidr")
+	cmd.Flags().MarkDeprecated("native-routing-cidr", "use --ipv4-native-routing-cidr instead")
```
tklauser, line 60: Should we also hide this flag from usage messages, i.e. `cmd.Flags().MarkHidden("native-routing-cidr")`?

tklauser (changes requested):
I think we also need to handle the new `IPv4NativeRoutingCIDR` here:

https://github.com/cilium/cilium-cli/blob/daf109dc1da2e002c3507d45ea020d2960404ad4/install/install.go#L1546-L1554

How about instead of having both `NativeRoutingCIDR` and `IPv4NativeRoutingCIDR` members in `install.Parameters`, we just keep IPv4NativeRoutingCIDR` and set it based on either one of the CLI flags before calling (*K8sInstaller).generateConfigMap` or `(*K8sInstaller).Install`. This would also have the advantage that we can report an error about the user specifying both flags at an earlier point.

internal/cli/cmd/install.go
kaworu: [MarkDeprecated](https://github.com/cilium/cilium-cli/blob/2206e30ed9e81777413e047735602b1871fc1d5a/vendor/github.com/spf13/pflag/flag.go#L408-L422) will hide it.

tklauser: TIL, thanks!

tklauser (approved):
Thanks :rocket:

tklauser:
Multicluster test failure looks like #600. All other tests passed, merging.

PR #638, opened 2021-12-01
.github: let dependabot ignore Cilium and Hubble dependencies

rolinh:
To properly update Cilium and Hubble, the replace directives need to be updated as well, something the bot can't do.


tklauser (approved):
Thanks!

PR #641, opened 2021-12-02
upgrade: determine correct operator image on AKS/EKS

### LINKED ISSUE #407: Upgrading via `cilium upgrade` on Azure / AKS uses the generic Operator image instead of the `-azure` flavor
I suspect this to also occur on Amazon EKS and other cloud-managed services. The upgrade command needs autodetection logic as well.

tklauser:
Use the correct operator image for `cilium upgrade` on AKS or EKS. Also fix a duplicate log message and a log message typo.

See commit messages for details

Fixes #407

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -140,9 +140,7 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
 		if k.params.DatapathMode != "" {
 			k.Log("🔮 Auto-detected datapath mode: %s", k.params.DatapathMode)
 		}
-	}
-
-	if k.params.DatapathMode != "" {
+	} else {
 		k.Log("🔮 Custom datapath mode: %s", k.params.DatapathMode)
 	}
 
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -57,7 +57,41 @@ func (k *K8sUninstaller) autodetect(ctx context.Context) error {
 	return nil
 }
 
-func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
+func (k *K8sInstaller) detectDatapathMode(withKPR bool) {
+	if k.params.DatapathMode != "" {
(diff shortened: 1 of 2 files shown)


tklauser:
EKS failure is real, we need to run with `operator-generic` in tunnel mode regardless of cloud provider kind.

tklauser:
> EKS failure is real, we need to run with `operator-generic` in tunnel mode regardless of cloud provider kind.

Fixed by determining datapath mode on upgrade as well (3rd commit).

install/autodetect.go
```diff
@@ -57,7 +57,40 @@ func (k *K8sUninstaller) autodetect(ctx context.Context) error {
 	return nil
 }
 
-func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
+func (k *K8sInstaller) detectDatapathMode(withKPR bool) {
+	if k.params.DatapathMode == "" {
```
ldelossa, line 61: Instead of this if/else, can you just early return if DataPath  mode is set? Should type out a bit cleaner.

install/autodetect.go
```diff
@@ -69,12 +102,22 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
...
 
+	return nil
+}
+
+func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
+	if err := k.autodetect(ctx); err != nil {
+		return err
+	}
+
+	f := k.flavor
```
ldelossa, line 113: Why take this copy? Doesn't look like we mutate `f`

ldelossa (changes requested):

christarazi (approved):
LGTM pending Louis's comments

install/autodetect.go
tklauser: The initial idea was to not change too much about the current code. Agree that the copy seems unnecessary, will change.

tklauser:
Multicluster test flaked on #600 and all other tests are green, merging.

PR #645, opened 2021-12-06
hubble: Make UI version configurable

### LINKED ISSUE #628: cilium CLI installs hardcoded, old version of Hubble UI
`cilium hubble enable --ui` installs a hard-coded version v0.7.9 of the Hubble UI ([here](https://github.com/cilium/cilium-cli/blob/79f04377d5b93586a822902d54aa0116d592f253/hubble/ui.go#L191)) for the hubble-ui container image, and similarly for the hubble-ui-backend image. 

Would be better if this was handled like other components, and if the user could override the version number.

twpayne:
Fixes #628.

Signed-off-by: Tom Payne <tom@isovalent.com>


twpayne:
CI failure looks like a recurrence of #367.

michi-covalent (approved):
:ship:

hubble/ui.go
```diff
@@ -188,7 +188,7 @@ func (k *K8sHubble) generateHubbleUIDeployment() *appsv1.Deployment {
 					Containers: []corev1.Container{
 						{
 							Name:            "frontend",
-							Image:           "quay.io/cilium/hubble-ui:v0.8.3@sha256:018ed122968de658d8874e2982fa6b3a8ae64b43d2356c05f977004176a89310",
```
tklauser, line 191: Not blocking this PR, but maybe as a follow-up: It would be nice to keep pinning the default image SHA256 sums. We should probably also add them for all other default images which don't currently have them.

tklauser: Created #648

PR #660, opened 2021-12-13
images: Use well-known digests where possible

### LINKED ISSUE #648: Pin default image version with SHA256 sums
_Extracted from https://github.com/cilium/cilium-cli/pull/645#discussion_r763852292_

It would be nice to pin SHA256 sums for all default image versions. Currently, we're only using tags which might change without us noticing.

The following images should be pinned using their respective SHA256 sum:

- [ ] quay.io/cilium/cilium, 
- [ ] quay.io/cilium/operator-generic
- [ ] quay.io/cilium/operator-aws
- [ ] quay.io/cilium/operator-azure
- [ ] quay.io/cilium/hubble-relay
- [ ] quay.io/cilium/hubble-ui
- [ ] quay.io/cilium/hubble-ui-backend
- [ ] quay.io/cilium/clustermesh-apiserver

twpayne:
Fixes #648.


defaults/defaults.go
```diff
@@ -116,3 +116,40 @@ var (
 	// CiliumPodSelector is the pod selector to be used for the Cilium agents.
 	CiliumPodSelector = "k8s-app=cilium"
 )
+
+// WellKnownImageDigests maps well known image paths and tags to their digests.
+// Do not remove old versions.
+var WellKnownImageDigests = map[string]string{
```
aanm, line 122: What do you think on having this in a file and load it with `go:embed`?

ldelossa (approved):

defaults/defaults.go
twpayne: Nice idea, done.

defaults/imagedigests.json
```diff
@@ -0,0 +1,43 @@
+{
+  "quay.io/cilium/cilium": {
+    "v1.8.13": "@sha256:070a57faa72ca55b045861453a2f1697e4d582a75cf2b24937e0397684abcb3f",
+    "v1.9.11": "@sha256:47f923325069a697d5baf5314c7fe936bdf34e7c8154666e6762a78be1ddc3ec",
+    "v1.10.5": "@sha256:0612218e28288db360c63677c09fafa2d17edda4f13867bcabf87056046b33bb",
```
tklauser, line 5: Should we switch this to 1.10.6 which was released last week? Same for the `operator-*`, `clustermesh-apiserver` and `hubble-relay` images.

defaults/imagedigests.go
```diff
@@ -0,0 +1,24 @@
+package defaults
+
+import (
+	_ "embed"
+	"encoding/json"
+)
+
+//go:embed imagedigests.json
```
tklauser, line 8: Doesn't need to be part of this PR, but maybe an idea for a follow-up: Would be nice to have a script/Go program which could be used in a `//go:generate` directive to update `imagedigest.json` to the latest versions/digests per release branch.

defaults/imagedigests.json
twpayne: Added.

defaults/imagedigests.go
twpayne: I've added a manual command to update this now. For `//go:generate` we need a method to find new versions.

tklauser (approved):
Thanks!

Issue #616, opened 2021-11-18
Bugtool output truncated or absent from sysdump


bmcustodio:
Sometimes sysdumps will be either entirely missing a given `cilium-bugtool-*` tar file, or contain an incomplete one. It's unclear if this is the same issue or two different ones, but one of them has been observed to be caused by an `unexpected EOF` error that also happens during `kubectl cp` (which is the same mechanism used to transfer the bugtool tar from the Cilium pods to the local machine):

```
I1118 16:22:32.676181   48429 connection.go:198] SPDY Ping failed: connection closed
F1118 16:22:32.676629   48429 helpers.go:116] error: unexpected EOF
goroutine 1 [running]:
k8s.io/kubernetes/vendor/k8s.io/klog/v2.stacks(0xc00013c001, 0xc0005a4640, 0x44, 0x91)
        /private/tmp/kubernetes-cli-20211118-19721-pemzfc/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/klog/v2/klog.go:1026 +0xb9
k8s.io/kubernetes/vendor/k8s.io/klog/v2.(*loggingT).output(0x6e0d540, 0xc000000003, 0x0, 0x0, 0xc0004f4310, 0x2, 0x64e8137, 0xa, 0x74, 0x400ff00)
        /private/tmp/kubernetes-cli-20211118-19721-pemzfc/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/klog/v2/klog.go:975 +0x1e5
k8s.io/kubernetes/vendor/k8s.io/klog/v2.(*loggingT).printDepth(0x6e0d540, 0xc000000003, 0x0, 0x0, 0x0, 0x0, 0x2, 0xc000712060, 0x1, 0x1)
        /private/tmp/kubernetes-cli-20211118-19721-pemzfc/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/klog/v2/klog.go:735 +0x185
k8s.io/kubernetes/vendor/k8s.io/klog/v2.FatalDepth(...)
        /private/tmp/kubernetes-cli-20211118-19721-pemzfc/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/klog/v2/klog.go:1500
k8s.io/kubernetes/vendor/k8s.io/kubectl/pkg/cmd/util.fatal(0xc000cb40a8, 0x15, 0x1)
[...]

PR #671, opened 2021-12-24
bugtool: Compress and copy with resume

sayboras:
### Description

This PR tries to address the below points mentioned in #616 

- Compress bugtool output before transfer
- Add resume capability if there is any transfer error

### Changes

Please refer to individual commit for more details

### Testing

Testing was done as per below

<details>
<summary>Make sure that the existing functionalities still working</summary>

```
$ ls -lrth cilium-sysdump-20211224-183*
-rw-rw-r-- 1 tammach tammach 7.7M Dec 24 18:34 cilium-sysdump-20211224-183404.zip --> from master branch
-rw-rw-r-- 1 tammach tammach 7.7M Dec 24 18:34 cilium-sysdump-20211224-183417.zip --> from this branch
```

</details>

<details>
<summary>Testing with a very large file (~10GB before compressed, 50MB after compressed)</summary>

```
# Manually create the file in cilium pod
dd if=/dev/zero of=filename bs=1 count=0 seek=50G
tar -czvf cilium-bugtool-every-large-file.tar.gz filename

# hard code the file name, and run cilium sysdump
$ ls -lrth cilium-sysdump-20211224-200925.zip                            
-rw-rw-r-- 1 tammach tammach 100M Dec 24 20:09 cilium-sysdump-20211224-200925.zip
```

</details>

### DIFF
--- a/internal/k8s/copy.go
+++ b/internal/k8s/copy.go
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 Authors of Cilium
+
+package k8s
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+)
+
+const (
+	defaultReadFromByteCmd = "tail -c+%d %s"
+	defaultMaxTries        = 5
+)
+
+// CopyFromPod is to copy srcFile in a given pod to local destFile with defaultMaxTries.
+func (c *Client) CopyFromPod(ctx context.Context, namespace, pod, container string, srcFile, destFile string) error {
+	pipe := newPipe(&CopyOptions{
+		MaxTries: defaultMaxTries,
+		ReadFunc: readFromPod(ctx, c, namespace, pod, container, srcFile),
+	})
+
+	outFile, err := os.Create(destFile)
+	if err != nil {
+		return err
+	}
+
+	if _, err = io.Copy(outFile, pipe); err != nil {
+		return err
+	}
+	return nil
+}
+
+func readFromPod(ctx context.Context, client *Client, namespace, pod, container, srcFile string) ReadFunc {
+	return func(offset uint64, writer io.Writer) error {
+		command := []string{"sh", "-c", fmt.Sprintf(defaultReadFromByteCmd, offset, srcFile)}
+		return client.execInPodWithWriters(ctx, ExecParameters{
+			Namespace: namespace,
+			Pod:       pod,
+			Container: container,
+			Command:   command,
+		}, writer, writer)
+	}
+}
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -18,7 +18,7 @@ import (
 	"github.com/cilium/cilium-cli/internal/utils"
 
 	"github.com/cilium/workerpool"
-	archiver "github.com/mholt/archiver/v3"
+	"github.com/mholt/archiver/v3"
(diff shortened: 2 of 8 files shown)


christarazi (approved):
:rocket:

sysdump/sysdump.go
```diff
@@ -899,21 +899,27 @@ func (c *Collector) submitBugtoolTasks(ctx context.Context, pods []*corev1.Pod,
 			if len(m) != 2 || len(m[1]) == 0 {
 				return fmt.Errorf("failed to collect 'cilium-bugtool' output for %q in namespace %q: output doesn't contain archive name: %s", p.Name, p.Namespace, outString)
 			}
-			// Grab the resulting archive's contents from the pod.
-			b, err := c.Client.ExecInPod(ctx, p.Namespace, p.Name, containerName, []string{"cat", m[1]})
+
+			// Gzip bugtool tar file to reduce the size of transferred file
+			o, e, err = c.Client.ExecInPodWithStderr(ctx, p.Namespace, p.Name, containerName, []string{"gzip", m[1]})
```
tklauser, line 904: Not blocking but might be worth considering as a follow-up: Could we also change `cilium-bugtool` to write a `.tar.gz` directly? This way we wouldn't have to run `gzip` as part of the sysdump for Cilium versions where the bugtool writes a compressed archive.

tklauser (approved):
LGTM with one non-blocking comment.

sysdump/sysdump.go
sayboras: I completely agree with the above, the reason that I am not doing it right now is mentioned here https://github.com/cilium/cilium-cli/issues/616#issuecomment-1000023489.

Will raise one small ticket to track this if it's not urgent.

tklauser: Thanks for the link, seems I missed that comment. Makes sense to raise a follow-up ticket for this 👍

internal/k8s/copy.go
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2020 Authors of Cilium
```
twpayne, line 2: nit: The copyright year should be 2021. This also applies to other files.

internal/k8s/copy.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *Client) CopyFromPod(ctx context.Context, namespace, pod, container string, srcFile, destFile string) error {
+	pipe := newPipe(&CopyOptions{
+		MaxTries: defaultMaxTries,
+		ReadFunc: readFromPod(ctx, c, namespace, pod, container, srcFile),
+	})
+
+	outFile, err := os.Create(destFile)
+	if err != nil {
+		return err
+	}
```
twpayne, line 28: This function is missing a `defer outFile.Close()` or similar.

twpayne (commented):
Apologies for the late review, but there are a couple of nits.

internal/k8s/copy.go
sayboras: my bad :(

internal/k8s/copy.go
sayboras: you are right, I forgot about this most of the time, and relying on linters. If I am not wrong, file is having finalizer i.e. fd will be garbage collected later, not idea in case of reuse of file descriptor.

Let me make a small update on this, thanks again.

PR #678, opened 2022-01-12
connectivity/check: use insecure.NewCredentials instead of WithInsecure

tklauser:
grpc.DialContext(..., grpc.WithInsecure()) is deprecated in favor of
grpc.DialContext(..., grpc.WithTransportCredentials(insecure.NewCredentials())
as of grpc v1.43.0, see [1]. Update to the later method.

[1] https://pkg.go.dev/google.golang.org/grpc@v1.43.0#WithInsecure

This will allow #663 to move forward.


rolinh (approved):
Thanks!

PR #699, opened 2022-02-01
connectivity test: allow node-local DNS cache

rastislavs:
Similarly to https://github.com/cilium/cilium/pull/18672, this PR allows running connectivity-check on clusters with [Nodelocal DNS Cache](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns/nodelocaldns) running in `hostNetwork=true` pods bound to a link-local IP (standard node-local DNS cache deployment, not using [Cilium LRP](https://docs.cilium.io/en/latest/gettingstarted/local-redirect-policy/#node-local-dns-cache)). For cases when cluster DNS communication should be allowed, it allows DNS requests to the link-local subnets recommended by the [node-local-dns configuration documentation](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration).

Due to https://github.com/cilium/cilium/issues/18644, it also allows DNS requests to the `host` entity.

Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>

### DIFF
--- a/connectivity/manifests/client-egress-only-dns.yaml
+++ b/connectivity/manifests/client-egress-only-dns.yaml
@@ -19,3 +19,22 @@ spec:
     - matchLabels:
         k8s:io.kubernetes.pod.namespace: kube-system
         k8s:k8s-app: kube-dns
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+      rules:
+        dns:
+        - matchPattern: "*"
+    toCIDR:
+    - 169.254.0.0/16
+    - fd00::/8
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+      rules:
+        dns:
+        - matchPattern: "*"
+    toEntities:
+    - host
--- a/connectivity/manifests/client-egress-to-echo.yaml
+++ b/connectivity/manifests/client-egress-to-echo.yaml
@@ -24,3 +24,16 @@ spec:
     - matchLabels:
         k8s:io.kubernetes.pod.namespace: kube-system
         k8s:k8s-app: kube-dns
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+    toCIDR:
+    - 169.254.0.0/16
+    - fd00::/8
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+    toEntities:
+    - host
--- a/connectivity/manifests/client-egress-to-entities-world.yaml
+++ b/connectivity/manifests/client-egress-to-entities-world.yaml
@@ -22,3 +22,16 @@ spec:
     - ports:
       - port: "53"
         protocol: ANY
+  - toCIDR:
+    - 169.254.0.0/16
+    - fd00::/8
+    toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+  - toEntities:
+    - host
+    toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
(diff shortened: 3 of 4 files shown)


twpayne (approved):
LGTM, thanks!

connectivity/manifests/client-egress-only-dns.yaml
```diff
@@ -22,3 +22,22 @@ spec:
...
+    - fd00::/8
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: ANY
+      rules:
+        dns:
+        - matchPattern: "*"
+    toEntities:
+    - host
```
aditighag, line 43: I'm hesitant to attribute the host identity to 169.254 CIDR since it doesn't necessarily mean local host only traffic. For example, the subnet is used for a remote metadata server in EKS deployments.

aditighag (changes requested):
I have reservations about the PR changes. In general, we don't want to special case the existing connectivity checks.

connectivity/manifests/client-egress-only-dns.yaml
joestringer: ^ Maybe relevant, I wonder if you are able to describe why both rules are necessary - it looks like this policy allows either to the 169.254.0.0/16 CIDR on port 53, or separately also to the host on port 53. Did you find that something didn't work if one of these rules is not present?

Ideally we'd have some way to detect whether node-local DNS was in use and decide to dynamically insert these policies based on that. In a sense, this is what we have with the existing kube-dns / coredns label matching logic, since pods would have to first be deployed with those labels for the policy to apply. But I don't know if there's an easy way to do that.

While it's maybe not ideal to poke holes for node-local DNS if it's not configured, I don't see a huge danger here as the user already has access to deploy pods and configure the networking policy for these pods, so they can control what the pods have access to; and furthermore, it's a bit hard to see how an adversary could leverage these policies if this is limited to the connectivity checks and also if they would have to deploy some rogue DNS server on the host or somehow hijack 169.254.x.y traffic on the node first before they can leverage this policy --- which would imply that they already have bigger privileges than what these configurations would allow.

rastislavs: > I wonder if you are able to describe why both rules are necessary - it looks like this policy allows either to the 169.254.0.0/16 CIDR on port 53, or separately also to the host on port 53. Did you find that something didn't work if one of these rules is not present?

@joestringer Yes, it is due to https://github.com/cilium/cilium/pull/18672. Depending on the order in which cilium pod & node-local-dns pod are started, either one or the other policy works. Only both policies work for all the cases.

rastislavs: > I'm hesitant to attribute the host identity to 169.254 CIDR since it doesn't necessarily mean local host only traffic. For example, the subnet is used for a remote metadata server in EKS deployments.

@aditighag I agree that whole link-local subnet does not necessarily mean localhost-only traffic. But I wasn't able to figure out anything else that wouldn't be too complex - selecting the pod by labels does not work in this case (not sure if that is a bug or not).

The only other option that I see would be to have the node-local-dns IP address configurable or auto-detected somehow. In CLI case this is doable, but not sure how to easily implement that in the manifest-based deployment (see https://github.com/cilium/cilium/pull/18672).

After all, these are test, aren't they? So I believe the discussion should be about how/when this change could affect the test results. I cannot think about such scenarios - and if there are some, I think they should be definitely less common than the standard node-local-dns deployment.

joestringer: My gut feeling here is that we should figure out the solution to #18644 first, then based on that solution we will have a better understanding of how these connectivity tests need to be written to account for the (hopefully now more reliable) implementation.

rastislavs: @joestringer OK, let me close this PR and wait for the resolution of #18644

rastislavs:
Closing as the ultimate solution will depend on the resolution of https://github.com/cilium/cilium/issues/18644

PR #708, opened 2022-02-10
Network performance test

jtaleric:
Allow the user to test network performance for [TCP,UDP]_STREAM,[TCP,UDP]_RR,TCP_CRR on a single node, as well as across nodes (if the deployment has > 1 node). 

`>  ./cilium connectivity test --test network-perf`

The user can override some of the default functionality,  
- The number of samples the test captures (default, 1 sample per test). 
- Test duration (default, 30 second per test) 

`> cilium connectivity test --perf --duration 5s --samples 3`

```
🔥 Performance Test Summary
-----------------------------------------------------------------------------------------------------------------------------
📋 Scenario                                           | Test            | Num Samples     | Duration        | Avg value      
-----------------------------------------------------------------------------------------------------------------------------
📋 perf-client-789bbd68f-vll7d                        | UDP_RR          | 3               | 5               | 37171.81 (OP/s)
📋 perf-client-other-node-6c4d4c69c6-tjbpv            | TCP_RR          | 3               | 5               | 6607.50 (OP/s)
📋 perf-client-other-node-6c4d4c69c6-tjbpv            | TCP_CRR         | 3               | 5               | 1082.05 (OP/s)
📋 perf-client-other-node-6c4d4c69c6-tjbpv            | TCP_STREAM      | 3               | 5               | 1740.09 (Mb/s)
📋 perf-client-other-node-6c4d4c69c6-tjbpv            | UDP_RR          | 3               | 5               | 6586.60 (OP/s)
📋 perf-client-other-node-6c4d4c69c6-tjbpv            | UDP_STREAM      | 3               | 5               | 933.03 (Mb/s)
📋 perf-client-789bbd68f-vll7d                        | TCP_RR          | 3               | 5               | 33898.99 (OP/s)
📋 perf-client-789bbd68f-vll7d                        | TCP_CRR         | 3               | 5               | 4538.74 (OP/s)
📋 perf-client-789bbd68f-vll7d                        | TCP_STREAM      | 3               | 5               | 1005.58 (Mb/s)
📋 perf-client-789bbd68f-vll7d                        | UDP_STREAM      | 3               | 5               | 900.42 (Mb/s)
-----------------------------------------------------------------------------------------------------------------------------

```

### DIFF
--- a/.github/in-cluster-test-scripts/aks.sh
+++ b/.github/in-cluster-test-scripts/aks.sh
@@ -14,5 +14,8 @@ sleep 10s
 # Run connectivity test
 cilium connectivity test --debug --all-flows
 
+# Run performance test
+cilium connectivity test --perf --perf-duration 1s
+
 # Retrieve Cilium status
 cilium status
--- a/connectivity/check/check.go
+++ b/connectivity/check/check.go
@@ -35,6 +35,9 @@ type Parameters struct {
 	Debug                 bool
 	PauseOnFail           bool
 	SkipIPCacheCheck      bool
+	Perf                  bool
+	PerfDuration          time.Duration
+	PerfSamples           int
 	CiliumBaseVersion     string
 }
 
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -39,6 +39,9 @@ type ConnectivityTest struct {
 	ciliumPods        map[string]Pod
 	echoPods          map[string]Pod
 	clientPods        map[string]Pod
+	perfClientPods    map[string]Pod
+	perfServerPod     map[string]Pod
+	PerfResults       map[PerfTests]PerfResult
 	echoServices      map[string]Service
 	externalWorkloads map[string]ExternalWorkload
 
@@ -46,6 +49,19 @@ type ConnectivityTest struct {
 	testNames map[string]struct{}
 
 	lastFlowTimestamps map[string]time.Time
+}
+
+type PerfTests struct {
+	Pod  string
+	Test string
+}
+
+type PerfResult struct {
+	Metric   string
+	Duration time.Duration
+	Samples  int
+	Values   []float64
+	Avg      float64
 }
 
 // verbose returns the value of the user-provided verbosity flag.
@@ -135,6 +151,9 @@ func NewConnectivityTest(client *k8s.Client, p Parameters) (*ConnectivityTest, e
 		ciliumPods:         make(map[string]Pod),
 		echoPods:           make(map[string]Pod),
 		clientPods:         make(map[string]Pod),
+		perfClientPods:     make(map[string]Pod),
+		perfServerPod:      make(map[string]Pod),
+		PerfResults:        make(map[PerfTests]PerfResult),
 		echoServices:       make(map[string]Service),
 		externalWorkloads:  make(map[string]ExternalWorkload),
 		tests:              []*Test{},
@@ -294,6 +313,19 @@ func (ct *ConnectivityTest) report() error {
 		}
 
 		return fmt.Errorf("%d tests failed", nf)
+	}
+
+	if ct.params.Perf {
+		// Report Performance results
+		ct.Header("🔥 Performance Test Summary")
+		ct.Logf("%s", strings.Repeat("-", 125))
+		ct.Logf("📋 %-50s | %-15s | %-15s | %-15s | %-15s", "Scenario", "Test", "Num Samples", "Duration", "Avg value")
+		ct.Logf("%s", strings.Repeat("-", 125))
+		for p, d := range ct.PerfResults {
+			ct.Logf("📋 %-50s | %-15s | %-15d | %-15s | %.2f (%s)", p.Pod, p.Test, d.Samples, d.Duration, d.Avg, d.Metric)
+			ct.Debugf("Individual Values from run : %s", d.Values)
+		}
+		ct.Logf("%s", strings.Repeat("-", 125))
 	}
 
 	ct.Headerf("✅ All %d tests (%d actions) successful, %d tests skipped, %d scenarios skipped.", nt-nst, na, nst, nss)
@@ -458,12 +490,24 @@ func (ct *ConnectivityTest) RandomClientPod() *Pod {
 	return nil
 }
 
+func (ct *ConnectivityTest) Params() Parameters {
+	return ct.params
+}
+
 func (ct *ConnectivityTest) CiliumPods() map[string]Pod {
 	return ct.ciliumPods
 }
 
 func (ct *ConnectivityTest) ClientPods() map[string]Pod {
 	return ct.clientPods
+}
+
+func (ct *ConnectivityTest) PerfServerPod() map[string]Pod {
+	return ct.perfServerPod
+}
+
+func (ct *ConnectivityTest) PerfClientPods() map[string]Pod {
+	return ct.perfClientPods
 }
 
 func (ct *ConnectivityTest) EchoPods() map[string]Pod {
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -23,6 +23,13 @@ import (
 )
 
 const (
+	PerfClientName = "PerfClient"
+	PerfServerName = "PerfServer"
+
+	PerfClientDeploymentName       = "perf-client"
+	PerfClientAcrossDeploymentName = "perf-client-other-node"
+	PerfServerDeploymentName       = "perf-server"
+
 	ClientDeploymentName  = "client"
 	Client2DeploymentName = "client2"
 
@@ -30,6 +37,7 @@ const (
 	echoOtherNodeDeploymentName = "echo-other-node"
 	kindEchoName                = "echo"
 	kindClientName              = "client"
+	kindPerfName                = "perf"
 )
 
 type deploymentParameters struct {
@@ -235,6 +243,152 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 		}
 	}
 
+	if ct.params.Perf {
+		// For performance workloads, we want to ensure the client/server are in the same zone
+		// If a zone has > 1 node, use that zone
+		zones := map[string]int{}
+		zone := ""
+		lz := ""
+		n, hasNodes := ct.client.ListNodes(ctx, metav1.ListOptions{})
+		if hasNodes != nil {
+			return fmt.Errorf("unable to query nodes")
+		}
+		for _, l := range n.Items {
+			if _, ok := zones[l.GetLabels()["topology.kubernetes.io/zone"]]; ok {
+				zone = l.GetLabels()["topology.kubernetes.io/zone"]
+				break
+			} else {
+				zones[l.GetLabels()["topology.kubernetes.io/zone"]] = 1
+				lz = l.GetLabels()["topology.kubernetes.io/zone"]
+			}
+		}
+		// No zone had > 1, use the last zone.
+		if zone == "" {
+			ct.Warn("Each zone only has a single node - could impact the performance test results")
+			zone = lz
+		}
+
+		// Need to capture the IP of the Server Deployment, and pass to the client to execute benchmark
+		_, err = ct.clients.src.GetDeployment(ctx, ct.params.TestNamespace, PerfClientDeploymentName, metav1.GetOptions{})
+		if err != nil {
+			ct.Logf("✨ [%s] Deploying Perf Client deployment...", ct.clients.src.ClusterName())
+			perfClientDeployment := newDeployment(deploymentParameters{
+				Name:  PerfClientDeploymentName,
+				Kind:  kindPerfName,
+				Port:  80,
+				Image: defaults.ConnectivityPerformanceImage,
+				Labels: map[string]string{
+					"client": "role",
+				},
+				Command: []string{"/bin/bash", "-c", "sleep 10000000"},
+				Affinity: &corev1.Affinity{
+					NodeAffinity: &corev1.NodeAffinity{
+						PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
+							{
+								Weight: 100,
+								Preference: corev1.NodeSelectorTerm{
+									MatchExpressions: []corev1.NodeSelectorRequirement{
+										{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpIn, Values: []string{zone}},
+									},
+								},
+							},
+						},
+					},
+				},
+			})
+			_, err = ct.clients.src.CreateDeployment(ctx, ct.params.TestNamespace, perfClientDeployment, metav1.CreateOptions{})
+			if err != nil {
+				return fmt.Errorf("unable to create deployment %s: %w", perfClientDeployment, err)
+			}
+		}
+
+		_, err = ct.clients.src.GetDeployment(ctx, ct.params.TestNamespace, PerfServerDeploymentName, metav1.GetOptions{})
+		if err != nil {
+			ct.Logf("✨ [%s] Deploying Perf Server deployment...", ct.clients.src.ClusterName())
+			perfServerDeployment := newDeployment(deploymentParameters{
+				Name: PerfServerDeploymentName,
+				Kind: kindPerfName,
+				Labels: map[string]string{
+					"server": "role",
+				},
+				Port:    5001,
+				Image:   defaults.ConnectivityPerformanceImage,
+				Command: []string{"/bin/bash", "-c", "netserver;sleep 10000000"},
+				Affinity: &corev1.Affinity{
+					NodeAffinity: &corev1.NodeAffinity{
+						PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
+							{
+								Weight: 100,
+								Preference: corev1.NodeSelectorTerm{
+									MatchExpressions: []corev1.NodeSelectorRequirement{
+										{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpIn, Values: []string{zone}},
+									},
+								},
+							},
+						},
+					},
+					PodAffinity: &corev1.PodAffinity{
+						RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+							{
+								LabelSelector: &metav1.LabelSelector{
+									MatchExpressions: []metav1.LabelSelectorRequirement{
+										{Key: "name", Operator: metav1.LabelSelectorOpIn, Values: []string{PerfClientDeploymentName}},
+									},
+								},
+								TopologyKey: "kubernetes.io/hostname",
+							},
+						},
+					},
+				},
+			})
+			_, err = ct.clients.src.CreateDeployment(ctx, ct.params.TestNamespace, perfServerDeployment, metav1.CreateOptions{})
+			if err != nil {
+				return fmt.Errorf("unable to create deployment %s: %w", perfServerDeployment, err)
+			}
+		}
+
+		// Deploy second client on a different node
+		if !ct.params.SingleNode {
+			_, err := ct.clients.src.GetDeployment(ctx, ct.params.TestNamespace, PerfClientAcrossDeploymentName, metav1.GetOptions{})
+			if err != nil {
+				ct.Logf("✨ [%s] Deploying Perf Client deployment...", ct.clients.src.ClusterName())
+				perfClientDeployment := newDeployment(deploymentParameters{
+					Name: PerfClientAcrossDeploymentName,
+					Kind: kindPerfName,
+					Port: 5001,
+					Labels: map[string]string{
+						"client": "role",
+					},
+					Image:   defaults.ConnectivityPerformanceImage,
+					Command: []string{"/bin/bash", "-c", "sleep 10000000"},
+					Affinity: &corev1.Affinity{
+						NodeAffinity: &corev1.NodeAffinity{
+							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.PreferredSchedulingTerm{
+								{
+									Weight: 100,
+									Preference: corev1.NodeSelectorTerm{
+										MatchExpressions: []corev1.NodeSelectorRequirement{
+											{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpIn, Values: []string{zone}},
+										},
+									},
+								},
+							},
+						},
+						PodAntiAffinity: &corev1.PodAntiAffinity{
+							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+								{Weight: 100, PodAffinityTerm: corev1.PodAffinityTerm{
+									LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
+										{Key: "name", Operator: metav1.LabelSelectorOpIn, Values: []string{PerfClientDeploymentName}}}},
+									TopologyKey: "kubernetes.io/hostname"}}}},
+					},
+				})
+				_, err = ct.clients.src.CreateDeployment(ctx, ct.params.TestNamespace, perfClientDeployment, metav1.CreateOptions{})
+				if err != nil {
+					return fmt.Errorf("unable to create deployment %s: %s", perfClientDeployment, err)
+				}
+			}
+		}
+	}
 	_, err = ct.clients.src.GetDeployment(ctx, ct.params.TestNamespace, ClientDeploymentName, metav1.GetOptions{})
 	if err != nil {
 		ct.Logf("✨ [%s] Deploying client deployment...", ct.clients.src.ClusterName())
@@ -327,7 +481,7 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 
 			_, err = ct.clients.dst.CreateDeployment(ctx, ct.params.TestNamespace, echoOtherNodeDeployment, metav1.CreateOptions{})
 			if err != nil {
-				return fmt.Errorf("unable to create deployment %s: %s", echoOtherNodeDeploymentName, err)
+				return fmt.Errorf("unable to create deployment %s: %w", echoOtherNodeDeploymentName, err)
 			}
 		}
 	}
@@ -337,9 +491,14 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 
 // deploymentList returns 2 lists of Deployments to be used for running tests with.
 func (ct *ConnectivityTest) deploymentList() (srcList []string, dstList []string) {
-	srcList = []string{ClientDeploymentName, Client2DeploymentName, echoSameNodeDeploymentName}
+	if !ct.params.Perf {
+		srcList = []string{ClientDeploymentName, Client2DeploymentName, echoSameNodeDeploymentName}
+	} else {
+		srcList = []string{PerfClientDeploymentName}
+		dstList = append(dstList, PerfServerDeploymentName)
+	}
 
-	if ct.params.MultiCluster != "" || !ct.params.SingleNode {
+	if (ct.params.MultiCluster != "" || !ct.params.SingleNode) && !ct.params.Perf {
 		dstList = append(dstList, echoOtherNodeDeploymentName)
 	}
 
@@ -384,7 +543,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	for _, client := range ct.clients.clients() {
 		ciliumPods, err := client.ListPods(ctx, ct.params.CiliumNamespace, metav1.ListOptions{LabelSelector: "k8s-app=cilium"})
 		if err != nil {
-			return fmt.Errorf("unable to list Cilium pods: %s", err)
+			return fmt.Errorf("unable to list Cilium pods: %w", err)
 		}
 		for _, ciliumPod := range ciliumPods.Items {
 			// TODO: Can Cilium pod names collide across clusters?
@@ -413,10 +572,35 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
+	perfPods, err := ct.client.ListPods(ctx, ct.params.TestNamespace, metav1.ListOptions{LabelSelector: "kind=" + kindPerfName})
+	if err != nil {
+		return fmt.Errorf("unable to list perf pods: %w", err)
+	}
+	for _, perfPod := range perfPods.Items {
+		ctx, cancel := context.WithTimeout(ctx, ct.params.ciliumEndpointTimeout())
+		defer cancel()
+		if err := ct.waitForCiliumEndpoint(ctx, ct.clients.src, ct.params.TestNamespace, perfPod.Name); err != nil {
+			return err
+		}
+		_, hasLabel := perfPod.GetLabels()["server"]
+		if hasLabel {
+			ct.perfServerPod[perfPod.Name] = Pod{
+				K8sClient: ct.client,
+				Pod:       perfPod.DeepCopy(),
+				port:      5201,
+			}
+		} else {
+			ct.perfClientPods[perfPod.Name] = Pod{
+				K8sClient: ct.client,
+				Pod:       perfPod.DeepCopy(),
+			}
+		}
+	}
+
 	for _, client := range ct.clients.clients() {
 		echoPods, err := client.ListPods(ctx, ct.params.TestNamespace, metav1.ListOptions{LabelSelector: "kind=" + kindEchoName})
 		if err != nil {
-			return fmt.Errorf("unable to list echo pods: %s", err)
+			return fmt.Errorf("unable to list echo pods: %w", err)
 		}
 		for _, echoPod := range echoPods.Items {
 			ctx, cancel := context.WithTimeout(ctx, ct.params.ciliumEndpointTimeout())
@@ -437,7 +621,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	for _, client := range ct.clients.clients() {
 		echoServices, err := client.ListServices(ctx, ct.params.TestNamespace, metav1.ListOptions{LabelSelector: "kind=" + kindEchoName})
 		if err != nil {
-			return fmt.Errorf("unable to list echo services: %s", err)
+			return fmt.Errorf("unable to list echo services: %w", err)
 		}
 
 		for _, echoService := range echoServices.Items {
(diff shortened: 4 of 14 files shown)


connectivity/check/deployment.go
```diff
@@ -235,6 +243,141 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+		} else {
+			zones[l.GetLabels()["topology.kubernetes.io/zone"]] = 1
+			lz = l.GetLabels()["topology.kubernetes.io/zone"]
+		}
+	}
+	// No zone had > 1, use the last zone.
+	if zone == "" {
+		ct.Debug("Each zone only has a single node - could impact the performance test results")
+		zone = lz
+	}
```
jtaleric, line 268: Seems like this should be a conditional. If the zone `label` exists. 

Looking at the Kind output, it seems it doesn't exist, so I should add some logic to have the `NodeAffinity` based on it existing.

connectivity/check/deployment.go
```diff
@@ -235,6 +243,141 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+			Name:  PerfClientDeploymentName,
+			Kind:  kindPerfName,
+			Port:  80,
+			Image: defaults.ConnectivityPerformanceImage,
+			Labels: map[string]string{
+				"client": "role",
+			},
+			Command: []string{"/bin/bash", "-c", "sleep 10000000"},
+			Affinity: &corev1.Affinity{
+				NodeAffinity: &corev1.NodeAffinity{
```
jtaleric, line 284: Needs to be a conditional, since the label for zone won't always exist.

connectivity/check/deployment.go
```diff
@@ -235,6 +243,141 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+					RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
```
jtaleric, line 285: To circumvent the above(changing the current logic to determine zone when it exists), I could change this to Preferred.

connectivity/check/context.go
```diff
@@ -135,7 +138,10 @@ func NewConnectivityTest(client *k8s.Client, p Parameters) (*ConnectivityTest, e
 		ciliumPods:         make(map[string]Pod),
 		echoPods:           make(map[string]Pod),
 		clientPods:         make(map[string]Pod),
+		perfClientPods:     make(map[string]Pod),
+		perfServerPod:      make(map[string]Pod),
 		echoServices:       make(map[string]Service),
+		PerfResults:        make(map[string]map[string]string),
```
tklauser, line 144: Nit: let's initialize these in the same order they're defined in `type ConnectivityTest`:

```suggestion
		perfClientPods:     make(map[string]Pod),
		perfServerPod:      make(map[string]Pod),
		PerfResults:        make(map[string]map[string]string),
		echoServices:       make(map[string]Service),
```

connectivity/check/context.go
```diff
@@ -289,6 +295,17 @@ func (ct *ConnectivityTest) report() error {
 
 		return fmt.Errorf("%d tests failed", nf)
 	}
+	// Report Performance results
+	if len(ct.PerfResults) > 0 {
```
tklauser, line 299: Nit: this check shouldn't be needed, the `for` loop below will not run for an empty `ct.PerfResults`.

connectivity/check/deployment.go
```diff
@@ -235,6 +243,141 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+			zone = l.GetLabels()["topology.kubernetes.io/zone"]
+			break
+		} else {
+			zones[l.GetLabels()["topology.kubernetes.io/zone"]] = 1
+			lz = l.GetLabels()["topology.kubernetes.io/zone"]
+		}
+	}
+	// No zone had > 1, use the last zone.
+	if zone == "" {
+		ct.Debug("Each zone only has a single node - could impact the performance test results")
```
tklauser, line 266: Should we log this as a warning instead given that this might impact the results?

connectivity/check/deployment.go
```diff
@@ -413,6 +556,31 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
...
+		return fmt.Errorf("unable to list perf pods: %s", err)
+	}
+	for _, perfPod := range perfPods.Items {
+		ctx, cancel := context.WithTimeout(ctx, ct.params.ciliumEndpointTimeout())
+		defer cancel()
+		if err := ct.waitForCiliumEndpoint(ctx, ct.clients.src, ct.params.TestNamespace, perfPod.Name); err != nil {
+			return err
+		}
+		_, err := perfPod.GetLabels()["server"]
+		if err {
```
tklauser, line 570: Nit: In Go `err` is usually a var of type `error`. How about naming this `hasServer` or similar?

```suggestion
		_, hasServer := perfPod.GetLabels()["server"]
		if hasServer {
```

connectivity/tests/perfpod.go
```diff
@@ -0,0 +1,120 @@
+package tests
```
tklauser, line 1: Could you please add a copyright header to the newly created file:

```
// SPDX-License-Identifier: Apache-2.0
// Copyright 2022 Authors of Cilium
```

connectivity/tests/perfpod.go
```diff
@@ -0,0 +1,120 @@
...
+		duration = defaults.PerfTestDuration
+	}
+	for i := 0; i < samples; i++ {
+		// Result data
+		r := make(map[string]string)
+
+		exec := []string{"/usr/bin/iperf3", "-c", sip, "-t", fmt.Sprintf("%d", duration), "-J"}
+		if udp {
+			exec = []string{"/usr/bin/iperf3", "-u", "-c", sip, "-t", fmt.Sprintf("%d", duration), "-J"}
+		}
```
tklauser, line 97: This can be moved out of the `for` loop, I think. It only depends on arguments passed to the function.

defaults/defaults.go
```diff
@@ -66,12 +66,15 @@ const (
 	ConnectivityCheckNamespace = "cilium-test"
 
 	ConnectivityCheckAlpineCurlImage = "quay.io/cilium/alpine-curl:v1.4.0@sha256:2550c747831ff575f2147149b088ea981c06f9b6bcd188756d1b82cc10997956"
+	ConnectivityPerformanceImage     = "quay.io/jtaleric/iperf3:latest"
```
tklauser, line 69: Would be nice to use a tagged release with image digest for this image as well. That way we can make image version updates in `cilium-cli` explicit via a PR and would catch potential issues in CI.

tklauser (commented):
Very nice! 🚀

Few comments/questions inline, mostly nits.

connectivity/check/context.go
jtaleric: habit... damnit python brain.

jtaleric:
@tklauser Let me know if there is anything else I should address. My only other thought is documentation explaining the workload? Where do we maintain that?

defaults/defaults.go
```diff
@@ -66,7 +66,7 @@ const (
 	ConnectivityCheckNamespace = "cilium-test"
 
 	ConnectivityCheckAlpineCurlImage = "quay.io/cilium/alpine-curl:v1.4.0@sha256:2550c747831ff575f2147149b088ea981c06f9b6bcd188756d1b82cc10997956"
-	ConnectivityPerformanceImage     = "quay.io/cilium/iperf3:ef49999ed25ef02ea378f77e050a53caebfcaddd@sha256:f2d58d560593e3212a004158c7139b79198ad87d9f53e66394a0f21ebdd67325"
+	ConnectivityPerformanceImage     = "quay.io/jtaleric/network-perf:latest@sha256:db30603e9ff45e24c44fdfb4813c17e3431fc0e61d0eaac52a629ba59cd8e555"
```
jtaleric, line 69: @tklauser I can move this to the cilium repo once we are good with the PR.

jtaleric: Also, this image has both iperf and netperf, just in case 😉

jtaleric:
AKS failures appear to be unrelated to the Perf tests...

tklauser (commented):
Given that running only the performance tests with default values (number of samples and duration) takes ~5min - essentially duplicating the current run time of connectivity tests - I wonder whether we should make these tests opt-in or decrease these default values?

Of course we'd still want to enable these in our CI. But users running `cilium connectivity test` e.g. as part of the [Getting Started Guides](https://docs.cilium.io/en/v1.11/gettingstarted/k8s-install-default/#validate-the-installation) by default wouldn't run performance tests and they would be able to get results on correct Cilium connectivity quicker.

What do other reviewers think? @ldelossa @ti-mo @pchaigno

Some small nits inline.

jtaleric:
> Given that running only the performance tests with default values (number of samples and duration) takes ~5min - essentially duplicating the current run time of connectivity tests - I wonder whether we should make these tests opt-in or decrease these default values?

Just to make sure I understand, unless the user explicitly passes`--test network-perf` these tests will not run? 

> 
> Of course we'd still want to enable these in our CI. But users running `cilium connectivity test` e.g. as part of the [Getting Started Guides](https://docs.cilium.io/en/v1.11/gettingstarted/k8s-install-default/#validate-the-installation) by default wouldn't run performance tests and they would be able to get results on correct Cilium connectivity quicker.

👍🏽 

> 
> What do other reviewers think? @ldelossa @ti-mo @pchaigno
> 
> Some small nits inline.

Thanks @tklauser !

tklauser:
> > Given that running only the performance tests with default values (number of samples and duration) takes ~5min - essentially duplicating the current run time of connectivity tests - I wonder whether we should make these tests opt-in or decrease these default values?
> 
> Just to make sure I understand, unless the user explicitly passes`--test network-perf` these tests will not run?

Yeah, this is what I had in mind. But thinking about it now, it might be slightly counter-intuitive to have a single test as opt-in while all other test are opt-out (i.e. they run, unless `--test !testname` is specified).

Not sure if that matters much, or whether we should split performance tests into their own sub-command (e.g. `cilium connectivity perf`) or add a specific flag (e.g. `cilium connectivity test --perf`) whereby each would only run the connectivity tests. WDYT?

> > What do other reviewers think? @ldelossa @ti-mo @pchaigno
> > Some small nits inline.
> 
> Thanks @tklauser !

Thank you for adding these tests!

jtaleric:
> Not sure if that matters much, or whether we should split performance tests into their own sub-command (e.g. `cilium connectivity perf`) or add a specific flag (e.g. `cilium connectivity test --perf`) whereby each would only run the connectivity tests. WDYT?

@tklauser Agreed. To avoid changing the current functionality, we should make this a parameter the user should pass. 

The latest commit has the `--perf` parameter. 

The only change I would need to add is to the tests, to include `--perf`. However, I want to get feedback on the current approach.

connectivity/tests/perfpod.go
```diff
@@ -0,0 +1,115 @@
...
+func netperf(ctx context.Context, sip string, podname string, test string, a *check.Action, result map[check.PerfTests]check.PerfResult, samples int) {
+	// Define test about to be executed and from which pod
+	k := check.PerfTests{
+		Pod:  podname,
+		Test: test,
+	}
+
+	// Allow the user to override aspects of the test
+	iteration, _ := strconv.Atoi(os.Getenv("samples"))
+	duration, _ := strconv.Atoi(os.Getenv("duration"))
```
christarazi, line 69: Where are these env vars coming from? Would be good to document that.

connectivity/check/deployment.go
```diff
@@ -235,6 +243,150 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+								},
+							},
+						},
+					},
+				},
+			},
+		})
+		_, err = ct.clients.src.CreateDeployment(ctx, ct.params.TestNamespace, perfClientDeployment, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("unable to create deployment %s: %s", perfClientDeployment, err)
```
christarazi: Errors should be formatted with `%w`

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

connectivity/check/deployment.go
```diff
@@ -235,6 +243,150 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+								MatchExpressions: []corev1.NodeSelectorRequirement{
+									{Key: "topology.kubernetes.io/zone", Operator: corev1.NodeSelectorOpIn, Values: []string{zone}},
+								},
+							},
+						},
+					},
+				},
+			},
+		})
+		_, err = ct.clients.src.CreateDeployment(ctx, ct.params.TestNamespace, perfClientDeployment, metav1.CreateOptions{})
```
christarazi: There's a missing error handling case here, but not sure it makes too much of a difference. The current code is assuming that the way to handle an error from fetching the deployment is to create it, but what if it already exists and it errored during the fetch?

Typically this is handled by doing

```go
if err != nil && k8serrors.IsNotFound(err) {
  // create logic
} else if err != nil {
  // error during fetching
}

https://github.com/kubernetes/apimachinery/blob/2936d3f03931b9c06a641799605d8e806cb2a58b/pkg/api/errors/errors.go#L527

File: connectivity/check/deployment.go [hunk 15]
```diff
@@ -413,6 +565,31 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
...
+		defer cancel()
+		if err := ct.waitForCiliumEndpoint(ctx, ct.clients.src, ct.params.TestNamespace, perfPod.Name); err != nil {
+			return err
+		}
+		_, hasLabel := perfPod.GetLabels()["server"]
+		if hasLabel {
+			ct.perfServerPod[perfPod.Name] = Pod{
+				K8sClient: ct.client,
+				Pod:       perfPod.DeepCopy(),
+				port:      5201,
```
christarazi, line 590: From a high-level glance it's not clear to me why in this branch we put the port number, but in the else we omit it.

connectivity/tests/perfpod.go
```diff
@@ -0,0 +1,115 @@
...
+		if len(d) < 5 {
+			a.Fatal("Unable to process netperf result")
+		}
+		nv := ""
+		if len(d[len(d)-1]) > 0 {
+			nv = d[len(d)-1]
+		} else {
+			nv = d[len(d)-2]
+		}
+		f, _ := strconv.ParseFloat(nv, 64)
```
christarazi, line 97: Generally we should avoid hiding ignoring because it can result in really weird behavior without signal to the user as to what went wrong. Is this something that's totally unexpected? I don't have enough context about netperf, etc to tell whether there's a need to be defensive or not.

.github/workflows/aks.yaml
```diff
@@ -29,6 +29,7 @@ env:
   name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
   location: westeurope
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
+  duration: 5
```
christarazi, line 32: Nit: the commit title should say `Update GH workflow duration to 5s` and then provide the justification in the body. When I read the gitlog, this commit will show as `Updating GH workflows`, which is not useful to me as a developer digging through git history, which we do often when bisecting failures, etc, so it would force me to look at the diff to make sure I understand what `Updating` meant.

christarazi (changes requested):
Awesome! Very excited for this.

I have a misc. comments below around various things, but nothing too serious. I noticed that the 3rd commit and the first commit are essentially the same thing, so we might as well squash those.

.github/workflows/aks.yaml
jtaleric: ack - this actually will eventually be moved as it was there to reduce the test duration for the performance test. The latest commit changes the behavior a bit -- perf test is only ran if the user explicitly passes `--perf`.

connectivity/tests/perfpod.go
jtaleric: This could leave to unwanted behavior. If the regex doesn't pick up the right value -- ie something change in netperf (highly unlikely).

connectivity/check/deployment.go
jtaleric: Setting the port number is purely informational.

tklauser:
> > Not sure if that matters much, or whether we should split performance tests into their own sub-command (e.g. `cilium connectivity perf`) or add a specific flag (e.g. `cilium connectivity test --perf`) whereby each would only run the connectivity tests. WDYT?
> 
> @tklauser Agreed. To avoid changing the current functionality, we should make this a parameter the user should pass.

:+1: 

> The latest commit has the `--perf` parameter.
> 
> The only change I would need to add is to the tests, to include `--perf`. However, I want to get feedback on the current approach.

Thanks @rook, the approach looks good to me. Looks like the additional commit introducing the flag has some unrelated changes as well (error string formatting and such). I presume you're going to squash this with the first commit once all review comments are addressed?

.github/workflows/aks.yaml
jtaleric: I since we are now modifying how we execute the test with `--perf` I will make these cmd line options vs envvars.

connectivity/tests/perfpod.go
jtaleric: Removing these... In previous commits I was using them, but with the latest change to have the performance tests exist on their own, I will be removing these 👍🏽

connectivity/check/deployment.go
jtaleric: Let's discuss this more. I am not sure I am tracking the comment there...

connectivity/check/deployment.go
```diff
@@ -413,6 +572,31 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
+	perfPods, err := ct.client.ListPods(ctx, ct.params.TestNamespace, metav1.ListOptions{LabelSelector: "kind=" + kindPerfName})
+	if err != nil {
+		return fmt.Errorf("unable to list perf pods: %s", err)
```
christarazi, line 577: A few more instances of `%s` that need to be `%w`

christarazi (approved):
LGTM, pending leftover nits for `%w`

connectivity/check/deployment.go
jtaleric: Found a couple other places in the code... amending those changes... Thanks again @christarazi

jtaleric:
depends on : https://github.com/cilium/image-tools/pull/170

connectivity/check/deployment.go
jtaleric: From our discussion, https://github.com/cilium/cilium-cli/issues/734

.github/workflows/kind.yaml
```diff
@@ -15,6 +15,7 @@ env:
   KIND_CONFIG: .github/kind-config.yaml
   TIMEOUT: 2m
   LOG_TIME: 30m
+  duration: 5
```
tklauser, line 18: Do we still need this change now that the duration is passed as a CLI flag to `cilium connectivity test --perf`?

.github/workflows/multicluster.yaml
```diff
@@ -30,6 +30,7 @@ env:
   clusterName2: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}-mesh-2
   zone: us-west2-a
   firewallRuleName: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}-rule
+  duration: 5
```
tklauser, line 33: Likewise, is this still needed?

.github/in-cluster-test-scripts/aks.sh
```diff
@@ -14,5 +14,8 @@ sleep 10s
 # Run connectivity test
 cilium connectivity test --debug --all-flows
 
+# Run performance test
+cilium connectivity test --perf --duration 1
```
tklauser, line 18: As I understand the current logic, this would also run the connectivity tests so we would effectively run them twice. I'd either fold in the `--perf` into the command above or change `--perf` behavior to only run the performance tests. Likewise for the other in-cluster test scripts.

internal/cli/cmd/connectivity.go
```diff
@@ -120,6 +121,9 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.Debug, "debug", "d", false, "Show debug messages")
 	cmd.Flags().BoolVarP(&params.PauseOnFail, "pause-on-fail", "p", false, "Pause execution on test failure")
 	cmd.Flags().BoolVar(&params.SkipIPCacheCheck, "skip-ip-cache-check", true, "Skip IPCache check")
+	cmd.Flags().BoolVar(&params.Perf, "perf", false, "Run network Performance tests")
+	cmd.Flags().IntVar(&params.Duration, "duration", 30, "Duration for the Performance test to run, Default:30s")
```
tklauser, line 125: I'd suggest to rename this to `--perf-duration` so it's apparent from the name that this relates to performance tests:

```suggestion
	cmd.Flags().IntVar(&params.Duration, "perf-duration", 30, "Duration for the Performance test to run, Default:30s")
```

internal/cli/cmd/connectivity.go
```diff
@@ -120,6 +121,9 @@ func newCmdConnectivityTest() *cobra.Command {
...
+	cmd.Flags().IntVar(&params.Samples, "samples", 1, "Number of Performance samples to capture (how many times to run each test), Default:1")
```
tklauser, line 126: Same here:
```suggestion
	cmd.Flags().IntVar(&params.Samples, "perf-samples", 1, "Number of Performance samples to capture (how many times to run each test), Default:1")
```

tklauser (changes requested):
I think the tests on AKS, EKS and GKE didn't run with the workflow changes yet. To test GitHub workflow changes in the scope of a PR is unfortunately a bit cumbersome, see the comment at the top of the workflow files for details, e.g. https://github.com/cilium/cilium-cli/blob/57777b80f355da9db0a9ac7d14db9bacc1d0a491/.github/workflows/gke.yaml#L5-L18

Some more comments inline.

.github/in-cluster-test-scripts/aks.sh
jtaleric: actually it will just run the perf test. Instead of double dipping.

.github/workflows/kind.yaml
jtaleric: right. left overs.

connectivity/suite.go
```diff
@@ -62,6 +62,14 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 
 	ct.Infof("Cilium version: %v", v)
 
+	// Network Performance Test
+	if (ct.Params()).Perf {
```
tklauser, line 66: Nit:

```suggestion
	if ct.Params().Perf {
```

.github/in-cluster-test-scripts/aks.sh
tklauser: Ah, right. Sorry for the noise.

internal/cli/cmd/connectivity.go
```diff
@@ -120,6 +121,9 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.Debug, "debug", "d", false, "Show debug messages")
 	cmd.Flags().BoolVarP(&params.PauseOnFail, "pause-on-fail", "p", false, "Pause execution on test failure")
 	cmd.Flags().BoolVar(&params.SkipIPCacheCheck, "skip-ip-cache-check", true, "Skip IPCache check")
+	cmd.Flags().BoolVar(&params.Perf, "perf", false, "Run network Performance tests")
+	cmd.Flags().IntVar(&params.Duration, "perf-duration", 30, "Duration for the Performance test to run, Default:30s")
```
tklauser, line 125:
```suggestion
	cmd.Flags().DurationVar(&params.PerfDuration, "perf-duration", 30*time.Second, "Duration for the Performance test to run")
```

Let's change this parameter to type `time.Duration` which has the advantage that e.g. `connectivity test --perf --perf-duration 30s` works. This is also in line with other parametes taking a duration, e.g. `--post-test-sleep`).

Also there is no need to duplicate the default value in the usage message, the auto-generated help will already include it:

```
cilium connectivity test -h
[...]
      --perf-duration int          Duration for the Performance test to run, Default:30s (default 30)
[...]
```

After applying the changes above this will show:

```
      --perf-duration duration     Duration for the Performance test to run (default 30s)
```

internal/cli/cmd/connectivity.go
```diff
@@ -120,6 +121,9 @@ func newCmdConnectivityTest() *cobra.Command {
...
+	cmd.Flags().IntVar(&params.Samples, "perf-samples", 1, "Number of Performance samples to capture (how many times to run each test), Default:1")
```
tklauser, line 126: Let's drop the default value from the usage here as well:

```suggestion
	cmd.Flags().IntVar(&params.Samples, "perf-samples", 1, "Number of Performance samples to capture (how many times to run each test)")
```

defaults/defaults.go
```diff
@@ -66,12 +66,15 @@ const (
...
 
 	ConnectivityCheckAlpineCurlImage = "quay.io/cilium/alpine-curl:v1.4.0@sha256:2550c747831ff575f2147149b088ea981c06f9b6bcd188756d1b82cc10997956"
+	ConnectivityPerformanceImage     = "quay.io/cilium/network-perf:bf58fb8bc57c4933dfa6e2a9581d3925c0a0571e@sha256:9bef508b2dcaeb3e288a496b8d3f065e8636a4937ba3aebcb1732afffaccea34"
 	ConnectivityCheckJSONMockImage   = "quay.io/cilium/json-mock:v1.3.0@sha256:2729064827fa9dbfface8d3df424feb6c792a0ba07117b844349635c93c06d2b"
 
 	ConfigMapName   = "cilium-config"
 	Version         = "v1.11.2"
 	HubbleUIVersion = "v0.8.5"
 
+	PerfTestDuration = 30
```
tklauser, line 76: This const currently does not seem to be used. Do we want to use it in the definition of the `--perf-duration` flag?

connectivity/check/check.go
```diff
@@ -35,6 +35,9 @@ type Parameters struct {
 	Debug                 bool
 	PauseOnFail           bool
 	SkipIPCacheCheck      bool
+	Perf                  bool
+	Duration              int
```
tklauser, line 39: Change this to `time.Duration` (ref. my comment on the CLI flags definition):

```suggestion
	Duration              time.Duration
```

connectivity/check/check.go
```diff
@@ -35,6 +35,9 @@ type Parameters struct {
...
+	Samples               int
```
tklauser, line 40: Change the type to `time.Duration` and rename to include the`Perf...` prefix (ref. my comment on the CLI flags definition):

```suggestion
	PerfDuration          time.Duration
	PerfSamples           int
```

connectivity/check/context.go
```diff
@@ -48,6 +51,19 @@ type ConnectivityTest struct {
...
 }
 
+type PerfTests struct {
+	Pod  string
+	Test string
+}
+
+type PerfResult struct {
+	Metric   string
+	Duration int
```
tklauser, line 61: Let's change this to a `time.Duration` as well (and change the format specifier in `ct.Logf` from `%-15d` to `%-15s` so printing the result will show a nicely formatted duration, e.g.

```
🔥 Performance Test Summary
-----------------------------------------------------------------------------------------------------------------------------
📋 Scenario                                           | Test            | Num Samples     | Duration        | Avg value      
-----------------------------------------------------------------------------------------------------------------------------
📋 perf-client-other-node-674f5b4c96-bxbsq            | TCP_RR          | 1               | 1s              | 9088.24 (OP/s)
📋 perf-client-other-node-674f5b4c96-bxbsq            | TCP_CRR         | 1               | 1s              | 2779.66 (OP/s)
📋 perf-client-56cf954b5c-hgq8x                       | TCP_RR          | 1               | 1s              | 26610.28 (OP/s)
📋 perf-client-56cf954b5c-hgq8x                       | UDP_STREAM      | 1               | 1s              | 689.69 (Mb/s)
📋 perf-client-56cf954b5c-hgq8x                       | UDP_RR          | 1               | 1s              | 34702.63 (OP/s)
📋 perf-client-other-node-674f5b4c96-bxbsq            | TCP_STREAM      | 1               | 1s              | 887.87 (Mb/s)
📋 perf-client-other-node-674f5b4c96-bxbsq            | UDP_RR          | 1               | 1s              | 10092.40 (OP/s)
📋 perf-client-other-node-674f5b4c96-bxbsq            | UDP_STREAM      | 1               | 1s              | 1048.76 (Mb/s)
📋 perf-client-56cf954b5c-hgq8x                       | TCP_CRR         | 1               | 1s              | 7271.43 (OP/s)
📋 perf-client-56cf954b5c-hgq8x                       | TCP_STREAM      | 1               | 1s              | 666.74 (Mb/s)
-----------------------------------------------------------------------------------------------------------------------------
```

tklauser (changes requested):
Somehow, I've messed up my previous review and forgot to add some suggested changes regarding the `--perf-duration` and `--perf-samples` CLI flags. Here's the complete diff for my suggested changes:

```diff
diff --git a/connectivity/check/check.go b/connectivity/check/check.go
index 975122165b1b..bad74e2c1b76 100644
--- a/connectivity/check/check.go
+++ b/connectivity/check/check.go
@@ -36,8 +36,8 @@ type Parameters struct {
 	PauseOnFail           bool
 	SkipIPCacheCheck      bool
 	Perf                  bool
-	Duration              int
-	Samples               int
+	PerfDuration          time.Duration
+	PerfSamples           int
 	CiliumBaseVersion     string
 }
 
diff --git a/connectivity/check/context.go b/connectivity/check/context.go
index 82f5f39fe32b..10e703711075 100644
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -58,7 +58,7 @@ type PerfTests struct {
 
 type PerfResult struct {
 	Metric   string
-	Duration int
+	Duration time.Duration
 	Samples  int
 	Values   []float64
 	Avg      float64
@@ -322,7 +322,7 @@ func (ct *ConnectivityTest) report() error {
 		ct.Logf("📋 %-50s | %-15s | %-15s | %-15s | %-15s", "Scenario", "Test", "Num Samples", "Duration", "Avg value")
 		ct.Logf("%s", strings.Repeat("-", 125))
 		for p, d := range ct.PerfResults {
-			ct.Logf("📋 %-50s | %-15s | %-15d | %-15d | %.2f (%s)", p.Pod, p.Test, d.Samples, d.Duration, d.Avg, d.Metric)
+			ct.Logf("📋 %-50s | %-15s | %-15d | %-15s | %.2f (%s)", p.Pod, p.Test, d.Samples, d.Duration, d.Avg, d.Metric)
 			ct.Debugf("Individual Values from run : %s", d.Values)
 		}
 		ct.Logf("%s", strings.Repeat("-", 125))
diff --git a/connectivity/tests/perfpod.go b/connectivity/tests/perfpod.go
index 2fa45fe6b077..22c8531defc5 100644
--- a/connectivity/tests/perfpod.go
+++ b/connectivity/tests/perfpod.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/cilium/cilium-cli/connectivity/check"
 )
@@ -39,8 +40,8 @@ func (s *netPerfPodtoPod) Name() string {
 }
 
 func (s *netPerfPodtoPod) Run(ctx context.Context, t *check.Test) {
-	samples := t.Context().Params().Samples
-	duration := t.Context().Params().Duration
+	samples := t.Context().Params().PerfSamples
+	duration := t.Context().Params().PerfDuration
 	for _, c := range t.Context().PerfClientPods() {
 		c := c
 		for _, server := range t.Context().PerfServerPod() {
@@ -57,7 +58,7 @@ func (s *netPerfPodtoPod) Run(ctx context.Context, t *check.Test) {
 	}
 }
 
-func netperf(ctx context.Context, sip string, podname string, test string, a *check.Action, result map[check.PerfTests]check.PerfResult, samples int, duration int) {
+func netperf(ctx context.Context, sip string, podname string, test string, a *check.Action, result map[check.PerfTests]check.PerfResult, samples int, duration time.Duration) {
 	// Define test about to be executed and from which pod
 	k := check.PerfTests{
 		Pod:  podname,
@@ -68,7 +69,7 @@ func netperf(ctx context.Context, sip string, podname string, test string, a *ch
 		metric = "Mb/s"
 	}
 
-	exec := []string{"/usr/local/bin/netperf", "-H", sip, "-l", fmt.Sprintf("%d", duration), "-t", test, "--", "-R", "1", "-m", fmt.Sprintf("%d", messageSize)}
+	exec := []string{"/usr/local/bin/netperf", "-H", sip, "-l", fmt.Sprintf("%.0f", duration.Seconds()), "-t", test, "--", "-R", "1", "-m", fmt.Sprintf("%d", messageSize)}
 	//  recv socketsize		send socketsize 	msg size|okmsg	duration	value
 	values := []float64{}
 	// Result data
diff --git a/internal/cli/cmd/connectivity.go b/internal/cli/cmd/connectivity.go
index 5ff491c74a4a..ade391e7734f 100644
--- a/internal/cli/cmd/connectivity.go
+++ b/internal/cli/cmd/connectivity.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -122,8 +123,8 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().BoolVarP(&params.PauseOnFail, "pause-on-fail", "p", false, "Pause execution on test failure")
 	cmd.Flags().BoolVar(&params.SkipIPCacheCheck, "skip-ip-cache-check", true, "Skip IPCache check")
 	cmd.Flags().BoolVar(&params.Perf, "perf", false, "Run network Performance tests")
-	cmd.Flags().IntVar(&params.Duration, "perf-duration", 30, "Duration for the Performance test to run, Default:30s")
-	cmd.Flags().IntVar(&params.Samples, "perf-samples", 1, "Number of Performance samples to capture (how many times to run each test), Default:1")
+	cmd.Flags().DurationVar(&params.PerfDuration, "perf-duration", 30*time.Second, "Duration for the Performance test to run")
+	cmd.Flags().IntVar(&params.PerfSamples, "perf-samples", 1, "Number of Performance samples to capture (how many times to run each test)")
 	cmd.Flags().MarkHidden("skip-ip-cache-check")
 	cmd.Flags().StringVar(&params.CiliumBaseVersion, "base-version", defaults.Version,
 		"Specify the base Cilium version for configuration purpose in case image tag doesn't indicate the actual Cilium version")
```

internal/cli/cmd/connectivity.go
jtaleric: Nice, wasn't aware of this! Should of looked for the reference! Thanks!

tklauser (approved):
Thanks @rook! 🚀

LGTM once the commits addressing review feedback are squashed into the first one.

tklauser:
Let's :shipit:

PR #710, opened 2022-02-10
Prevent unmanaged pods in GKE's containerd flavors (cilium/cilium#18486)

bmcustodio:
Porting https://github.com/cilium/cilium/pull/18486 to the CLI.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -101,14 +101,14 @@ jobs:
             ${{ env.cost_reduction }} \
             --no-wait
 
-          # Create user node pool tainted with `node.cilium.io/agent-not-ready=true:NoSchedule`
+          # Create user node pool tainted with `node.cilium.io/agent-not-ready=true:NoExecute`
           az aks nodepool add \
             --resource-group ${{ env.name }} \
             --cluster-name ${{ env.name }} \
             --name userpool \
             --mode user \
             --node-count 2 \
-            --node-taints "node.cilium.io/agent-not-ready=true:NoSchedule" \
+            --node-taints "node.cilium.io/agent-not-ready=true:NoExecute" \
             ${{ env.cost_reduction }} \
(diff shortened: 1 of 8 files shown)


ldelossa (changes requested):
@bmcustodio 

After the CNI is installed, we expect any taints to be removed from nodes correct? I do not see any updates where we update the *removal* of the tain to now remove "NoExecute" taint. 

Am I missing something?

bmcustodio:
> I do not see any updates where we update the removal of the taint to now remove "NoExecute" taint.

@ldelossa I am not sure if I got the question right, but [Cilium does remove the taint regardless of its effect](https://github.com/cilium/cilium/blob/master/pkg/k8s/node.go#L251-L288).

bmcustodio:
It's also probably pointing out that this is just a port of a piece of code that has been [reviewed](https://github.com/cilium/cilium/pull/18486) upstream.

ldelossa:
@bmcustodio gotchya, I was assuming the taint being removed was hard coded, and we needed to update it.

christarazi (approved):

christarazi (commented):
What is the second commit msg referencing by saying?

```
[ upstream commit b049574 ]
```

bmcustodio:
@christarazi whenever I (back-)port something, I usually always include a reference to the original commit. In this case, it's just a reference to the original `c/c` commit, but it has no other meaning or isn't used by any tooling that I am aware of.

christarazi:
@bmcustodio That makes sense to me, but it's not clear where that commit SHA is coming from, so I'd suggest either linking it as a GH link or mentioning it's a cilium/cilium commit.

bmcustodio:
@christarazi sure, makes sense! I've rebased the PR and replaced the "upstream" mentions with "cilium/cilium" ones.

tklauser (approved):

Issue #719, opened 2022-02-23
Cilium connectivity check broken with Cilium 1.10 or earlier


joestringer:
```
[=] Test [no-policies]
................................
[=] Test [allow-all-except-world]

  ℹ️  📜 Applying CiliumNetworkPolicy 'allow-all-except-world' to namespace 'cilium-test'..
  ℹ️  Cilium agent kube-system/cilium-g4ncx logs since 2022-02-23 03:04:32.6667[45](https://github.com/cilium/cilium/runs/5297927757?check_suite_focus=true#step:14:45)278 +0000 UTC m=+40.3326265[46](https://github.com/cilium/cilium/runs/5297927757?check_suite_focus=true#step:14:46):

  ℹ️  Cilium agent kube-system/cilium-kvbqf logs since 2022-02-23 03:04:32.66674[52](https://github.com/cilium/cilium/runs/5297927757?check_suite_focus=true#step:14:52)78 +0000 UTC m=+40.332626[54](https://github.com/cilium/cilium/runs/5297927757?check_suite_focus=true#step:14:54)6:

  🔥 Running test allow-all-except-world: setting up test: applying network policies: policy application failed: CiliumNetworkPolicy.cilium.io "allow-all-except-world" is invalid: [spec.ingress.fromEntities: Unsupported value: "kube-apiserver": supported values: "all", "world", "cluster", "host", "init", "unmanaged", "remote-node", "health", "none", spec.egress.toEntities: Unsupported value: "kube-apiserver": supported values: "all", "world", "cluster", "host", "init", "unmanaged", "remote-node", "health", "none"]

[=] Test [client-ingress]
..
[=] Test [echo-ingress]
....
[=] Test [client-egress]
....
[=] Test [to-entities-world]
......
[=] Test [to-cidr-1111]
....
[=] Skipping Test [echo-ingress-l7]

[=] Skipping Test [client-egress-l7]

[=] Test [dns-only]
..........
[=] Skipping Test [to-fqdns]

📋 Test Report
❌ 1/8 tests failed (0/[62](https://github.com/cilium/cilium/runs/5297927757?check_suite_focus=true#step:14:62) actions), 3 tests skipped, 0 scenarios skipped:
Connectivity test failed: 1 tests failed
Test [allow-all-except-world]:
Error: Process completed with exit code 1.
```

PR #720, opened 2022-02-23
Use specific policy depending on Cilium version

christarazi:
- connectivity: Run setup & validation before registering tests
- connectivity: Use specific policy depending on Cilium version

Fixes: https://github.com/cilium/cilium-cli/issues/719

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -354,6 +354,26 @@ func (ct *ConnectivityTest) logAggregationMode(ctx context.Context, client *k8s.
 	return strings.ToLower(v), nil
 }
 
+// FetchCiliumPodImageTag fetches the first Cilium pod's image's tag (e.g.
+// v1.11.1 from quay.io/cilium/cilium:v1.11.1).
+func (ct *ConnectivityTest) FetchCiliumPodImageTag() string {
+	var img string
+	for _, pod := range ct.ciliumPods {
+		cntrs := pod.Pod.Spec.Containers
+		for _, c := range cntrs {
+			if strings.Contains(c.Name, "cilium-agent") && c.Image != "" {
+				img = c.Image
+				break
+			}
+		}
+	}
+	spl := strings.Split(img, ":")
+	if len(spl) > 1 {
+		return spl[1]
+	}
+	return ""
+}
+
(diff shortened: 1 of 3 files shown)


michi-covalent (approved):
:ship:

connectivity/check/context.go
```diff
@@ -348,6 +354,26 @@ func (ct *ConnectivityTest) logAggregationMode(ctx context.Context, client *k8s.
...
 }
 
+// FetchCiliumPodImageTag fetches the first Cilium pod's image's tag (e.g.
+// v1.11.1 from quay.io/cilium/cilium:v1.11.1).
+func (ct *ConnectivityTest) FetchCiliumPodImageTag() string {
+	var img string
+	for _, pod := range ct.ciliumPods {
+		cntrs := pod.Pod.Spec.Containers
+		for _, c := range cntrs {
+			if strings.Contains(c.Name, "cilium-agent") && c.Image != "" {
```
tklauser, line 364: Nit: s/"cilium-agent"/defaults.AgentContainerName/

connectivity/suite.go
```diff
@@ -58,19 +63,35 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
...
-			tests.PodToService(""),
-			// We are skipping the following checks because NodePort is
-			// intended to be used for N-S traffic, which conflicts with
-			// policies. See GH-17144.
-			// tests.PodToRemoteNodePort(""),
-			// tests.PodToLocalNodePort(""),
-			tests.PodToHost(""),
-			tests.PodToExternalWorkload(""),
-		)
+	if strings.HasPrefix(ct.FetchCiliumPodImageTag(), "v1.11") {
```
tklauser, line 66: I think this will not work e.g. for CI images in `cilium/cilium`, where the tag will be some hash. I assume we'd mistakenly use the pre 1.11 version in such cases.

To fix this, I'd suggest to add a `--base-version` CLI arg for `cilium connectivity test` (similar to the one for `cilium install`) which will allow to set the assumed base version for such cases in CI. Alternatively we could just fall back to `defaults.Version` in case the image tag is not a semver version.

PR #732, opened 2022-03-01
connectivity: Don't ignore errors during setup & validation

pchaigno:
This was causing a timeout during the resolution of services to go unnoticed as it was never logged.

cc @christarazi @tklauser 

Fixes: https://github.com/cilium/cilium-cli/pull/720.


tklauser (approved):
Thanks

PR #739, opened 2022-03-03
Fix 'node-init' in GKE's 'cos' images.

bmcustodio:
Turns out that in GKE's `cos` images the `containerd` binary is still present even though `/etc/containerd/config.toml` is not. Hence, the kubelet wrapper would still be installed for these images according to the current check, even though it's not necessary. What's worse, starting the kubelet would fail because the `sed` command targeting the aforementioned file would fail.

This PR changes the check to rely on the presence of the `--container-runtime-endpoint` flag in the kubelet, which is probably a more reliable way of detecting `*_containerd` flavours and only applying the fix in these cases.

### DIFF
--- a/install/node_init.go
+++ b/install/node_init.go
@@ -34,9 +34,11 @@ fi
 echo 'Done!'
 `),
 		k8s.KindGKE: bash(`
-# Check if we're running on a GKE containerd flavor.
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
:ship:

PR #744, opened 2022-03-04
install: Handle rancher-desktop CNI location

chancez:
Automatically detect if being installed onto rancher-desktop by checking
if the kubeconfig context/cluster name matches 'rancher-desktop'


ldelossa (changes requested):
Just some sanity checks:
- Will this always be the name of the cluster with ranger?
- I'm assuming if the cluster name CAN change, context will not, yes?

Only block on questions, code looks fine.

chancez:
> Will this always be the name of the cluster with ranger?

Unless the user changes the name of their cluster/context, it should not change unless rancher-desktop changes how they update the kubeconfig.

> I'm assuming if the cluster name CAN change, context will not, yes?

Same as above.

In general it's similar to Minikube.

ldelossa (approved):

aanm:
Let's wait until https://github.com/cilium/cilium-cli/pull/717 is merged

sayboras (approved):
LGTM 💯 

Unlike other flavours (e.g. kind, eks, gke, etc), rancher-desktop requires some override configuration as mentioned in https://github.com/cilium/cilium/pull/19049#pullrequestreview-901386060 before running cilium install command, so it will be great if we can print link pointing to such docs, so that it's more visible to user.

chancez:
@sayboras  The changes are similar to what's required for Minikube, I'm not sure what else could be done besides the docs which we already merged?

sayboras:
> The changes are similar to what's required for Minikube, I'm not sure what else could be done besides the docs which we already merged?

All good, the above comment might be added before I checked related PR in cilium repo.

tklauser:
> Let's wait until #717 is merged

This PR needs a rebase now that #717 is merged.

PR #748, opened 2022-03-08
internal: clarify key=value syntax for cilium install --config

### LINKED ISSUE #747: cilium install --config only takes one key=value
## Bug report

**General Information**

- Cilium CLI version (run `cilium version`) 0.10.4
- Orchestration system version in use (e.g. `kubectl version`, ...) 1.23.4
- Platform / infrastructure information (e.g. AWS / Azure / GCP, image / kernel versions) n/a
- Link to relevant artifacts (policies, deployments scripts, ...) n/a
- Generate and upload a system zip: `cilium sysdump` n/a

I would like to request clarification in the `--help` that `--config` needs to be called for each key=value that you wish to set.

I just set up my cluster with ` --config enable-ipv6=true enable-ipv6-masquerade=false k8s-api-server=https://ebpf-test.example.com:6443` and i was a bit surprised that it didn't take the second+third key=value because the current flag states the plural `strings` not `string`.

Perhaps I should report this as a bug with the --config option instead because it only parses the first string?

Output from cilium install with `--config enable-ipv6=true enable-ipv6-masquerade=false k8s-api-server=https://ebpf-test.example.com:6443`
```
🚀 Creating ConfigMap for Cilium version 1.11.2...
ℹ️ Manual overwrite in ConfigMap: enable-ipv6=true
🚀 Creating Agent DaemonSet...
```
And with a  `--config` for each key-value:
```
🚀 Creating Cluster roles...
🚀 Creating ConfigMap for Cilium version 1.11.2...
ℹ️ Manual overwrite in ConfigMap: enable-ipv6=true
ℹ️ Manual overwrite in ConfigMap: enable-ipv6-masquerade=false
ℹ️ Manual overwrite in ConfigMap: k8s-api-server=https://ebpf-test.example.com:6443
🚀 Creating Agent DaemonSet...
```

vosdev:
internal: clarify key=value syntax for cilium install --config

Signed-off-by: Daniel Vos <danielvos@outlook.com>

fixes #747

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -66,7 +66,7 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().BoolVar(&params.RestartUnmanagedPods, "restart-unmanaged-pods", true, "Restart pods which are not being managed by Cilium")
 	cmd.Flags().StringVar(&params.Encryption, "encryption", "disabled", "Enable encryption of all workloads traffic { disabled | ipsec | wireguard }")
 	cmd.Flags().BoolVar(&params.NodeEncryption, "node-encryption", false, "Enable encryption of all node to node traffic")
-	cmd.Flags().StringSliceVar(&params.ConfigOverwrites, "config", []string{}, "Set ConfigMap entries (key=value)")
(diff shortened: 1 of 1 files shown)


twpayne (approved):
Thanks!

PR #750, opened 2022-03-08
Make CRR standalone workload

### LINKED ISSUE #749: [Performance] Break CRR into its own test
## Proposal / RFE

Separate CRR into its own test due to conntrack limits. 

```cat /proc/sys/net/nf_conntrack_max 
131072
```

Running CRR for multiple samples / extended duration can result in test failures. To avoid the user hitting this problem, we should still allow the user to run CRR, but stand-alone with pre-defined samples and duration. 

**Is your feature request related to a problem?**

Yes. Extended runs for CRR will result in test failures.

jtaleric:
CRR can quickly burn through `nf_conntrack_max` if the user passes too
many `--perf-samples` or an extended `--perf-duration`.

Adding `--perf-crr` to allow the user to run the CRR test, but it will
be a single sample of 30s.

closes #749 

Signed-off-by: Joe Talerico <rook@isovalent.com>

### DIFF
--- a/connectivity/check/check.go
+++ b/connectivity/check/check.go
@@ -37,6 +37,7 @@ type Parameters struct {
 	SkipIPCacheCheck      bool
 	Perf                  bool
 	PerfDuration          time.Duration
+	PerfCRR               bool
 	PerfSamples           int
 	CiliumBaseVersion     string
 }
--- a/connectivity/tests/perfpod.go
+++ b/connectivity/tests/perfpod.go
@@ -42,17 +42,21 @@ func (s *netPerfPodtoPod) Name() string {
 func (s *netPerfPodtoPod) Run(ctx context.Context, t *check.Test) {
 	samples := t.Context().Params().PerfSamples
 	duration := t.Context().Params().PerfDuration
+	crr := t.Context().Params().PerfCRR
 	for _, c := range t.Context().PerfClientPods() {
 		c := c
 		for _, server := range t.Context().PerfServerPod() {
-			action := t.NewAction(s, "iperf-tcp", &c, server)
+			action := t.NewAction(s, "netperf", &c, server)
 			action.CollectFlows = false
 			action.Run(func(a *check.Action) {
-				netperf(ctx, server.Pod.Status.PodIP, c.Pod.Name, "TCP_RR", a, t.Context().PerfResults, samples, duration)
-				netperf(ctx, server.Pod.Status.PodIP, c.Pod.Name, "TCP_CRR", a, t.Context().PerfResults, samples, duration)
-				netperf(ctx, server.Pod.Status.PodIP, c.Pod.Name, "TCP_STREAM", a, t.Context().PerfResults, samples, duration)
-				netperf(ctx, server.Pod.Status.PodIP, c.Pod.Name, "UDP_RR", a, t.Context().PerfResults, samples, duration)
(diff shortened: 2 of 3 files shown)


connectivity/tests/perfpod.go
```diff
@@ -42,17 +42,21 @@ func (s *netPerfPodtoPod) Name() string {
 func (s *netPerfPodtoPod) Run(ctx context.Context, t *check.Test) {
 	samples := t.Context().Params().PerfSamples
 	duration := t.Context().Params().PerfDuration
+	crr := t.Context().Params().PerfCRR
 	for _, c := range t.Context().PerfClientPods() {
 		c := c
 		for _, server := range t.Context().PerfServerPod() {
-			action := t.NewAction(s, "iperf-tcp", &c, server)
+			action := t.NewAction(s, "netperf", &c, server)
```
twpayne, line 49: nit: This changes the benchmarking tool, which is not mentioned in the commit message. Is this change intentional?

internal/cli/cmd/connectivity.go
```diff
@@ -66,11 +66,12 @@ func newCmdConnectivityTest() *cobra.Command {
 
 			// Instantiate the test harness.
 			cc, err := check.NewConnectivityTest(k8sClient, params)
+
 			if params.PerfSamples > 3 {
-				cc.Warn("Having Performance Samples > 3 can cause issues")
+				cc.Warn("Having Performance samples > 3 can cause issues")
```
twpayne, line 71: Is it worth documenting what the "issues" are? Such documentation would be helpful for future users of the code.

Also, for CRR, `params.PerfSamples` and `params.PerfDuration` are ignored, so the warning could be misleading here, as the values supplied by the user are not actually used.

connectivity/check/check.go
```diff
@@ -37,6 +37,7 @@ type Parameters struct {
 	SkipIPCacheCheck      bool
 	Perf                  bool
 	PerfDuration          time.Duration
+	PerfCRR               bool
```
twpayne, line 40: If tweaks for other workloads are going to be required, it might be worth using a `string` to specify the mode (e.g. `--perf=crr`) rather than one `bool` for each mode. This is a bit more extensible, and ensures that only one workload type can be specified.

internal/cli/cmd/connectivity.go
jtaleric: ah, happy to. I linked the issue that this PR fixes : https://github.com/cilium/cilium-cli/issues/749 

Samples / Duration are only not applied if the user passes `-perf-crr. `

jtaleric: ah! I had to look again, i see your point!

jtaleric: With CRR being an opinionated workload (not allowing the user to modify the samples,duration) we can effectively remove these warnings.

connectivity/check/check.go
jtaleric: Moving CRR to an island wrt the benchmarks should avoid us from having to partition the workload(s) any further. However, it might make sense in a follow on to allow the user to pass a string as you mentioned above in a different PR.

Thoughts?

twpayne (approved):

ldelossa (approved):
LGTM.

PR #760, opened 2022-03-17
Moving pkill to cleanup

jtaleric:
Making the workflow look more closely to what we do in cilium/cilium,
where we kill the port-forward before uninstalling.

Signed-off-by: Joe Talerico <rook@isovalent.com>


jtaleric:
cc @nbusseneau

joestringer (approved):

tklauser (approved):

PR #767, opened 2022-03-28
Deprecate old CLI flags

aanm:
Follow up of https://github.com/cilium/cilium-cli/pull/717

The PR is deprecating flags with `MarkDeprecate` method and giving priority to helm flags in case both CLI and helm flags are used. If all what people care is about `MarkDeprecate` then we can discuss that afterwards leave the flag prioritization on this PR.

### DIFF
--- a/.github/in-cluster-test-scripts/aks-install.sh
+++ b/.github/in-cluster-test-scripts/aks-install.sh
@@ -6,10 +6,10 @@ set -e
 # Install Cilium
 cilium install \
   --disable-check=az-binary \
-  --azure-subscription-id "${AZURE_SUBSCRIPTION_ID}" \
-  --azure-node-resource-group "${AZURE_NODE_RESOURCE_GROUP}" \
-  --azure-tenant-id "${AZURE_TENANT_ID}" \
-  --azure-client-id "${AZURE_CLIENT_ID}" \
-  --azure-client-secret "${AZURE_CLIENT_SECRET}" \
+  --helm-set=azure.subscriptionID="${AZURE_SUBSCRIPTION_ID}" \
+  --helm-set=azure.resourceGroup="${AZURE_NODE_RESOURCE_GROUP}" \
+  --helm-set=azure.tenantID="${AZURE_TENANT_ID}" \
+  --helm-set=azure.clientID="${AZURE_CLIENT_ID}" \
+  --helm-set=azure.clientSecret="${AZURE_CLIENT_SECRET}" \
   --wait=false \
-  --config monitor-aggregation=none
+  --helm-set=bpf.monitorAggregation=none
--- a/.github/in-cluster-test-scripts/external-workloads-install.sh
+++ b/.github/in-cluster-test-scripts/external-workloads-install.sh
@@ -5,11 +5,11 @@ set -e
 
 # Install Cilium in cluster
 cilium install \
-  --cluster-name "${CLUSTER_NAME}" \
-  --config monitor-aggregation=none \
-  --config tunnel=vxlan \
-  --kube-proxy-replacement=strict \
-  --ipv4-native-routing-cidr="${CLUSTER_CIDR}"
+  --helm-set=cluster.name="${CLUSTER_NAME}" \
+  --helm-set=bpf.monitorAggregation=none \
+  --helm-set=tunnel=vxlan \
+  --helm-set=kubeProxyReplacement=strict \
+  --helm-set=ipv4NativeRoutingCIDR="${CLUSTER_CIDR}"
 
 # Enable Relay
 cilium hubble enable
--- a/install/install.go
+++ b/install/install.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cilium/cilium/api/v1/models"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	"github.com/cilium/cilium/pkg/versioncheck"
+	"github.com/spf13/pflag"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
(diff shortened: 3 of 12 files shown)


.github/in-cluster-test-scripts/aks-install.sh
```diff
@@ -6,10 +6,10 @@ set -e
...
-  --azure-client-id "${AZURE_CLIENT_ID}" \
-  --azure-client-secret "${AZURE_CLIENT_SECRET}" \
+  --helm-set=azure.subscriptionID="${AZURE_SUBSCRIPTION_ID}" \
+  --helm-set=azure.resourceGroup="${AZURE_NODE_RESOURCE_GROUP}" \
+  --helm-set=azure.tenantID="${AZURE_TENANT_ID}" \
+  --helm-set=azure.clientID="${AZURE_CLIENT_ID}" \
+  --helm-set=azure.clientSecret="${AZURE_CLIENT_SECRET}" \
   --wait=false \
-  --config monitor-aggregation=none
+  --helm-set=extraConfig.monitor-aggregation=none
```
sayboras, line 15: understand that it's existing way, just curious if we can use bpf.monitorAggregation directly here ?

.github/in-cluster-test-scripts/external-workloads-install.sh
```diff
@@ -5,11 +5,11 @@ set -e
 
 # Install Cilium in cluster
 cilium install \
-  --cluster-name "${CLUSTER_NAME}" \
-  --config monitor-aggregation=none \
+  --helm-set=cluster.name="${CLUSTER_NAME}" \
+  --helm-set=extraConfig.monitor-aggregation=none \
   --config tunnel=vxlan \
```
sayboras, line 10: should we change this as well ? basically, I am fine with either direction (e.g. --helm-set=extraConfig vs --config), just more about consistency in the same file/step.

internal/cli/cmd/install.go
```diff
@@ -48,39 +48,53 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	}
 
 	cmd.Flags().StringVarP(&params.Namespace, "namespace", "n", "kube-system", "Namespace to install Cilium into")
+	// Deprecated
```
sayboras, line 51: we can mark this flag as deprecated i.e. this flag will still work, but will not show up in help output.

```
	cmd.Flags().MarkDeprecated("cluster-name", "please use abc instead")
```

internal/cli/cmd/install.go
```diff
@@ -91,6 +105,20 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
 	cmd.Flags().StringArrayVar(&params.HelmOpts.FileValues, "helm-set-file", []string{}, "Set helm values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)")
 	cmd.Flags().StringVar(&params.HelmGenValuesFile, "helm-auto-gen-values", "", "Write an auto-generated helm values into this file")
 
+	for flagName, helmOpt := range install.DeprecatedFlagsToHelmOpts {
```
sayboras, line 108: related to the above, pflag can print usageMessage for deprecated flag for us.

sayboras (commented):
Couple of comments as per below, tested this pr changes so far so good :100:

internal/cli/cmd/install.go
```diff
@@ -91,6 +105,20 @@ cilium install --context kind-cluster1 --cluster-id 1 --cluster-name cluster1
...
+		msg := fmt.Sprintf("use --helm-set=%s<=value> instead", helmOpt)
+		err := cmd.Flags().MarkDeprecated(flagName, msg)
+		if err != nil {
+			panic(err)
+		}
+		install.DeprecatedFlagsValues[flagName] = cmd.Flags().Lookup(flagName).Value
+	}
+	err := cmd.Flags().MarkDeprecated("config", "use --helm-set=extraConfig.<key>=<value>[,extraConfig.<key>=<value>] instead")
```
tklauser, line 116: Personally I think `--config` is a useful short-hand to set a value in the `cilium-config` CM from the command line. I'd thus be inclined to keep this as an alias and not deprecate it.

Also, as mentioned in https://github.com/cilium/cilium-cli/pull/717#issuecomment-1068830584 we also use `--config` in our getting started guides. So users would get deprecation warnings when using commands copied from the guides. I case we decide to deprecate `--config` I think we should at least make sure to update the guides for v1.10/v1.11/master to use the new flag syntax.

internal/cli/cmd/install.go
sayboras: yup, I missed that

sayboras (approved):
As discussed offline, passing version with image suffix (e.g. --version -ci:4fac771179959ca575eb6f993d566653d3bfa167) is not supported right now, this could be addressed seperately if required.

rolinh:
I concur with @ti-mo and @chancez that removing old flags probably isn't a good move for user experience as it removes the discoverability aspect. If we are committed to removing them, could we at least build advanced completion for the `--helm-set` flag?

install/install.go
```diff
@@ -167,6 +168,28 @@ type AzureParameters struct {
...
+var (
+	// DeprecatedFlagsToHelmOpts maps the deprecated install flags to the helm
+	// options
+	DeprecatedFlagsToHelmOpts = map[string]string{
+		"agent-image":              "image.override",
+		"azure-client-id":          "azure.clientID",
+		"azure-client-secret":      "azure.clientSecret",
+		"azure-resource-group":     "azure.resourceGroup",
+		"azure-subscription-id":    "azure.subscriptionID",
+		"azure-tenant-id":          "azure.tenantID",
```
gandro, line 180: In my opinion, we should not deprecate those flag. It will add unnecessary cognitive load (what is helm? what is this helm-set doing? where is the reference for valid `helm-set` values?) to anyone trying out Cilium for the first time on Azure. 

With the PR as is, running `cilium install` on Azure will now also tell users to use a deprecated flag:
https://github.com/cilium/cilium-cli/blob/ad8d0d82d8ff16550bb9b05159fbe12999875cf7/install/azure.go#L115

I think this is a clear downgrade in user experience.

gandro (changes requested):

install/install.go
gandro: Yes. I think having overlap is not necessarily a bad thing, if it helps with discovery of important options (such as those parameters  which every AKS user has to set).

install/install.go
```diff
@@ -167,6 +168,28 @@ type AzureParameters struct {
 	ClientSecret         string
 }
 
+var (
+	// DeprecatedFlagsToHelmOpts maps the deprecated install flags to the helm
```
gandro, line 172: nit: Now that these flags are not deprecated anymore, should we name this maybe just `FlagsToHelmOpts`. Or what's your long-term plan for them (also with regards to the TODOs you added)?

gandro (approved):

install/install.go
aanm: Renamed. I've left the code commented for now if we later decide to mark them as deprecated.

tklauser (approved):

PR #773, opened 2022-03-30
Autodetect K3s cluster flavor

tklauser:
The first commit removes an unnecessary always-nil error return when auto-detecting the cluster flavor. The second commit add support for auto-detecting K3s clusters. Example `cilium install` output:

```
$ cilium install
🔮 Auto-detected Kubernetes kind: K3s
ℹ️  using Cilium version "v1.11.2"
🔮 Auto-detected cluster name: default
🔮 Auto-detected IPAM mode: cluster-pool
ℹ️  helm template --namespace kube-system cilium cilium/cilium --version 1.11.2 --set cluster.name=default,ipam.mode=cluster-pool,kubeProxyReplacement=disabled,operator.replicas=1,serviceAccounts.cilium.name=cilium,serviceAccounts.operator.name=cilium-operator
🔑 Created CA in secret cilium-ca
🔑 Generating certificates for Hubble...
🚀 Creating Service accounts...
🚀 Creating Cluster roles...
🚀 Creating ConfigMap for Cilium version 1.11.2...
🚀 Creating Agent DaemonSet...
level=warning msg="spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[1].matchExpressions[0].key: beta.kubernetes.io/os is deprecated since v1.14; use \"kubernetes.io/os\" instead" subsys=klog
level=warning msg="spec.template.metadata.annotations[scheduler.alpha.kubernetes.io/critical-pod]: non-functional in v1.16+; use the \"priorityClassName\" field instead" subsys=klog
🚀 Creating Operator Deployment...
⌛ Waiting for Cilium to be installed and ready...
♻️  Restarting unmanaged pods...
♻️  Restarted unmanaged pod kube-system/helm-install-traefik-crd--1-gxtmd
♻️  Restarted unmanaged pod kube-system/helm-install-traefik--1-j9pr6
♻️  Restarted unmanaged pod kube-system/local-path-provisioner-84bb864455-v4z6f
♻️  Restarted unmanaged pod kube-system/metrics-server-ff9dbcb6c-6stf6
♻️  Restarted unmanaged pod kube-system/coredns-96cc4f57d-f7kcl
✅ Cilium was successfully installed! Run 'cilium status' to view installation health
```

PR #777, opened 2022-04-02
refactor: Expose k8s package

### LINKED ISSUE #618: Make internal k8sInstallerImplementation implementation public.
## Proposal / RFE

I'd like to embed the cilium installation into another application. Currently, the codebase and packages are really nicely decoupled into packages for general use and the "transport layer" with the CLI which only constructs the needed clients and calls functions (e.g. `Install()`) on them.
That way it should be fairly easy to reuse that.

The only thing missing is a public reusable implementation of the [`k8sInstallerImplementation` interface](https://github.com/cilium/cilium-cli/blob/master/install/install.go#L951) which is needed for the [`K8sInstaller`](https://github.com/cilium/cilium-cli/blob/master/install/install.go#L1137). 
If I understood the code correctly the one used by the CLI is in the [internal package](https://github.com/cilium/cilium-cli/blob/master/internal/k8s/client.go). IMO that's already a pretty nice, reusable implementation of the interface.
Therefore, it would be really great if you could move that out of the internal package and into the same package as the `K8sInstaller`, maybe.

### LINKED ISSUE #776: Using packages from cilium-cli as a vendored lib to install cilium programmatically inside an environment
## Proposal / RFE
Recently I was experimenting with `kind` inside a golang project to programmatically spin up a k8s cluster. The support for `kind` via `go` is pretty good but taking it to the next was to install `cilium` in the env for which I found that I have 2 options:
- run `helm` and deploy the cilium chart
- run `cilium-cli` and have this install + configure cilium inside my `kind` cluster

Both of the options work well but you need to have the `helm` binary, `helm` chart locally or the `cilium-cli` binary. Did you folks consider exposing the internals of the `cilium-cli` for other consumers to vendor in their code and use? I was chatting with @michi-covalent  about this and would be pretty nifty to have access to https://github.com/cilium/cilium-cli/blob/0af00d450144f2c656394f5d6391657050937825/install/install.go#L293-L305. Right now I cannot vendor this because it's included in the `internal` package which `golang` forbids:
```
could not import github.com/cilium/cilium-cli/internal/k8s (invalid use of internal package github.com/cilium/cilium-cli/internal/k8s)
```

michi-covalent:
- Move internal/k8s/ to k8s/ to make it public. This allows external
  projects to import the package to provision Cilium programmatically.
- Add a package level documentation to clarify that the APIs might
  change before cilium-cli hits v1.0.0.
- Update CODEOWNERS.

Fixes #618, fixes #776


tklauser:
Oops, merging #773 lead to a merge conflict here 🥺

clustermesh/certs.go
```diff
@@ -14,7 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/cilium-cli/defaults"
-	"github.com/cilium/cilium-cli/internal/k8s"
+	"github.com/cilium/cilium-cli/pkg/k8s"
```
tklauser, line 17: I'd probably move the package to `github.com/cilium/cilium-cli/k8s`, i.e. without the `pkg/` part. All other public packages are at the top level already.

tklauser (changes requested):
The following entry in `CODEOWNERS` will need an update too:

https://github.com/cilium/cilium-cli/blob/084aa4c0c51f284c02aff517f2735a53086fc452/CODEOWNERS#L31

Also, we might consider removing the `type k8s.*Implementation interface` types from all the packages now that the `k8s.Client` type is public. But that's for a follow-up PR, probably.

clustermesh/certs.go
michi-covalent: updated!

sayboras (approved):
LGTM :) 

I just did a quick test, seems like k8s is the only package that we need to expose. From consumer perspective, there might be some work to improve UX (such as defaults value for each component, etc), but let's wait for feedback from actual usage.

<details>
<summary>Simple program to test</summary>

```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/cilium/cilium-cli/clustermesh"
	"github.com/cilium/cilium-cli/hubble"
	"github.com/cilium/cilium-cli/install"
	"github.com/cilium/cilium-cli/k8s"
)

func main() {
	client, err := k8s.NewClient("minikube", "/home/tammach/.kube/config")
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	installer, err := install.NewK8sInstaller(client, install.Parameters{
		Namespace:   "kube-system",
		Writer:      os.Stdout,
		ClusterName: "minikube",
		Version:     defaults.Version,
		Wait:        true,
		ClusterID:   1,
		Encryption:  "disabled",
	})
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	err = installer.Install(context.TODO())
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	h := hubble.NewK8sHubble(client, hubble.Parameters{
		Namespace: "kube-system",
		Writer:    os.Stdout,
		UI:        true,
	})
	if err = h.Enable(context.TODO()); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	checker, err := check.NewConnectivityTest(client, check.Parameters{
		CiliumNamespace:   "kube-system",
		TestNamespace:     defaults.ConnectivityCheckNamespace,
		FlowValidation:    "disabled",
		Writer:            os.Stdout,
		CiliumBaseVersion: defaults.Version,
	})
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	if err = checker.Run(context.TODO()); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}

	clusterMesh := clustermesh.NewK8sClusterMesh(client, clustermesh.Parameters{
		Namespace:   "kube-system",
		ServiceType: "NodePort",
		Writer:      os.Stdout,
	})
	if err = clusterMesh.Enable(context.TODO()); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}

```

</details>

tklauser (approved):
🎉

PR #780, opened 2022-04-06
sysdump - Add --cilium-operator-namespace option

jaredledvina:
Small pull request to allow passing in `--cilium-operator-namespace` to support use cases where the `cilium-operator` runs in a different namespace than the `cilium-agent`.


michi-covalent (approved):
lgtm. added @aanm and @tklauser as reviewers as well just to double check it's ok to run operator and cilium in different namespaces.

aanm:
@jaredledvina currently it's not supported installing Cilium Operator on a different namespace other than Cilium. How did you install Cilium Operator in a way that it was running on a different namespace?

aanm (approved):

PR #782, opened 2022-04-09
store helm values into a Kubernetes secret

aanm:
Similarly to helm behavior, Cilium-cli will store its generated yaml
values file in a secret.


michi-covalent (approved):
no need to do anything in this pr, but i do wonder if we could just run `helm install` and let helm keep track of the installation state.

aanm:
> no need to do anything in this pr, but i do wonder if we could just run `helm install` and let helm keep track of the installation state.

@michi-covalent it depends on what we decide to use in the future. Right now, cilium-cli has the control over which manifests we want to install. Also perhaps in the future we will have the ability to control how we perform the upgrade of each manifests and which configmap options should be enabled or not. We kind of lose control of that by doing helm install

twpayne (approved):

sayboras (approved):
LGTM :shipit:

tklauser (approved):

PR #784, opened 2022-04-10
hubble: redact cilium-ca certificate key when printing to terminal

aanm:
This commit redacts the certificate key when printing to terminal. User
can still revert this mode by running
--redact-helm-certificate-keys=false


michi-covalent (approved):

hubble/hubble.go
```diff
@@ -156,6 +159,18 @@ func NewK8sHubble(client k8sHubbleImplementation, p Parameters) *K8sHubble {
 }
 
 func (k *K8sHubble) Log(format string, a ...interface{}) {
+	if k.params.RedactHelmCertKeys {
+		formattedString := fmt.Sprintf(format+"\n", a...)
+		for _, certKey := range []string{
+			base64.StdEncoding.EncodeToString(k.certManager.CAKeyBytes()),
```
twpayne, line 165: This assumes that the CA key bytes were encoded with `base64.StdEncoding` and will break if the encoding changes in the future. I suggest factoring out `base64.StdEncoding.EncodeToString` into a separate `encodeCAKeyBytes` function and calling this function both here and where this function is called (i.e. at lines 361 and 362 of this file).

aanm:
@tklauser Fixed, I noticed that yesterday as well but the fix was on a different branch. It should be good now.

hubble/hubble.go
```diff
@@ -161,6 +163,18 @@ func NewK8sHubble(client k8sHubbleImplementation, p Parameters) *K8sHubble {
...
+	if k.params.RedactHelmCertKeys {
+		formattedString := fmt.Sprintf(format+"\n", a...)
+		for _, certKey := range []string{
+			certs.EncodeCertBytes(k.certManager.CAKeyBytes()),
+		} {
+			if certKey != "" {
+				format = strings.ReplaceAll(formattedString, certKey, "[--- REDACTED WHEN PRINTING TO TERMINAL (USE --redact-helm-certificate-keys=false TO PRINT) ---]")
+			}
+		}
+		fmt.Fprint(k.params.Writer, format)
```
tklauser, line 175:
```suggestion
			if certKey != "" {
				formattedString = strings.ReplaceAll(formattedString, certKey, "[--- REDACTED WHEN PRINTING TO TERMINAL (USE --redact-helm-certificate-keys=false TO PRINT) ---]")
			}
		}
		fmt.Fprint(k.params.Writer, formattedString)
```

otherwise the format specifiers won't be interpreted, leading to e.g. the following output on `cilium uninstall`:
 
```
🔥 Deleting Hubble UI...🔥 Deleting Hubble UI certificates...🔥 Deleting Relay...🔥 Deleting Relay certificates...✨ Patching ConfigMap %s to disable Hubble...🚀 Creating ConfigMap for Cilium version %s...♻️  Restarted Cilium podsℹ️  Storing helm values file in %s/%s Secret✅ Hubble was successfully disabled.🔮 Auto-detected Kubernetes kind: GKE
```

tklauser (changes requested):

tklauser (approved):

PR #787, opened 2022-04-12
connectivity: Make fatal errors more visible

twpayne:
The fire emoji does not stand out in logs. Replace it with one that
does.

Signed-off-by: Tom Payne <tom@isovalent.com>


michi-covalent (approved):
🚢

tklauser (approved):
LGTM. FWIW the previous emoji also collides with the performance tests using it in a log header: https://github.com/cilium/cilium-cli/blob/77cc2f4ebde7143387032dcc8307cea139fdf336/connectivity/check/context.go#L320

PR #790, opened 2022-04-13
defaults: add image digests for Cilium v1.12.0-rc1

tklauser:


tklauser:
These digest are now only used for `cilium upgrade` and `cilium clustermesh enable`. For these it's not worth adding RC release digests.

PR #792, opened 2022-04-13
Transform flags context and namespace to be global for cilium-cli

thorbenvh8:
I would like the `cilium-cli` to not fail if `--context` or `--namespace` are provided to high level commands like in `kubectl`. We use this very heavily to add this via a little function and read from Environment variables to use tools like kubectl stern etc easily inside a terminal without needing to provide `--context` and `--namespace` specifically every time, or even run commands in different environments with different context.

Signed-off-by: Thorben von Hacht [thorbenvh8@gmail.com](mailto:thorbenvh8@gmail.com)


twpayne (approved):
Looks great, thanks!

Issue #716, opened 2022-02-17
connectivity: Start with initial DNS check


pchaigno:
When basic connectivity across nodes is not working, the connectivity tests often fail with:
```
$ ./cilium connectivity test --namespace cilium
⌛ [test-ipsec-1] Waiting for deployments [client client2 echo-same-node] to become ready...
⌛ [test-ipsec-1] Waiting for deployments [echo-other-node] to become ready...
⌛ [test-ipsec-1] Waiting for CiliumEndpoint for pod cilium-test/client-7b7bf54b85-nrpzc to appear...
⌛ [test-ipsec-1] Waiting for CiliumEndpoint for pod cilium-test/client2-666976c95b-pnwq4 to appear...
⌛ [test-ipsec-1] Waiting for CiliumEndpoint for pod cilium-test/echo-other-node-697d5d69b7-t92s9 to appear...
⌛ [test-ipsec-1] Waiting for CiliumEndpoint for pod cilium-test/echo-same-node-7967996674-spmhf to appear...
⌛ [test-ipsec-1] Waiting for Service cilium-test/echo-other-node to become ready...

[...]

PR #799, opened 2022-04-20
Perform DNS check earlier in the tests

alexander-demichev:
This PR does the following:

- Check that DNS works before trying to wait for the service, because the service wait function relies on nslookup.
- If service wait fails, make logs print only namespace and name of the service, not dump of the entire object.

Fixes: https://github.com/cilium/cilium-cli/issues/716


tklauser (approved):
Thanks @alexander-demichev! Sorry this took so long to review.

PR #806, opened 2022-04-25
fix(helm): force extraConfigMap values as string

raphink:
Signed-off-by: Raphaël Pinson <raphael@isovalent.com>

### DIFF
--- a/install/helm.go
+++ b/install/helm.go
@@ -228,7 +228,7 @@ func (k *K8sInstaller) generateManifests(ctx context.Context) error {
 	// Store all the options passed by --config into helm extraConfig
 	extraConfigMap := map[string]interface{}{}
 	for k, v := range deprecatedCfgOpts {
-		extraConfigMap[k] = v
+		extraConfigMap[k] = string(v)
 	}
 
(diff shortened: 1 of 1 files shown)


install/helm.go
```diff
@@ -228,7 +228,7 @@ func (k *K8sInstaller) generateManifests(ctx context.Context) error {
 	// Store all the options passed by --config into helm extraConfig
 	extraConfigMap := map[string]interface{}{}
 	for k, v := range deprecatedCfgOpts {
-		extraConfigMap[k] = v
+		extraConfigMap[k] = string(v)
```
twpayne, line 231: This means that `extraConfigMap[k]` will have a `string` value if `v` is a string, or the empty string otherwise. Is this the desired behavior? Alternative code code be:

```go
if s, ok := v.(string); ok {
    extraConfigMap[k] = s // Happy case: v is a string
} else {
    log.Warn("%k value %v is not a string, ignoring", k, v)
}
```

raphink:
It seems I actually got on a wrong track trying to fix the service mesh beta lab. This might not be necessary after all.

PR #809, opened 2022-04-25
install: Add Peer Service on Install

nathanjsweet:
If the Peer Service file is present it should be installed.

Signed-off-by: Nate Sweet <nathanjsweet@pm.me>

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -65,6 +65,31 @@ func (k *K8sInstaller) generateAgentDaemonSet() *appsv1.DaemonSet {
 	return &ds
 }
 
+func (k *K8sInstaller) generatePeerService() *corev1.Service {
+	var (
+		svcFilename string
+	)
+	ciliumVer := k.getCiliumVersion()
+	switch {
+	case versioncheck.MustCompile(">1.11.0")(ciliumVer):
+		svcFilename = "templates/cilium-agent/peer-service.yaml"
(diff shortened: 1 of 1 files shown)


install/install.go
```diff
@@ -65,6 +65,31 @@ func (k *K8sInstaller) generateAgentDaemonSet() *appsv1.DaemonSet {
...
 
+func (k *K8sInstaller) generatePeerService() *corev1.Service {
+	var (
+		svcFilename string
+	)
+	ciliumVer := k.getCiliumVersion()
+	switch {
+	case versioncheck.MustCompile(">1.11.0")(ciliumVer):
+		svcFilename = "templates/cilium-agent/peer-service.yaml"
+	case versioncheck.MustCompile(">1.9.0")(ciliumVer):
```
michi-covalent, line 76: why do we need this for v1.10? isn't this only for v1.12 or later?

aanm: @nathanjsweet @michi-covalent this is not in the right place, it should be in the installer of hubble-relay [here](https://github.com/cilium/cilium-cli/blob/5b0e2d5d3e5fc28d0afc2db439800db7ee11afc5/hubble/relay.go#L162-L161). This also needs to be uninstalled, so that we clean up when hubble is removed which should be done [here](https://github.com/cilium/cilium-cli/blob/5b0e2d5d3e5fc28d0afc2db439800db7ee11afc5/hubble/relay.go#L94-L93)

PR #810, opened 2022-04-25
relay: Move Peer Service to Hubble Relay Enablement

nathanjsweet:
Signed-off-by: Nate Sweet <nathanjsweet@pm.me>

### DIFF
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -178,6 +178,31 @@ func (k *K8sHubble) Log(format string, a ...interface{}) {
 	fmt.Fprintf(k.params.Writer, format+"\n", a...)
 }
 
+func (k *K8sHubble) generatePeerService() *corev1.Service {
+	var (
+		svcFilename string
+	)
+	ciliumVer := k.semVerCiliumVersion
+	switch {
+	case versioncheck.MustCompile(">1.11.0")(ciliumVer):
+		svcFilename = "templates/cilium-agent/peer-service.yaml"
+	case versioncheck.MustCompile(">1.9.0")(ciliumVer):
+		svcFilename = "templates/cilium-agent-peer-service.yaml"
+	}
+	if svcFilename == "" {
+		return nil
+	}
+
+	svcFile, ok := k.manifests[svcFilename]
+	if !ok || len(strings.TrimSpace(svcFile)) == 0 {
+		return nil
+	}
+
+	var svc corev1.Service
+	utils.MustUnmarshalYAML([]byte(svcFile), &svc)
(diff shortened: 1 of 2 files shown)


michi-covalent (approved):

hubble/relay.go
```diff
@@ -128,6 +158,13 @@ func (k *K8sHubble) enableRelay(ctx context.Context) (string, error) {
 		return relayDeployment.GetName(), nil
 	}
 
+	if peerSvc := k.generatePeerService(); peerSvc != nil {
+		k.Log("🚀 Creating Peer Service...")
+		if _, err := k.client.CreateService(ctx, k.params.Namespace, peerSvc, metav1.CreateOptions{}); err != nil {
+			return "", err
+		}
+	}
+
```
aanm, line 167: Sorry, I meant [here](https://github.com/cilium/cilium-cli/blob/8cb9027f770a159ee2b1374197b4a84e975d8132/hubble/hubble.go#L510)

hubble/relay.go
```diff
@@ -92,6 +118,10 @@ func (k *K8sHubble) disableRelay(ctx context.Context) error {
 	}
 	k.client.DeleteService(ctx, relaySvc.GetNamespace(), relaySvc.GetName(), metav1.DeleteOptions{})
 
+	if peerSvc := k.generatePeerService(); peerSvc != nil {
```
aanm, line 121: Sorry, I meant [here](https://github.com/cilium/cilium-cli/blob/8cb9027f770a159ee2b1374197b4a84e975d8132/hubble/hubble.go#L256-L255)

aanm (changes requested):

hubble/hubble.go
```diff
@@ -539,6 +568,12 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		warnFreePods = append(warnFreePods, podsName)
 	}
 
+	if peerSvc := k.generatePeerService(); peerSvc != nil {
```
aanm, line 571: I'm fine with the changes but it's up to you now, does it matter if the peer is only created after Relay and hubble? If it matters then the services needs to be created on line 561, otherwise here is fine.

aanm (approved):

hubble/hubble.go
nathanjsweet: No, it won't matter. Even if the relay fails to pick up the peer service (which it won't because it needs some time to spin up), it will try again later.

michi-covalent: alright let's merge it

Issue #519, opened 2020-04-13
Collect gops stats when cilium is not pid 1 inside the container


ungureanuvladvictor:
In our prod env we wrap the cilium agent when we start it to process some logs for it thus it does not have pid 1 anymore. From https://github.com/cilium/cilium-sysdump/blob/master/cilium-sysdump/sysdumpcollector.py#L211 looks like cilium-sysdump assumes that cilium is running as pid 1.

PR #817, opened 2022-04-28
Add option to automatically detect the pid of the gops agent

ungureanuvladvictor:
If we run `gops` standalone inside the container the output marks which process runs the agent by appending a `*` next to the binary name:
```
25863 0     gops          unknown Go version /usr/bin/gops
25852 25847 cilium        unknown Go version /usr/bin/cilium
10    1     cilium-agent* unknown Go version /usr/bin/cilium-agent
1     0     custom-binary        go1.16.3           /usr/local/bin/custom-binary
```

Opted to parse this output. An alternative is to implement what @nathanjsweet committed in https://github.com/cilium/cilium-sysdump/commit/688a81430a71e155e4bbc516e9bd17f8479f6129. I think the downside with that approach is that not all container images will have `cat` inside them (e.g., `cilium-operator` which is based on `FROM scratch` + `gops` + `ca-certificates` + the operator binary -- [Dockerfile](https://github.com/cilium/cilium/blob/master/images/operator/Dockerfile)).

Happy to discuss pro/cons on which one is better. I think both approaches need an extra k8s call. Added this as a configurable behavior so that users who do not wrap the process will not incur the extra k8s call.

Fixes https://github.com/cilium/cilium-cli/issues/519.

### DIFF
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -103,6 +103,9 @@ func newCmdSysdump() *cobra.Command {
 	cmd.Flags().StringArrayVar(&sysdumpOptions.CiliumBugtoolFlags,
 		"cilium-bugtool-flags", nil,
 		"Optional set of flags to pass to cilium-bugtool command.")
+	cmd.Flags().BoolVar(&sysdumpOptions.DetectGopsPID,
+		"detect-gops-pid", false,
+		"Whether to automatically detect the gops agent PID.")
 
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


nathanjsweet (approved):
Looks good. A couple of questions:

1. Why shouldn't we turn this on by default?
2. If there is more than one agent does that present a problem?

ungureanuvladvictor:
@nathanjsweet :
1. Did not make it default so we don't eat away another k8s call and make the sysdump process longer. Can def change that if you think it's ok from a UX perspective.
2. It will just find the 1st one -- I added a test for it at https://github.com/cilium/cilium-cli/pull/817/files#diff-55d826599a0e4be47cab8f4dd77dc5d06203959e2c67e656e4a9bcec94e7a999R128-R136. Not sure how to catch this case or what the behavior should be? Iterate over all the agents and extract from all of them?

PR #820, opened 2022-04-29
install: Wait for test pods to be cleaned up

aditighag:
If test pods deletion takes longer than that of (cni)
 cilium pods, test pods are stuck in the "terminating"
 state. This ends up failing CI jobs in the clean-up stage.
 Hence, wait until test pods are deleted before moving
 onto the CNI related clean-up.

Signed-off-by: Aditi Ghag <aditi@cilium.io>

### DIFF
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -50,6 +50,21 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 	k.Log("🔥 Deleting %s namespace...", k.params.TestNamespace)
 	k.client.DeleteNamespace(ctx, k.params.TestNamespace, metav1.DeleteOptions{})
 
+	// To avoid cases where test pods are stuck in terminating state because
+	// cni (cilium) pods were deleted sooner, wait until test pods are deleted
+	// before moving onto deleting cilium pods.
+	if k.params.Wait {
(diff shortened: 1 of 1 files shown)


install/uninstall.go
```diff
@@ -50,6 +50,21 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
...
+	// cni (cilium) pods were deleted sooner, wait until test pods are deleted
+	// before moving onto deleting cilium pods.
+	if k.params.Wait {
+	retryNamespace:
+		// Wait for the test namespace to be terminated. Subsequent connectivity checks would fail
+		// if the test namespace is in Terminating state.
+		_, err := k.client.GetNamespace(ctx, k.params.TestNamespace, metav1.GetOptions{})
+		if err == nil {
+			time.Sleep(defaults.WaitRetryInterval)
+			k.Log("⌛ Waiting for %s namespace to be terminated...", k.params.TestNamespace)
```
aanm, line 63: shouldn't this message be printed on line 56-57?

aanm: ah if that's the case it's all good it's just that the behavior as different before.

michi-covalent (approved):

PR #822, opened 2022-04-29
Fix Cilium CLI installation for Cilium stable branches

aanm:
read per commit basis


squeed:
@aanm can we close this :-) ?

PR #823, opened 2022-04-29
Extract all deamonsets of Cilium

ungureanuvladvictor:
In our environments we deploy multiple Cilium agent deamonsets because of memory pressure. We have hosts where the `cilium-agent` is consuming an order of magnitude compared to other low-workload hosts where `cilium-agent` is coasting related to memory usage. If we run the `sysdump` command it will collect just the 1st deamonset that it detects leaving the rest out.

I decided to write to file the full `DaemonSetList` rather than separate files per deamonset.

Signed-off-by: Vlad Ungureanu <ungureanuvladvictor@gmil.com>

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -485,20 +485,20 @@ func (c *Collector) Run() error {
 			},
 		},
 		{
-			Description: "Collecting the Cilium daemonset",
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):

ldelossa (approved):
LGTM.

PR #826, opened 2022-05-02
install: don't repeat log message when waiting for test namespace to be deleted

tklauser:
Avoid repeating the same log message upon test namespace deletion.

Noticed while debugging AKS failure on https://github.com/cilium/cilium/pull/19665

Fixes: b3d0e95f29bc ("install: Wait for test pods to be cleaned up")

### DIFF
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -55,13 +55,13 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 	// cni (cilium) pods were deleted sooner, wait until test pods are deleted
 	// before moving onto deleting cilium pods.
 	if k.params.Wait {
+		k.Log("⌛ Waiting for %s namespace to be terminated...", k.params.TestNamespace)
 	retryNamespace:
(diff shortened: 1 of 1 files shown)


tklauser:
See https://github.com/cilium/cilium/runs/6257588501?check_suite_focus=true for an example where these log messages were repeated over and over again.

aditighag:
Thanks!

> See https://github.com/cilium/cilium/runs/6257588501?check_suite_focus=true for an example where these log messages were repeated over and over again.

Does that not mean we have another flake(s) on AKS?

twpayne (approved):

tklauser:
> Thanks!
> 
> > See https://github.com/cilium/cilium/runs/6257588501?check_suite_focus=true for an example where these log messages were repeated over and over again.
> 
> Does that mean we have another flake(s) on AKS?

Possibly. I'll look into it. AKS workflow has been flaky or even broken recently, so might also be just another failure mode.

Issue #769, opened 2022-03-28
Include CNI configuration file from each node in sysdump


seanmwinn:
## Proposal / RFE
Presently, the sysdump tool does not collect /etc/cni/net.d/05-cilium.conf from each node.
This file would be useful to collect in order to validate that custom CNI configuration is working as intented.
**Is your feature request related to a problem?**

**Describe the solution you'd like**
Include /etc/cni/net.d/05-cilium.conf from each node when collecting a sysdump

Issue #766, opened 2022-03-28
Collect cni-configuration configmap with sysdump


seanmwinn:
## Proposal / RFE

Presently the configmap for cni-configuration is not included in collected sysdumps. Include this file so customers utilizing custom CNI configurations do not need to provide this data manually.

**Is your feature request related to a problem?**

**Describe the solution you'd like**

cilium sysdump includes the contents of the ConfigMap cni-configuration in kube-system namespace

PR #827, opened 2022-05-02
Add support to collect CNI config files from the system

ungureanuvladvictor:
I added 2 options to the sysdump command:
- `cni-config-directory` which defaults to `/etc/cni/net.d/`. In most of our systems we do not use this default path and thought it would be useful to allow users to specify if they have a custom path
- `cni-configmap-name` which defaults to `cni-configuration`. Opted to expose this because we allow this to be configured via Helm as well. In the code I opted to not error out and just log at debug if the configmap is not present.

Also I looked at using the backing implementation of `kubectl cp` to extract the files from the machines but it seemed more overhead to deal with tgz archives.

Fixes https://github.com/cilium/cilium-cli/issues/769
Fixes https://github.com/cilium/cilium-cli/issues/766

Signed-off-by: Vlad Ungureanu <ungureanuvladvictor@gmail.com>

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -81,6 +81,10 @@ type Options struct {
 	CiliumBugtoolFlags []string
 	// Whether to automatically detect the gops agent PID
 	DetectGopsPID bool
+	// Directory where CNI configs are located
+	CNIConfigDirectory string
+	// The name of the CNI config map
+	CNIConfigMapName string
 }
 
 // Task defines a task for the sysdump collector to execute.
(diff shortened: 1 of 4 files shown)


sysdump/sysdump.go
```diff
@@ -1038,6 +1077,39 @@ func extractGopsPID(output string) (string, error) {
...
+				if err != nil {
+					return err
+				}
+				if err := c.WriteBytes(fmt.Sprintf(cniConfigFileName, cniFileName, p.GetName()), cniConfigFileContent.Bytes()); err != nil {
+					return fmt.Errorf("failed to write CNI configuration %q for %q in namespace %q: %w", cniFileName, p.GetName(), p.Namespace, err)
+				}
+			}
+			return nil
+		}); err != nil {
+			return fmt.Errorf("failed to submit coniconflist task for %q: %w", p.Name, err)
```
tklauser, line 1107: Nit:
```suggestion
			return fmt.Errorf("failed to submit cniconflist task for %q: %w", p.Name, err)
```

ungureanuvladvictor: Added!

michi-covalent (approved):

PR #845, opened 2022-05-12
.github: explicitly set Cilium version in workflows

tklauser:
Currently, the workflows will always install the Cilium version
specified in defaults.Version which is currently set to 1.11.4.

To ensure compatibility of changes in cilium-cli we might want to run CI
with a different Cilium version, e.g. an rc release. This change allows
specifying cilium_version in the workflows to test against a different
version.

### DIFF
--- a/.github/workflows/aks.yaml
+++ b/.github/workflows/aks.yaml
@@ -29,6 +29,7 @@ env:
   name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
   location: westeurope
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
+  cilium_version: v1.11.4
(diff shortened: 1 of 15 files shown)


sayboras (approved):

.github/workflows/aks.yaml
```diff
@@ -29,6 +29,7 @@ env:
   name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
   location: westeurope
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
+  cilium_version: v1.11.4
```
nbusseneau, line 32: Do we actually need these since we are already setting the right value in `values.yaml`? I personally would say we should we do one or the other, but probably not both.

PR #847, opened 2022-05-13
go.mod, vendor: bump k8s deps from 0.24.0-alpha.0 to 0.24.0

tklauser:


tklauser:
Looks like this needs an update in `github.com/cilium/cilium/pkg/slim/k8s` first.

PR #864, opened 2022-05-20
README.md: fix binary installation instructions

tklauser:
These were accidentially included in commit 3ab53c02aa85 ("Update Go to
1.17.6") instead of being submitted in their own comment.

Fix them up so they work for any supported $GOOS/$GOARCH.


michi-covalent (approved):
i guess it's fair to assume people have go installed 👀

PR #869, opened 2022-05-23
install: change default agent health check port to avoid conflicts

jibi:
The default value 9876 for the agent health port introduced in commit
cilium/cilium@efffbdb ("daemon: expose HTTP endpoint on localhost for
health checks") conflicts with Istio's ControlZ port. The latter has
been in use for longer. Because the port is only exposed on localhost
for use by liveness/readiness probe, we can change the default value
without breaking users. Thus, change it to 9879 for which there
doesn't seem to be any documented use.

Signed-off-by: Gilberto Bertin <jibi@cilium.io>


install/install.go
```diff
@@ -241,6 +241,16 @@ var operatorClusterRole = &rbacv1.ClusterRole{
 }
 
 func (k *K8sInstaller) generateAgentDaemonSet() *appsv1.DaemonSet {
+	healthCheckPort := 9876
+
+	version, err := utils.ParseCiliumVersion(k.params.Version, k.params.BaseVersion)
+	if err != nil {
+		k.Log("Unable to parse the provided version %q, assuming %v for health checking port compatibility", k.params.Version, defaults.Version)
+	} else if version.GE(versioncheck.MustVersion("v1.10.12")) ||
+		version.GE(versioncheck.MustVersion("v1.11.6")) {
```
michi-covalent, line 250: i think v0.10 CLIs are only meant to support cilium v1.10 installation, so i think we can just do this instead:

```suggestion
	} else if version.GE(versioncheck.MustVersion("v1.10.12")) {
```

michi-covalent (approved):
lgtm one minor comment. let's wait for @tklauser to review this as well.

install/install.go
tklauser: Agree, I don't think installation for Cilium v1.11+ will work with cilium-cli v0.10.x.

tklauser (approved):
LGTM once Michi's comment re.v1.11 is addressed. Thanks!

I assume we'll need another v0.10.x release once this is merged?

tklauser:
pod-to-world tests in most workflows seem to be failing even when re-running. Looks like a Cloudflare issue as it’s also occurring on master and on cilium/cilium PRs. This particular change wouldn't affect the connectivity tests anyway, so mergeing.

PR #873, opened 2022-05-24
Add ctags/cscope target

joestringer:
I use "make tags" instinctively in Cilium repos to generate databases
for navigating the codebase. Add some simple make targets here to
facilitate this workflow.


michi-covalent (approved):
ctags huh

PR #883, opened 2022-05-31
install: Wait for test pods to be cleaned up

jibi:
[ upstream commit b3d0e95f29bca4b9a9bb11982553a988c6e5fcc4 ]

 If test pods deletion takes longer than that of (cni)
 cilium pods, test pods are stuck in the "terminating"
 state. This ends up failing CI jobs in the clean-up stage.
 Hence, wait until test pods are deleted before moving
 onto the CNI related clean-up.


jibi:
After this gets merged we'll need a new 0.10 release in order to fix the `ci-gke-1.10` test

PR #888, opened 2022-06-01
install: Add --list-versions flag

michi-covalent:
When this flag is passed to "cilium install" command, it prints out all
the available versions and exits without actually installing Cilium.

Help message:

    --list-versions                        List all the available versions without actually installing

Sample output:

    % ./cilium install --list-versions
    v1.12.0-rc2
    v1.12.0-rc1
    v1.12.0-rc0
    v1.11.5 (default)
    v1.11.4
    v1.11.3
    ...

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/internal/certs"
+	"github.com/cilium/cilium-cli/internal/helm"
 	"github.com/cilium/cilium-cli/internal/utils"
 	"github.com/cilium/cilium-cli/k8s"
 	"github.com/cilium/cilium-cli/status"
@@ -297,6 +298,9 @@ type Parameters struct {
 	// HelmValuesSecretName is the name of the secret where helm values will be
 	// stored.
 	HelmValuesSecretName string
+
+	// ListVersions lists all the available versions for install without actually installing.
+	ListVersions bool
 }
 
 type rollbackStep func(context.Context)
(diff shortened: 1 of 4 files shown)


aanm (approved):

install/install.go
```diff
@@ -559,7 +563,28 @@ func (k *K8sInstaller) restartUnmanagedPods(ctx context.Context) error {
...
+		}
+	}
+	return err
+}
+
 func (k *K8sInstaller) Install(ctx context.Context) error {
+	// If --list-versions flag is specified, print available versions and return.
+	if k.params.ListVersions {
+		return k.printVersions()
+
```
christarazi, line 586: extra newline

christarazi (approved):

install/install.go
```diff
@@ -559,7 +563,28 @@ func (k *K8sInstaller) restartUnmanagedPods(ctx context.Context) error {
...
+			fmt.Println(versions[i])
+		}
+	}
+	return err
+}
+
 func (k *K8sInstaller) Install(ctx context.Context) error {
+	// If --list-versions flag is specified, print available versions and return.
+	if k.params.ListVersions {
+		return k.printVersions()
```
twpayne, line 585: silly nit, but the flag uses "list" but the function is called "print". Maybe be consistent? silly nit though.

twpayne (approved):

tklauser (approved):
Thanks. LGTM once Chris' and Tom's review comments are addressed.

install/install.go
michi-covalent: it's good to be consistent 👍

PR #890, opened 2022-06-01
Remove --base-version flag from connectivity and hubble subcommands

michi-covalent:
might be easier to review commit by commit:

- add a wrapper function to read cilium-cli-helm-values secret and parse its data.
- remove --base-version from connectivity command.
- remove --base-version from hubble command.
- remove base version parameter from ParseCiliumVersion function.

### DIFF
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -60,18 +60,17 @@ type k8sHubbleImplementation interface {
 	GetDaemonSet(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*appsv1.DaemonSet, error)
 	CiliumStatus(ctx context.Context, namespace, pod string) (*models.StatusResponse, error)
 	ListCiliumEndpoints(ctx context.Context, namespace string, opts metav1.ListOptions) (*ciliumv2.CiliumEndpointList, error)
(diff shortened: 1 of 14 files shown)


sayboras (approved):
Looks legit :100: :heavy_check_mark:

hubble/hubble.go
```diff
@@ -60,18 +60,18 @@ type k8sHubbleImplementation interface {
...
 
 type K8sHubble struct {
-	client              k8sHubbleImplementation
-	params              Parameters
-	certManager         *certs.CertManager
-	ciliumVersion       string
-	manifests           map[string]string
-	semVerCiliumVersion semver.Version
-	helmYAMLValues      string
+	ctx            context.Context
```
tklauser, line 68: I think we shouldn't include the context inside the struct but rather pass it to the functions that need it. See https://pkg.go.dev/context which states:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

Also see https://go.dev/blog/context-and-structs

tklauser (changes requested):

tklauser (approved):

PR #895, opened 2022-06-03
gha: Couple of changes to fix EKS failures

sayboras:
### Description

Fixing issue with invalid apiVersion

```
Error: Kubernetes cluster unreachable: exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"
```

### Testing

Testing was done with temp commit, successful runs are available below

CNI: https://github.com/cilium/cilium-cli/runs/6726007799?check_suite_focus=true
Tunnel: https://github.com/cilium/cilium-cli/runs/6726008680?check_suite_focus=true

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -54,6 +54,14 @@ jobs:
           sudo tar xzvfC eksctl_$(uname -s)_amd64.tar.gz /usr/bin
           rm eksctl_$(uname -s)_amd64.tar.gz
 
+      - name: Install helm
+        uses: azure/setup-helm@18bc76811624f360dbd7f18c2d4ecb32c7b87bab
+        with:
+          # Due to the below issue, v3.8.2 is pinned currently to avoid
+          # exec plugin: invalid apiVersion "client.authentication.k8s.io/v1alpha1"
+          # https://github.com/helm/helm/issues/10975
+          version: v3.8.2
+
       - name: Set up AWS CLI credentials
         uses: aws-actions/configure-aws-credentials@05b148adc31e091bafbaf404f745055d4d3bc9d2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/eks-tunnel.yaml
```diff
@@ -54,6 +54,11 @@ jobs:
           sudo tar xzvfC eksctl_$(uname -s)_amd64.tar.gz /usr/bin
           rm eksctl_$(uname -s)_amd64.tar.gz
 
+      - name: Install helm
+        uses: azure/setup-helm@18bc76811624f360dbd7f18c2d4ecb32c7b87bab
+        with:
+          version: v3.8.2
+
```
nbusseneau, line 61: I'm curious: why is that needed? Helm is already pre-installed on GH Runners: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

nbusseneau: Scratch that, I just saw the commit description. In this case I would suggest

sayboras: Yup, however there is issue with latest helm version e.g. 3.9.0, so we need to use lower version.

Also, avoiding version drift is not a bad thing in general.

https://github.com/helm/helm/issues/10975

.github/workflows/eks-tunnel.yaml
```diff
@@ -54,6 +54,11 @@ jobs:
           sudo tar xzvfC eksctl_$(uname -s)_amd64.tar.gz /usr/bin
           rm eksctl_$(uname -s)_amd64.tar.gz
 
+      - name: Install helm
+        uses: azure/setup-helm@18bc76811624f360dbd7f18c2d4ecb32c7b87bab
+        with:
+          version: v3.8.2
```
tklauser, line 60: Maybe add a comment here linking to https://github.com/helm/helm/issues/10975 (and optionally a newly created cilium-cli  issue), explaining why we pin to this version specifically. This will hopefully remind us to check again after some time whether the issue is fixed in newer helm versions.

michi-covalent (approved):
lgtm. probably better to do `Install helm` thingy for all the workflows to be consistent.

PR #898, opened 2022-06-03
install: make `datapath-mode` the main flag, deprecate `ipam`

nbusseneau:
Since we are now using the Helm charts as basis to generate manifests for `cilium install`, most of the configuration flags we used to have are superseded by Helm options that can be set via `helm-set` flags.

This commit aims:
- To make `datapath-mode` the only remaining "main configuration flag", so that user experience is kept simple.
- To rework the auto-detection logic in a way that makes `datapath-mode` the only remaining "main configuration flag", so that the Helm chart configuration code is kept simple and centralized in `helm.go`.

In doing so, we are able to mark deprecated some of the older configuration flags that are superseded by Helm options.


michi-covalent (approved):
the code change looks good, but i don't have enough context for all he IPAM mode settings. requesting review from cilium/ipam.

nbusseneau:
> the code change looks good, but i don't have enough context for all he IPAM mode settings. requesting review from cilium/ipam.

IPAM changes should be transparent, in terms of code the goal is to have it work exactly as before except that the IPAM is auto-detected alongside `datapath-mode` rather than on its own.

gandro (approved):
I think I agree that deprecating the IPAM flag is the right way forward. It rarely can be used on it's own and almost always also requires other options to be set as well.

While I know that this has already been discussed, seeing it implemented I find the current `datapath-mode` flag a bit confusing. There is e.g. no way to set up direct routing mode (which I would argue is a common datapath mode), and it also doesn't allow the user to override the flavor to e.g. use `kind`.

Maybe, exposing the `k.flavor.Kind` flag as `--flavor` somehow, and deprecating both `--ipam` and `--datapath-mode` in favor of Helm flags might be the better option. This would also allow us to provide sane defaults for commonly used flavors like `gke`, `eni`, `azure`, but also other platforms such as `kind`, `microk8s` and `rancher`, and in the future maybe even `aws-cni-chaining`, `mke` etc. If anyone needs to set up a the datapath in a more advanced fashion (e.g. using direct routing), they anyway need to use Helm instead, so the `--datapath-mode` flag is not helpful there anyways.

In any case, that's just my impression from seeing this implemented. Not blocking the PR however, since this has already been discussed.

tklauser:
Looks like this PR picked up a merge conflict and needs to be rebased.

tklauser (approved):
I agree with [Sebastian's comment](https://github.com/cilium/cilium-cli/pull/898#pullrequestreview-997860297) re. exposing providing a `--flavor` flag and deprecating `--datapath-mode`. This PR is a step in the right direction and I think we can follow-up with further steps once it's merged. I'll create an issue for further discussion re. the `flavor` flag.

nbusseneau:
@gandro @tklauser I agree. If that's OK with you I can quickly rework the PR to that end, so that we don't have multiple hops in terms of deprecation (bad from a user perspective, IMO).

tklauser:
> @gandro @tklauser I agree. If that's OK with you I can quickly rework the PR to that end, so that we don't have multiple hops in terms of deprecation (bad from a user perspective, IMO).

That sounds good to me, thanks Nicolas!

gandro (approved):

nbusseneau:
The external workloads failure seems completely unrelated to the PR.

tklauser:
> The external workloads failure at https://github.com/cilium/cilium-cli/actions/runs/2468063257 seems completely unrelated to the PR.

Yes, it seems to consistently fail on `master` as well. AFAIK @michi-covalent has been investigating said failure in #908. Let's fix these separately.

Merging this PR to unblock AKS re-enabling.

PR #913, opened 2022-06-11
fix(helm): support setting string slices

ernado:


ernado:
Actually I've decided to just use helm directly.

PR #924, opened 2022-06-15
ci: re-add AKS (Azure IPAM) workflow

nbusseneau:
This is almost the same as the one that was removed in ba5ecd6e92d1d8a127afaf5beeebbab4a47280ca, except that we have removed taints as they are not working due to an AKS-side change.

Taint removal is sub-optimal from a user experience standpoint because one might get nodes with pods that are not properly handled by Cilium until they are restarted, but in CI it basically makes no difference and at least allows us to verify that the Azure IPAM is working as intended.


pchaigno:
> one might get nodes with pods that are not properly handled by Cilium until they are restarted, but in CI it basically makes no difference

Why can't that cause issues in CI? If a pod is not managed by Cilium we may fail to enforce policies to/from it.

pchaigno (approved):

PR #926, opened 2022-06-15
Use OCI container reference parser to extract Cilium's version

YutaroHayakawa:
The current mechanism to detect the image version of the Cilium has a
problem that it misinterpret the image name with registry's port number
(e.g. localhost:5000/cilium/cilium:latest). This is because it splits
the image string with strings.SplitN(s, 2) which result is wrongly
splitted slice ["localhost", "5000/cilium/cilium:latest"]. The version
estimation fails in this case.

Obviously, writing our own OCI image parser is painful. Instead, we can
rely on the official parser. In this commit, we rewrite the image
version estimation with the parser in
github.com/distribution/distribution. It also adds unit test to make
sure some major cases work.

Signed-off-by: Yutaro Hayakawa <yutaro.hayakawa@isovalent.com>

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
+	"github.com/distribution/distribution/reference"
 	"helm.sh/helm/v3/pkg/chartutil"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


rolinh (approved):

k8s/client_test.go
```diff
@@ -0,0 +1,32 @@
...
+		name            string
+		image           string
+		expectedVersion string
+	}{
+		{"registry name and tag", "quay.io/cilium/cilium:latest", "latest"},
+		{"registry name and tag with digest", "quay.io/cilium/cilium:latest@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2", "latest"},
+		{"no registry name and tag", "cilium/cilium:latest", "latest"},
+		{"no registry name and tag with digest", "cilium/cilium:latest@sha256:45b23dee08af5e43a7fea6c4cf9c25ccf269ee113168c19722f87876677c5cb2", "latest"},
+		{"no tag", "quay.io/cilium/cilium", "latest"},
+		{"with hyphen", "quay.io/cilium/cilium-ci", "-ci:latest"},
```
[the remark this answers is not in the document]
YutaroHayakawa: Thanks! Applied!

tklauser:
@YutaroHayakawa I think you need to rebase the PR branch on top of latest `master` to fix the failing workflows (i.e. pick up #930).

k8s/client.go
```diff
@@ -723,37 +724,57 @@ func (c *Client) GetLogs(ctx context.Context, namespace, name, container string,
...
-				// skip non cilium images
-				continue
-			}
-			// default to "latest" as k8s may not include it explicitly
-			version := "latest"
-			if len(image) > 2 {
-				// skip image with too many colons
+			version, err := getCiliumVersionFromImage(container.Image)
+			if err != nil {
 				continue
```
tommyp1ckles, line 777: Should we add some debug logging for this error?

YutaroHayakawa: Oh, it is merged before I reply to this, sorry. Yeah, I think we should. I'll create a separate PR.

PR #945, opened 2022-06-22
make: run release build as regular user

tklauser:
Instead of adding the checkout directory to git's safe directories as
done in commit 023ccfb8466e ("make: allow running `git status` during
release build"), change the release target to create a regular user with
the same UID/GID as the current user (assumed to be the user owning the
working directory) inside the container and use it to run `make
release`.

Follows cilium/hubble#751

Also bump the alpine version used for `make release` to 3.16.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ release:
 	docker run \
 		--rm \
 		--workdir /cilium \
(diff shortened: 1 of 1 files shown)


gandro (approved):
Looks good to me!

PR #952, opened 2022-07-01
hubble metric

### LINKED ISSUE #947: install command doesn't create metric services
i installed cilium with this values file:
```
hubble:
  metrics:
    enabled:
    - flow
prometheus:
  enabled: true
```

`cilium install --version v1.12.0-rc3 --helm-values ./values.yaml` doesn't create `cilium-agent` / `hubble-metrics` services.

Akshit42-hue:
Signed-off-by: Akshit42-hue <patelakshit2025@gmail.com>
Fixes #947

### DIFF
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -178,7 +178,7 @@ func (k *K8sHubble) Log(format string, a ...interface{}) {
 	fmt.Fprintf(k.params.Writer, format+"\n", a...)
 }
 
-func (k *K8sHubble) () *corev1.Service {
+func (k *K8sHubble) generatePeerService() *corev1.Service {
 	var (
 		svcFilename string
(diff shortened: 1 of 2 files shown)


Akshit42-hue:
Hey @aanm, please do have a look :)

aanm (changes requested):
What I meant in the GH issue was that we need to generate the manifests for the hubble metrics services that are missing. Similar to what is done [here](https://github.com/cilium/cilium-cli/blob/1e1bf42ba00e58f681eddb92f752fbc81d5df885/hubble/hubble.go#L275-L278) but not exactly the same function.

tklauser:
The corresponding issue was fixed in #955. Thanks for your contribution.

PR #958, opened 2022-07-06
Add JSON status (fix #956)

### LINKED ISSUE #956: JSON output for status
## Proposal / RFE

It would be useful if the output of the status commands could be easily parsed.

For example, I'm using `cilium clustermesh status` to check if clusters are meshed with one another, something like:

```
cilium clustermesh status --context cluster1 | grep -q cluster2 && echo "meshed"
```

but it'd be more precise to get a `-o json` or `-o jsonpath` and filter on a specific field of the output.

raphink:
- feat: add json output for clustermesh status
- feat: add json output for status
- feat: downcase json export fields for status
- feat: downcase json export fields for clustermesh status

Fixes: #956

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -468,6 +469,7 @@ type Parameters struct {
 	ConfigOverwrites     []string
 	Retries              int
 	HelmValuesSecretName string
+	Output               string
 }
 
 func (p Parameters) validateParams() error {
@@ -629,16 +631,16 @@ func (k *K8sClusterMesh) Enable(ctx context.Context) error {
 }
 
 type accessInformation struct {
-	ServiceIPs           []string
-	ServicePort          int
-	ClusterID            string
-	ClusterName          string
-	CA                   []byte
-	ClientCert           []byte
-	ClientKey            []byte
-	ExternalWorkloadCert []byte
-	ExternalWorkloadKey  []byte
-	Tunnel               string
+	ServiceIPs           []string `json:"service_ips,omitempty"`
+	ServicePort          int      `json:"service_port,omitempty"`
+	ClusterID            string   `json:"cluster_id,omitempty"`
+	ClusterName          string   `json:"cluster_name,omitempty"`
+	CA                   []byte   `json:"ca,omitempty"`
+	ClientCert           []byte   `json:"client_cert,omitempty"`
+	ClientKey            []byte   `json:"client_key,omitempty"`
+	ExternalWorkloadCert []byte   `json:"external_workload_cert,omitempty"`
+	ExternalWorkloadKey  []byte   `json:"external_workload_key,omitempty"`
+	Tunnel               string   `json:"tunnel,omitempty"`
 }
 
 func (ai *accessInformation) etcdConfiguration() string {
@@ -1042,9 +1044,9 @@ func (k *K8sClusterMesh) Disconnect(ctx context.Context) error {
 }
 
 type Status struct {
-	AccessInformation *accessInformation
-	Service           *corev1.Service
-	Connectivity      *ConnectivityStatus
+	AccessInformation *accessInformation  `json:"access_information,omitempty"`
+	Service           *corev1.Service     `json:"service,omitempty"`
+	Connectivity      *ConnectivityStatus `json:"connectivity,omitempty"`
 }
 
 func (k *K8sClusterMesh) statusAccessInformation(ctx context.Context, log bool, getExternalWorkloadSecret bool) (*accessInformation, error) {
@@ -1104,23 +1106,23 @@ func (k *K8sClusterMesh) waitForDeployment(ctx context.Context) error {
 }
 
 type StatisticalStatus struct {
-	Min int64
-	Avg float64
-	Max int64
+	Min int64   `json:"min,omitempty"`
+	Avg float64 `json:"avg,omitempty"`
+	Max int64   `json:"max,omitempty"`
 }
 
 type ClusterStats struct {
-	Configured int
-	Connected  int
+	Configured int `json:"configured,omitempty"`
+	Connected  int `json:"connected,omitempty"`
 }
 
 type ConnectivityStatus struct {
-	GlobalServices StatisticalStatus
-	Connected      StatisticalStatus
-	Clusters       map[string]*ClusterStats
-	Total          int64
-	NotReady       int64
-	Errors         status.ErrorCountMapMap
+	GlobalServices StatisticalStatus        `json:"global_services,omitempty"`
+	Connected      StatisticalStatus        `json:"connected,omitempty"`
+	Clusters       map[string]*ClusterStats `json:"clusters,omitempty"`
+	Total          int64                    `json:"total,omitempty"`
+	NotReady       int64                    `json:"not_ready,omitempty"`
+	Errors         status.ErrorCountMapMap  `json:"errors,omitempty"`
 }
 
 func (c *ConnectivityStatus) addError(pod, cluster string, err error) {
@@ -1295,6 +1297,15 @@ func (k *K8sClusterMesh) Status(ctx context.Context) (*Status, error) {
 	}
 
 	s.Connectivity, err = k.statusConnectivity(ctx)
+
+	if k.params.Output == status.OutputJSON {
+		jsonStatus, err := json.MarshalIndent(s, "", " ")
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal status to JSON")
+		}
+		fmt.Println(string(jsonStatus))
+		return s, nil
+	}
 
 	if s.Connectivity != nil {
 		if s.Connectivity.NotReady > 0 {
--- a/internal/cli/cmd/clustermesh.go
+++ b/internal/cli/cmd/clustermesh.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/cilium/cilium-cli/clustermesh"
 	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/status"
 )
 
 func newCmdClusterMesh() *cobra.Command {
@@ -152,6 +153,12 @@ func newCmdClusterMeshStatus() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			params.Namespace = namespace
 
+			if params.Output == status.OutputJSON {
+				// Write status log messages to stderr to make sure they don't
+				// clutter JSON output.
+				params.Writer = os.Stderr
+			}
+
 			cm := clustermesh.NewK8sClusterMesh(k8sClient, params)
 			if _, err := cm.Status(context.Background()); err != nil {
 				fatalf("Unable to determine status:  %s", err)
@@ -163,6 +170,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait until status is successful")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
 	cmd.Flags().BoolVar(&params.SkipServiceCheck, "skip-service-check", false, "Do not require service IP of remote cluster to be available")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", status.OutputSummary, "Output format. One of: json, summary")
 
 	return cmd
 }
--- a/internal/cli/cmd/status.go
+++ b/internal/cli/cmd/status.go
@@ -5,7 +5,9 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
@@ -29,10 +31,21 @@ func newCmdStatus() *cobra.Command {
 			}
 
 			s, err := collector.Status(context.Background())
-			// Report the most recent status even if an error occurred.
-			fmt.Print(s.Format())
 			if err != nil {
+				// Report the most recent status even if an error occurred.
+				fmt.Fprint(os.Stderr, s.Format())
 				fatalf("Unable to determine status:  %s", err)
+			}
+			if params.Output == status.OutputJSON {
+				jsonStatus, err := json.MarshalIndent(s, "", " ")
+				if err != nil {
+					// Report the most recent status even if an error occurred.
+					fmt.Fprint(os.Stderr, s.Format())
+					fatalf("Unable to marshal status to JSON:  %s", err)
+				}
+				fmt.Println(string(jsonStatus))
+			} else {
+				fmt.Print(s.Format())
 			}
 			return err
 		},
@@ -43,6 +56,7 @@ func newCmdStatus() *cobra.Command {
 	cmd.Flags().IntVar(&params.WorkerCount,
 		"worker-count", status.DefaultWorkerCount,
 		"The number of workers to use")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", status.OutputSummary, "Output format. One of: json, summary")
 
 	return cmd
 }
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -44,6 +44,9 @@ type K8sStatusParameters struct {
 
 	// The number of workers to use.
 	WorkerCount int
+
+	// The output format
+	Output string
 }
 
 type K8sStatusCollector struct {
--- a/status/status.go
+++ b/status/status.go
@@ -26,6 +26,11 @@ const (
 	Reset   = "\033[0m"
 )
 
+const (
+	OutputJSON    = "json"
+	OutputSummary = "summary"
+)
+
 // MapCount is a map to count number of occurrences of a string
 type MapCount map[string]int
 
@@ -35,10 +40,10 @@ type MapMapCount map[string]MapCount
 // PodStateCount counts the number of pods in the k8s cluster
 type PodsCount struct {
 	// All is the number of all pods in the k8s cluster
-	All int
+	All int `json:"all"`
 
 	// ByCilium is the number of all the pods in the k8s cluster
-	ByCilium int
+	ByCilium int `json:"by_cilium"`
 }
 
 // PodStateCount counts the number of pods in a particular state
@@ -77,29 +82,29 @@ type ErrorCountMapMap map[string]ErrorCountMap
 type Status struct {
 	// ImageCount is a map counting the number of images in use indexed by
 	// the image name
-	ImageCount MapMapCount
+	ImageCount MapMapCount `json:"image_count,omitempty"`
 
 	// PhaseCount is a map counting the number of pods in each phase
 	// (running, failing, scheduled, ...)
-	PhaseCount MapMapCount
+	PhaseCount MapMapCount `json:"phase_count,omitempty"`
 
 	// PodState counts the number of pods matching conditions such as
 	// desired, ready, available, and unavailable
-	PodState PodStateMap
+	PodState PodStateMap `json:"pod_state,omitempty"`
 
 	// PodsCount is the number of pods in the k8s cluster
 	// all pods, and pods managed by cilium
-	PodsCount PodsCount
+	PodsCount PodsCount `json:"pods_count,omitempty"`
 
-	CiliumStatus CiliumStatusMap
+	CiliumStatus CiliumStatusMap `json:"cilium_status,omitempty"`
 
 	// Errors is the aggregated errors and warnings of all pods of a
 	// particular deployment type
-	Errors ErrorCountMapMap
+	Errors ErrorCountMapMap `json:"errors,omitempty"`
 
 	// CollectionErrors is the errors that accumulated while collecting the
 	// status
-	CollectionErrors []error
+	CollectionErrors []error `json:"collection_errors,omitempty"`
 
 	mutex *sync.Mutex
 }


status/status.go
```diff
@@ -77,29 +77,29 @@ type ErrorCountMapMap map[string]ErrorCountMap
...
 
 	// Errors is the aggregated errors and warnings of all pods of a
 	// particular deployment type
-	Errors ErrorCountMapMap
+	Errors ErrorCountMapMap `json:"errors"`
 
 	// CollectionErrors is the errors that accumulated while collecting the
 	// status
-	CollectionErrors []error
+	CollectionErrors []error `json:"collection_errors"`
```
aditighag, line 102: Since you've added tags, how about skipping null fields with `omitempty` attribute?

aditighag (changes requested):

internal/cli/cmd/clustermesh.go
```diff
@@ -162,6 +162,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait until status is successful")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
 	cmd.Flags().BoolVar(&params.SkipServiceCheck, "skip-service-check", false, "Do not require service IP of remote cluster to be available")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "compact", "Output format. One of: (json, compact).")
```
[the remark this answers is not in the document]
raphink: That was just to give a name to the default. Should I name it `summary` instead, or just use an empty string?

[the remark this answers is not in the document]
aditighag: I meant that the default option is compact/summary so users needn't specify it by passing `-o compact`. Users will use `-o` only when they want to print the output in the json format.

status/status.go
aditighag: Only for those fields that are not populated by default, or can show empty values like {} (e.g., list).

internal/cli/cmd/clustermesh.go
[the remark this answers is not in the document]
raphink: OK, so then I'd check against the empty string for the default value?

status/status.go
raphink: Right, so no `int` and other mandatory values then.

internal/cli/cmd/status.go
```diff
@@ -30,7 +31,15 @@ func newCmdStatus() *cobra.Command {
 
 			s, err := collector.Status(context.Background())
 			// Report the most recent status even if an error occurred.
-			fmt.Print(s.Format())
+			if params.Output == "json" {
+				jsonStatus, err := json.MarshalIndent(s, "", " ")
+				if err != nil {
+					fmt.Printf("Unable to marshal status to JSON:  %s", err)
```
squeed, line 37: Should probably be `fatalf` here.

internal/cli/cmd/clustermesh.go
```diff
@@ -162,6 +162,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait until status is successful")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
 	cmd.Flags().BoolVar(&params.SkipServiceCheck, "skip-service-check", false, "Do not require service IP of remote cluster to be available")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: (json, summary).")
```
aditighag, line 165: Nit: As mentioned before, users need not specify "summary" output as you have already set it as default. Take a look at the above flag if you want a reference - 

`cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")`

aditighag (commented):
Looks good! Thanks for the revisions. Can you please squash the commits where you renamed compact -> summary, and the commit with `omitempty` additions?

I didn't take a look at the clustermesh changes closely as Tam is already in the reviewer's list.

internal/cli/cmd/clustermesh.go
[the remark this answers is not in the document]
sayboras: IMO, the default value is automatically added in help command by cobra, which makes it clear enough to user to know all applicable values, and they can also omit specifying default value (e.g. summary).

```
  -o, --output string            Output format. One of: json, summary (default). (default "summary")
```

clustermesh/clustermesh.go
```diff
@@ -990,9 +994,9 @@ func (k *K8sClusterMesh) Disconnect(ctx context.Context) error {
 }
 
 type Status struct {
-	AccessInformation *accessInformation
-	Service           *corev1.Service
-	Connectivity      *ConnectivityStatus
+	AccessInformation *accessInformation  `json:"access_information,omitempty"`
+	Service           *corev1.Service     `json:"service,omitempty"`
```
sayboras, line 998: corev1.Service is kind of verbose with all the fields such as managedFields, etc, so normally, we are using cilium slim type instead.

Please find below the patch if you want to address this point

<details>
<summary>diff</summary>

```diff
diff --git a/clustermesh/clustermesh.go b/clustermesh/clustermesh.go
index 63c70380..b2d6f8ba 100644
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -25,7 +25,9 @@ import (
    "k8s.io/apimachinery/pkg/util/intstr"
 
    "github.com/cilium/cilium/api/v1/models"
+   ciliumK8s "github.com/cilium/cilium/pkg/k8s"
    ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+   slim_corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
 
    "github.com/cilium/cilium-cli/defaults"
    "github.com/cilium/cilium-cli/internal/certs"
@@ -1037,9 +1039,9 @@ func (k *K8sClusterMesh) Disconnect(ctx context.Context) error {
 }
 
 type Status struct {
-   AccessInformation *accessInformation  `json:"access_information,omitempty"`
-   Service           *corev1.Service     `json:"service,omitempty"`
-   Connectivity      *ConnectivityStatus `json:"connectivity,omitempty"`
+   AccessInformation *accessInformation   `json:"access_information,omitempty"`
+   Service           *slim_corev1.Service `json:"service,omitempty"`
+   Connectivity      *ConnectivityStatus  `json:"connectivity,omitempty"`
 }
 
 func (k *K8sClusterMesh) statusAccessInformation(ctx context.Context, log bool, getExternalWorkloadSecret bool) (*accessInformation, error) {
@@ -1062,7 +1064,7 @@ retry:
    return ai, err
 }
 
-func (k *K8sClusterMesh) statusService(ctx context.Context) (*corev1.Service, error) {
+func (k *K8sClusterMesh) statusService(ctx context.Context) (*slim_corev1.Service, error) {
    w := utils.NewWaitObserver(ctx, utils.WaitParameters{Log: func(err error, wait string) {
        k.Log("⌛ Waiting (%s) for ClusterMesh service to be available: %s", wait, err)
    }})
@@ -1081,7 +1083,12 @@ retry:
        return nil, fmt.Errorf("clustermesh-apiserver cannot be found: %w", err)
    }
 
-   return svc, nil
+   r, ok := ciliumK8s.ConvertToK8sService(svc).(*slim_corev1.Service)
+   if !ok {
+       return nil, fmt.Errorf("unable to convert service %q to slim_corev1.Service", svc.Name)
+   }
+
+   return r, nil
 }
 
 func (k *K8sClusterMesh) waitForDeployment(ctx context.Context) error {
@@ -1275,7 +1282,7 @@ func (k *K8sClusterMesh) Status(ctx context.Context) (*Status, error) {
 
    k.Log("✅ Service %q of type %q found", defaults.ClusterMeshServiceName, s.Service.Spec.Type)
 
-   if s.Service.Spec.Type == corev1.ServiceTypeLoadBalancer {
+   if s.Service.Spec.Type == slim_corev1.ServiceTypeLoadBalancer {
        if len(s.AccessInformation.ServiceIPs) == 0 {
            k.Log("❌ Service is of type LoadBalancer but has no IPs assigned")
            return nil, fmt.Errorf("no IP available to reach cluster")
@@ -1687,7 +1694,7 @@ func (k *K8sClusterMesh) ExternalWorkloadStatus(ctx context.Context, names []str
 
    k.Log("✅ Service %q of type %q found", defaults.ClusterMeshServiceName, svc.Spec.Type)
 
-   if svc.Spec.Type == corev1.ServiceTypeLoadBalancer {
+   if svc.Spec.Type == slim_corev1.ServiceTypeLoadBalancer {
        if len(ai.ServiceIPs) == 0 {
            k.Log("❌ Service is of type LoadBalancer but has no IPs assigned")
            return fmt.Errorf("no IP available to reach cluster")
```

</details>

sayboras (approved):
I have couple of small comments as per below, but the changes are LGTM in general :100:

clustermesh/clustermesh.go
raphink: I'm fine with using `slim_core` @sayboras . That means we need to bump the `cilium/cilium` vendoring though. Should I add this to this PR?

internal/cli/cmd/clustermesh.go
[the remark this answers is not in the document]
raphink: OK. Sorry, it took me a while to understand. I'm guessing what you and @aditighag mean is:

```go
cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json.")
```

since `summary` will be mentioned anyway, right?

ldelossa (approved):

raphink:
Rebased

raphink:
I've squashed a few commits. I don't know if you'd like me to squash more than that.

raphink:
I don't understand why the cross-build is failing 🤔

tklauser:
> I don't understand why the cross-build is failing thinking

It looks like this is an issue with the newly vendored Cilium version:

> 2022-11-17T13:14:33.1736986Z # github.com/cilium/cilium/pkg/mtu
2022-11-17T13:14:33.1847236Z ##[error]vendor/github.com/cilium/cilium/pkg/mtu/mtu.go:111:15: undefined: getMTUFromIf
2022-11-17T13:14:33.1860112Z ##[error]vendor/github.com/cilium/cilium/pkg/mtu/mtu.go:113:15: undefined: autoDetect
2022-11-17T13:14:33.2627923Z # github.com/cilium/cilium/pkg/node
2022-11-17T13:14:33.2629974Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:96:14: undefined: firstGlobalV4Addr
2022-11-17T13:14:33.2631853Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:131:27: undefined: firstGlobalV6Addr
2022-11-17T13:14:33.2633262Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:166:23: undefined: firstGlobalV4Addr
2022-11-17T13:14:33.2634706Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:176:16: undefined: firstGlobalV4Addr
2022-11-17T13:14:33.2636663Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:187:23: undefined: firstGlobalV6Addr
2022-11-17T13:14:33.2638502Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:197:16: undefined: firstGlobalV6Addr
2022-11-17T13:14:33.2639815Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:218:15: undefined: firstGlobalV4Addr
2022-11-17T13:14:33.2641201Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:229:15: undefined: firstGlobalV4Addr
2022-11-17T13:14:33.2642740Z ##[error]vendor/github.com/cilium/cilium/pkg/node/address.go:742:9: undefined: getCiliumHostIPsFromNetDev
2022-11-17T13:14:57.0224362Z ##[error]Process completed with exit code 2.

Unfortunately, we currently don't check upstream whether the packages imported by cilium-cli are cross-buildable, also see https://github.com/cilium/cilium/issues/16843

raphink:
@tklauser I rebased but cross compile is still failing 😢

tklauser:
> @tklauser I rebased but cross compile is still failing cry

The fix was not yet backported, we'll need to wait until the respective backport PR (https://github.com/cilium/cilium/pull/22308) is merged and then update the vendored version in this PR to the latest `v1.12` branch.

raphink:
OK. Sorry for the noise, I thought it was merged already.

clustermesh/clustermesh.go
```diff
@@ -1294,6 +1305,15 @@ func (k *K8sClusterMesh) Status(ctx context.Context) (*Status, error) {
 
 	s.Connectivity, err = k.statusConnectivity(ctx)
 
+	if k.params.Output == "json" {
+		jsonStatus, err := json.MarshalIndent(s, "", " ")
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal status to JSON")
+		}
+		fmt.Println(string(jsonStatus))
+		return nil, err
```
tklauser, line 1314: I think this should be

```suggestion
		return s, nil
```

for consistency.

internal/cli/cmd/clustermesh.go
```diff
@@ -163,6 +163,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait until status is successful")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
 	cmd.Flags().BoolVar(&params.SkipServiceCheck, "skip-service-check", false, "Do not require service IP of remote cluster to be available")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json")
```
tklauser, line 166:
```suggestion
	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json, summary")
```

internal/cli/cmd/status.go
```diff
@@ -30,10 +31,18 @@ func newCmdStatus() *cobra.Command {
 
 			s, err := collector.Status(context.Background())
 			// Report the most recent status even if an error occurred.
-			fmt.Print(s.Format())
```
tklauser, line 33: As the comment indicates, we still want to print the status in case of an error. At least in case of `params.Output != "json"`.

internal/cli/cmd/status.go
```diff
@@ -43,6 +52,7 @@ func newCmdStatus() *cobra.Command {
 	cmd.Flags().IntVar(&params.WorkerCount,
 		"worker-count", status.DefaultWorkerCount,
 		"The number of workers to use")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json")
```
tklauser, line 55:
```suggestion
	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json, summary")
```

status/k8s.go
```diff
@@ -44,6 +44,9 @@ type K8sStatusParameters struct {
 
 	// The number of workers to use.
 	WorkerCount int
+
+	// The output format
+	Output string
```
tklauser, line 49: Would be nice to introduce a dedicated type for this field that will only take valid values `json` and `summary`.

tklauser (changes requested):
Looks good overall, a few minor remarks inline that should be addressed before merge.

Please also squash all the commits into one for the final version.

internal/cli/cmd/clustermesh.go
tklauser: I just notice the previous discussion about this: https://github.com/cilium/cilium-cli/pull/958#discussion_r962268710

I don't think the default value is added to the list automatically as Tam suggested. There is additional text added to say what the default value is, though:

```
  -o, --output string            Output format. One of: json (default "summary")
```

But I think that `summary` should still be included in the "One of" list, as we do for similar flags in the per-node `cilium` binary.

internal/cli/cmd/status.go
raphink: It's still done on line 37 (in the `else` statement).

clustermesh/clustermesh.go
```diff
@@ -466,6 +469,7 @@ type Parameters struct {
 	ConfigOverwrites     []string
 	Retries              int
 	HelmValuesSecretName string
+	Output               string
```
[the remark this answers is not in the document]
raphink: Do you mean an enum type like this?

```go
type ParameterOutput int

const (
   Json ParameterOutput = iota
  Summary
)
```

internal/cli/cmd/status.go
tklauser: But the `if err != nil { fatalf() }` on [line 34](https://github.com/cilium/cilium-cli/pull/958/files#diff-20d430f1433a9b02bb41f6cd3db12b6133f75c7ba6a7ab4030d78a6bc3222fa1L34) will cause the `else` to never be reached in case of an error. I think we need to print the status here.

clustermesh/clustermesh.go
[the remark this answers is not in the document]
tklauser: Yeah, something like that. I didn't consider the implications wrt. `StringVarP`. If this turns out too cumbersome to change, then let's leave it as is.

internal/cli/cmd/status.go
raphink: Oh I see. I hadn't considered the error.

raphink: Just pushed a change which should address it.

raphink:
I think I've addressed everything (I've given up on the enum because it clashed with the option declaration).

internal/cli/cmd/status.go
tklauser: I think this is still not quite right. The change added reporting of the status in the `if params.Output == "json"` branch below, however we want to report it in case of `params.Output == "summary"`. I think something like the following should work:

```go
s, err := collector.Status(context.Background())
if params.Output == "summary" {
	// Report the most recent status even if an error occurred.
	fmt.Print(s.Format())
}
if err != nil {
	fatalf("Unable to determine status:  %s", err)
}
if params.Output == "json" {
	jsonStatus, err := json.MarshalIndent(s, "", " ")
	if err != nil {
		fmt.Print(s.Format())
		fatalf("Unable to marshal status to JSON:  %s", err)
	}
	fmt.Println(string(jsonStatus))
} 
```

(and drop the `else` branch below).

clustermesh/clustermesh.go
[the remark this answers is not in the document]
tklauser: Thinking about it again, could we at least define the constants as plain strings and use them rather than repeating `"json"` and `"summary"` several times, i.e.

```go
const (
  OutputJSON = "json"
  OutputSummary = "summary"
)

File: internal/cli/cmd/status.go (hunk 8 above)
Comment (raphink) [reply internal/cli/cmd/status.go:33]: OK. That will not report status in case of error when output is set to json though.

Comment (raphink) [reply internal/cli/cmd/status.go:33]: I believe my last push fixes this.

File: internal/cli/cmd/clustermesh.go [hunk 12]
```diff
@@ -163,6 +163,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait until status is successful")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", 15*time.Minute, "Maximum time to wait")
 	cmd.Flags().BoolVar(&params.SkipServiceCheck, "skip-service-check", false, "Do not require service IP of remote cluster to be available")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json, summary")
```
tklauser, line 166:
```suggestion
	cmd.Flags().StringVarP(&params.Output, "output", "o", status.OutputSummary, "Output format. One of: json, summary")
```

clustermesh/clustermesh.go
```diff
@@ -466,8 +469,14 @@ type Parameters struct {
...
 }
 
+const (
+	OutputJSON    = "json"
+	OutputSummary = "summary"
+)
```
tklauser, line 478: Let's define these in the `status` package instead, so we can use them here and in `internal/cli/cmd`.

internal/cli/cmd/status.go
```diff
@@ -29,11 +30,22 @@ func newCmdStatus() *cobra.Command {
...
 
 			s, err := collector.Status(context.Background())
-			// Report the most recent status even if an error occurred.
-			fmt.Print(s.Format())
 			if err != nil {
+				// Report the most recent status even if an error occurred.
+				fmt.Print(s.Format())
 				fatalf("Unable to determine status:  %s", err)
 			}
+			if params.Output == "json" {
```
tklauser, line 38:
```suggestion
			if params.Output == status.OutputJSON {
```

internal/cli/cmd/status.go
```diff
@@ -43,6 +55,7 @@ func newCmdStatus() *cobra.Command {
 	cmd.Flags().IntVar(&params.WorkerCount,
 		"worker-count", status.DefaultWorkerCount,
 		"The number of workers to use")
+	cmd.Flags().StringVarP(&params.Output, "output", "o", "summary", "Output format. One of: json, summary")
```
tklauser, line 58:
```suggestion
	cmd.Flags().StringVarP(&params.Output, "output", "o", status.OutputSummary, "Output format. One of: json, summary")
```

tklauser (approved):
LGTM, only minor nits inline regarding where `OutputJSON` and `OutputSummary` are defined. Once these are addressed and all commits squashed into one, this is ready to merge. Thanks a lot!

clustermesh/clustermesh.go
```diff
@@ -26,7 +27,9 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/cilium/cilium/api/v1/models"
+	ciliumK8s "github.com/cilium/cilium/pkg/k8s"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+	slim_corev1 "github.com/cilium/cilium/pkg/k8s/slim/k8s/api/core/v1"
```
squeed, line 32: Is there any particular reason to use the slim types? They're useful for storing lots and lots and lots of objects in memory, but we're not doing that here, right?

clustermesh/clustermesh.go
```diff
@@ -494,7 +503,9 @@ func NewK8sClusterMesh(client k8sClusterMeshImplementation, p Parameters) *K8sCl
 }
 
 func (k *K8sClusterMesh) Log(format string, a ...interface{}) {
-	fmt.Fprintf(k.params.Writer, format+"\n", a...)
+	if k.params.Output == OutputSummary {
```
squeed, line 506: This seems like the wrong place to put this conditional.

IMHO it would make more sense for the status command to compute the objects, then either output json or summary as configured.

internal/cli/cmd/status.go
```diff
@@ -29,11 +30,22 @@ func newCmdStatus() *cobra.Command {
 			}
 
 			s, err := collector.Status(context.Background())
-			// Report the most recent status even if an error occurred.
-			fmt.Print(s.Format())
 			if err != nil {
+				// Report the most recent status even if an error occurred.
+				fmt.Print(s.Format())
```
squeed, line 35: Might as well send this to `stderr` so we don't break scripts assuming json.

clustermesh/clustermesh.go
raphink: I actually don't know. IIRC @sayboras suggested it (but it was months ago, so my memory may fail me here).

clustermesh/clustermesh.go
raphink: I agree in principle. But the status is actually printed bit by bit at the moment, using this `Log` method.
In order to support JSON output, I need to:
- either make this method silent
- or build up all status strings as an array and print it at the end, which is a much larger refactoring of the summary logic

tklauser: I've fixed this up as follows:

```diff
diff --git a/clustermesh/clustermesh.go b/clustermesh/clustermesh.go
index 2095c90565e9..9f30afd39501 100644
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -500,9 +500,7 @@ func NewK8sClusterMesh(client k8sClusterMeshImplementation, p Parameters) *K8sCl
 }
 
 func (k *K8sClusterMesh) Log(format string, a ...interface{}) {
-       if k.params.Output == status.OutputSummary {
-               fmt.Fprintf(k.params.Writer, format+"\n", a...)
-       }
+       fmt.Fprintf(k.params.Writer, format+"\n", a...)
 }
 
 func (k *K8sClusterMesh) GetClusterConfig(ctx context.Context) error {
diff --git a/internal/cli/cmd/clustermesh.go b/internal/cli/cmd/clustermesh.go
index dd394835a78a..9ebbdb6d4fb1 100644
--- a/internal/cli/cmd/clustermesh.go
+++ b/internal/cli/cmd/clustermesh.go
@@ -153,6 +153,12 @@ func newCmdClusterMeshStatus() *cobra.Command {
                RunE: func(cmd *cobra.Command, args []string) error {
                        params.Namespace = namespace
 
+                       if params.Output == status.OutputJSON {
+                               // Write status log messages to stderr to make sure they don't
+                               // clutter JSON output.
+                               params.Writer = os.Stderr
+                       }
+
                        cm := clustermesh.NewK8sClusterMesh(k8sClient, params)
                        if _, err := cm.Status(context.Background()); err != nil {
                                fatalf("Unable to determine status:  %s", err)
```

clustermesh/clustermesh.go
tklauser: I don't think there is a reason to use the slim types. Users might actually be interested in fields only provided by the corev1 type. Otherwise they can still filter them out e.g. using `jq`.

I've reverted to use `corev1`. As Casey mentioned, memory usage is not a concern here (which is the primary reason to use the slim types).

Issue #918, opened 2022-06-14
CI: unable to forward port because pod is not running in EKS (tunnel) job


tklauser:
```
2022-06-14T13:10:29.398640353Z ❌ Unable to execute "kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80":
2022-06-14T13:10:29.398668197Z  error: unable to forward port because pod is not running. Current status=Pending
2022-06-14T13:10:29.398683818Z 
2022-06-14T13:10:29.398706720Z 
2022-06-14T13:10:29.398714138Z Error: Unable to port forward: unable to execute "kubectl port-forward -n kube-system svc/hubble-relay --address 0.0.0.0 --address :: 4245:80": exit status 1
2022-06-14T13:10:38.759227006Z + pgrep -f 'cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay'
2022-06-14T13:10:38.75925[80](https://github.com/cilium/cilium-cli/runs/6880678172?check_suite_focus=true#step:12:81)90Z + wc -l
2022-06-14T13:10:38.759934374Z + '[[' 0 '==' 2 ]]
```

https://github.com/cilium/cilium-cli/runs/6880678172?check_suite_focus=true

tklauser:
Also seen on a scheduled run of the same test job: https://github.com/cilium/cilium-cli/runs/6868302411?check_suite_focus=true

Haven't found any occurrences on other test jobs so far.

tklauser:
Now also seen on the AKS (BYOCNI) workflow: https://github.com/cilium/cilium-cli/runs/6913084344?check_suite_focus=true

PR #961, opened 2022-07-08
ci: explicitly wait for cilium status after enabling Hubble relay on AKS

tklauser:
It seems #918 occasionally occurs on AKS as well, e.g.
https://github.com/cilium/cilium-cli/runs/7243578074?check_suite_focus=true

Apply the same workaround as already added for EKS in commit
a0faf2510f6b ("ci: explicitly wait for cilium status after enabling
Hubble relay")

### DIFF
--- a/.github/in-cluster-test-scripts/aks.sh
+++ b/.github/in-cluster-test-scripts/aks.sh
@@ -6,6 +6,10 @@ set -e
 # Enable Relay
 cilium hubble enable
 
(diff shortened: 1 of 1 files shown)


sayboras (approved):
Same approach and command are used in CLI https://github.com/cilium/cilium-cli/commit/2482027ec0bb502266b904b2e658c5c9c1364242.

LGTM :100:

PR #962, opened 2022-07-08
connectivity: test L7 ingress policy with HTTP headers

tklauser:
The first 5 commits are cleanups, no functional changes intended.

The 6th commit contains the extension of the existing L7 HTTP ingress policy test to also test HTTP headers:

    The echo service we use for L7 HTTP policy testing conveniently provides
    two endpoints: /public and /private (see [1]). Adjust the existing
    echo-ingress-l7-http.yaml policy and the respective test to match
    against both endpoints. To test L7 HTTP policies with HTTP headers [2],
    the /private endpoint requires the X-Very-Secret-Token HTTP header
    to be set to 42.
    
    With this extended test in place the L7 policy connectivity tests cover
    the same functionality as the l7_demos.go Star Wars demo test [3] in the
    cilium/cilium repository which can now be dropped.
    
    [1] https://github.com/cilium/json-mock/blob/master/default.json
    [2] https://docs.cilium.io/en/v1.12.0-rc3/policy/language/#all-get-path1-and-put-path2-when-header-set
    [3] https://github.com/cilium/cilium/blob/4298879e53a2c9ff6bce3442920f9ab8b9d812da/test/k8s/l7_demos.go


twpayne (approved):
Perfetto

PR #964, opened 2022-07-09
hubble port-forward: Remove dependency on `cilium-cli-helm-values` se…

### LINKED ISSUE #904: `cilium hubble port-forward` broken if Cilium was installed via Helm
## Bug report

**General Information**

- Cilium CLI version (run `cilium version`): `cilium-cli: v0.11.7 compiled with go1.18.2 on linux/amd64`

1. Install Cilium via Helm (not Cilium-CLI) with Hubble Relay
2. Try to use `cilium hubble port-forward`. It fails with the following error:
```
cilium hubble port-forward
Unable to parse the provided version "", assuming v1.11.5 for ConfigMap compatibility

Error: Unable to port forward: unable to retrieve helm values secret kube-system/cilium-cli-helm-values: secrets "cilium-cli-helm-values" not found
```

This used to work before we switched the internals to use Helm, so this is a *regression*.
There is no reason why we shouldn't be able to port-forward if Cilium was installed via Helm.

patrickpichler:
If cilium is installed via helm, the `cilium-cli-helm-values` secret is
missing. This causes the `cilium hubble port-forward` command to fail,
since the factory method used to create a `K8sHubble` instance was
loading it.

Since the port-forward command doesn't use any of the values from the
secret, it has now been refactored to only use the already filled
`Parameters` instead of `K8sHubble`.

fixes #904

Signed-off-by: Patrick Pichler <git@patrickpichler.dev>


gandro (approved):
Given this very similar to how the Hubble UI port-forward is implemented, this looks good to me. Thanks for taking care of this.

tklauser (approved):
LGTM, thanks!

PR #967, opened 2022-07-12
internal/helm: pass cilium version as semver.Version instead of string

tklauser:
Only (implicitly) convert it to a string when actually needed as such.


twpayne (approved):
Nice clean-up, thanks!

Issue #871, opened 2022-05-24
sysdump: Collect bugtool report if agent is crashing


pchaigno:
When the agent on a node is crashing, the sysdump doesn't have the corresponding bugtool report. That makes it a lot harder to debug issues offline.

We should still be able to collect a bugtool report from nodes where the Cilium agent is crashing. As an example, we can collect it with the following command:
```
docker run --privileged --network host --pid host -v /tmp:/tmp -v /var/run/cilium:/var/run/cilium -v /sys/fs/bpf:/sys/fs/bpf -v /dev:/dev quay.io/cilium/cilium:[your_version] cilium-bugtool
```
It should be possible to turn this into a host-networked pod to deploy as part of the sysdump collection.

cilium-bugtool is specific to the Cilium version running so it will be important here to select the appropriate image. That said, I'm not aware of any breakage in cilium-bugtool recently.

squeed:
Taking a look at this.

Ideally it would use ephemeral containers on the Cilium pod, but those are still beta in v1.23. But we can detect if they're enabled by probing for the subresource.

PR #970, opened 2022-07-13
sysdump: run bugtool even when cilium agent is not running

squeed:
When we try and run cilium-bugtool in the agent container, it fails if that container is not running (e.g. because it's in CrashLoopBackoff). So, if that's not running, fall back to an EphemeralContainer or, if that fails, create another pod for exec purposes.

Fixes: #871

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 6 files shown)


tklauser (approved):
Looks great, thank you!

k8s/client.go
```diff
@@ -871,3 +880,35 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
 		Values:  values,
 	}, nil
 }
+
+// CreateEphemeralContainer will create a EphemeralContainer (debug container) in the specified pod.
+// EphemeralContainers are special containers which can be added after-the-fact in running pods. They're
+// useful for debugging, either when the target container image doesn't have necessary tools, or because
```
sayboras, line 886: > when the target container image doesn't have necessary tools

This could be a good usage for us going forward, a few customers are asking if cilium agent can be with scratch/distroless/minimal base image. Right now, we rely on installing new tools (e.g. tcpdump) some one cases during troubleshooting.

sayboras (approved):
LGTM :100:

sysdump/sysdump.go
```diff
@@ -1036,13 +1036,7 @@ func (c *Collector) submitBugtoolTasks(ctx context.Context, pods []*corev1.Pod,
...
 				return fmt.Errorf("failed to collect 'cilium-bugtool' output for %q in namespace %q: output doesn't contain archive name: %s", p.Name, p.Namespace, outString)
 			}
-
-			// Gzip bugtool tar file to reduce the size of transferred file
-			o, e, err = c.Client.ExecInPodWithStderr(ctx, p.Namespace, p.Name, containerName, []string{"gzip", m[1]})
-			if err != nil {
-				return fmt.Errorf("failed to compress 'cilium-bugtool' output for %q in namespace %q: %w: %s", p.Name, p.Namespace, err, e.String())
-			}
-			tarGzFile := m[1] + ".gz"
+			tarGzFile := m[1]
```
pchaigno, line 1052: `cilium sysdump` may be used on Cilium versions v1.9 to v1.12. Does cilium-bugtool support generating `.tar.gz` files on all those versions?

tklauser: It looks like it does. PR cilium/cilium#4264 added support for `--archiveType=gz` to the bugtool command. That flag is used here and produces a `.tar.gz` file. The respective commit seems to be present in all release branches from v1.9 to v1.12.

pchaigno: Ah, perfect! Yeah, from 4y ago, so should be everywhere.

squeed: I must confess, I assumed 4 years was old enough :-). Thank you indeed for checking!

k8s/client.go
squeed: Indeed, it can be a big space savings, but it will be a long time before we can rely on that feature existing. Right now it's in beta, auto-enabled but behind an optional feature gate. It is scheduled to go GA in 1.25.

PR #854, opened 2022-05-18
cmd: Add CLI argument for specifying image in  'connectivity test' subcommand

feifeifeimoon:
reference 
`cilium install --agent-image quay.mirrors.ustc.edu.cn/cilium/cilium:v1.11.4`

add three argument in `cilium connectivity test` subcommand

- `--curl-image`
- `--json-mock-image`
- `--performance-image`

Example usage:

```
$ cilium connectivity test -h
Validate connectivity in cluster

Usage:
  cilium connectivity test [flags]

Flags:
      --all-flows                  Print all flows during flow validation
      --curl-image string          Image path to use for curl (default "quay.io/cilium/alpine-curl:v1.4.0@sha256:2550c747831ff575f2147149b088ea981c06f9b6bcd188756d1b82cc10997956")
  -d, --debug                      Show debug messages
      --flow-validation string     Enable Hubble flow validation { disabled | warning | strict } (default "warning")
      --force-deploy               Force re-deploying test artifacts
  -h, --help                       help for test
      --host-net                   Use host networking during network performance tests
      --hubble                     Automatically use Hubble for flow validation & troubleshooting (default true)
      --hubble-server string       Address of the Hubble endpoint for flow validation (default "localhost:4245")
      --json-mock-image string     Image path to use for json mock (default "quay.io/cilium/json-mock:v1.3.0@sha256:2729064827fa9dbfface8d3df424feb6c792a0ba07117b844349635c93c06d2b")
      --multi-cluster string       Test across clusters to given context
  -p, --pause-on-fail              Pause execution on test failure
      --perf                       Run network Performance tests
      --perf-crr                   Run Netperf CRR Test. --perf-samples and --perf-duration ignored
      --perf-duration duration     Duration for the Performance test to run (default 10s)
      --perf-samples int           Number of Performance samples to capture (how many times to run each test) (default 1)
      --performance-image string   Image path to use for performance (default "quay.io/cilium/network-perf:bf58fb8bc57c4933dfa6e2a9581d3925c0a0571e@sha256:9bef508b2dcaeb3e288a496b8d3f065e8636a4937ba3aebcb1732afffaccea34")
      --post-test-sleep duration   Wait time after each test before next test starts
      --print-flows                Print flow logs for each test
      --single-node                Limit to tests able to run on a single node
      --test strings               Run tests that match one of the given regular expressions, skip tests by starting the expression with '!', target Scenarios with e.g. '/pod-to-cidr'
[...]

PR #973, opened 2022-07-14
[connectivity tests] Support private CoreDNS image location

jaredledvina:
https://github.com/cilium/cilium-cli/pull/854 added support for having private images however, this image is hardcoded currently. 

This PR adds support for passing in `--dns-test-server-image` allowing folks to use the connectivity tests without having to connect to any third party registry. I also updated the image to be explicitly from docker.io follow the format of the other images.

### DIFF
--- a/connectivity/check/check.go
+++ b/connectivity/check/check.go
@@ -44,6 +44,7 @@ type Parameters struct {
 	PerformanceImage      string
 	JSONMockImage         string
 	AgentDaemonSetName    string
+	DNSTestServerImage    string
 }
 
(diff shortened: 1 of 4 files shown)


tklauser:
It looks like this PR needs a rebase after #974 was merged. There is also a [Go static check issue](https://github.com/cilium/cilium-cli/runs/7346165555?check_suite_focus=true) about unused imports. Could you please fix these up?

jaredledvina:
Thanks @tklauser - I'll do it first thing in the morning w/ some ☕

jaredledvina:
@tklauser - Rebased and fixed, 🤞 this works, thanks for the review!

Issue #143, opened 2021-03-29
install: invalid cluster-id should be blocked at creation time


cmluciano:
I was following the instructions for [setting up clustermesh](https://docs.cilium.io/en/latest/gettingstarted/clustermesh/clustermesh/#specify-the-cluster-name-and-id)  on GKE and didn't notice that line that mentioned cluster-id should be between 0..255 

Command I ran  `cilium install --cluster-name cml-cluster-a --cluster-id 14480` . This caused my cilium install to hang until it timeout out. It was then quite clear from the Cilium daemonset's pod logs that this cluster-id number was outside the expected bounds.

I realize that this is my fault for just skipping over that message in the docs, but we might consider some form of linting to short-circuit this eventual feedback loop for others that may have missed the parameter limits.

Environment info:
GKE - `v1.18.15-gke.1501`
Cilium - `1.9.5`
Cilium-CLI - `v0.5-dev@master-15910bd`

PR #978, opened 2022-07-16
clustermesh: Validate cluster-id value as per requirement

sayboras:
This commit is to make sure that cluster-id is numeric and within range
of 1-255.

https://docs.cilium.io/en/stable/gettingstarted/clustermesh/clustermesh/#gs-clustermesh

> Each cluster must be assigned a unique human-readable name as well as a numeric cluster ID (1-255)

Fixes: #143

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -905,6 +905,14 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
 			aiLocal.ClusterName, aiLocal.ClusterID)
 	}
 
+	cid, err := strconv.Atoi(aiRemote.ClusterID)
+	if err != nil {
+		return fmt.Errorf("remote cluster has non-numeric cluster ID %s. Only numeric values 1-255 are allowed", aiRemote.ClusterID)
+	}
+	if cid < 1 || cid > 255 {
+		return fmt.Errorf("remote cluster has cluster ID %d out of acceptable range (1-255)", cid)
+	}
+
 	if aiRemote.ClusterName == aiLocal.ClusterName {
 		return fmt.Errorf("remote and local cluster have the same, non-unique name: %s", aiLocal.ClusterName)
 	}


clustermesh/clustermesh.go
```diff
@@ -905,6 +905,16 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
 			aiLocal.ClusterName, aiLocal.ClusterID)
 	}
 
+	cid, err := strconv.Atoi(aiRemote.ClusterID)
+	if err != nil {
+		k.Log("❌ Cluster has the non-numeric ID 0 (cluster ID: %s): %s", aiLocal.ClusterID, err)
```
tklauser, line 910: I'm not entirely sure this message is correct wrt. ID 0. I think this could be put a bit more succinctly:
 
```suggestion
		k.Log("❌ Cluster has non-numeric cluster ID 0 %s. Only numeric values 1-255 are allowed", aiRemote.ClusterID)
```

clustermesh/clustermesh.go
```diff
@@ -905,6 +905,16 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
...
+		return err
+	}
+	if cid > 255 || cid < 1 {
+		return fmt.Errorf("local cluster has the cluster ID out of acceptable range (e.g. 1-255) (cluster ID: %d)",
```
tklauser, line 914: Isn't this the remote cluster ID?

```suggestion
		return fmt.Errorf("remote cluster has the cluster ID out of acceptable range (e.g. 1-255) (cluster ID: %d)",
```

clustermesh/clustermesh.go
```diff
@@ -905,6 +905,16 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
 			aiLocal.ClusterName, aiLocal.ClusterID)
 	}
 
+	cid, err := strconv.Atoi(aiRemote.ClusterID)
+	if err != nil {
+		k.Log("❌ Cluster has the non-numeric ID 0 (cluster ID: %s): %s", aiLocal.ClusterID, err)
+		return err
```
tklauser, line 911: I think we should return the above log message as an error instead. The strconv.Atoi error itself is not very useful to the user.

tklauser (changes requested):

tklauser (approved):

PR #983, opened 2022-07-19
Prepare for v0.12.0 release

tklauser:


tklauser:
> just taking a note here, it's probably better to update the release process so that stable.txt gets updated after the release is available. currently there is a small time window where the stable.txt points to a non-existing version.

Good point. Let's change `RELEASE.md` accordingly after this release.

tklauser:
> just taking a note here, it's probably better to update the release process so that stable.txt gets updated after the release is available. currently there is a small time window where the stable.txt points to a non-existing version.

Done in https://github.com/cilium/cilium-cli/pull/992

PR #987, opened 2022-07-21
connectivity: Detect version from cilium pods

gandro:
This commit changes the version detection away from the version state
stored during `cilium install`, and instead relies on the actual version
reported by `cilium version` in the Cilium pods. This fixes an issue
where `cilium connectivity test` could not be used if Cilium was not
installed via CLI (e.g. via Helm).

The version of all Cilium pods is parsed, and only the minimum version
is used. This ensures that we can use the lowest common denominator in
terms of features.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/blang/semver/v4"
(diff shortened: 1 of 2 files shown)


tklauser (approved):
Very nice, thanks Sebastian!

PR #992, opened 2022-07-26
RELEASE.md: bump version in stable.txt after release

tklauser:
Update stable.txt only after the new version was tagged and released.
Otherwise, there is a small time window where stable.txt would point to
an inexistent version.

Ref. https://github.com/cilium/cilium-cli/pull/983#issuecomment-1189148903


gandro (approved):

michi-covalent (approved):
let's roll

PR #995, opened 2022-07-27
connectivity: Add node-local-dns match labels

aditighag:
When node-local dns is deployed along with local
redirect policy [1], the DNS traffic is redirected
to node-local dns cache pods. Hence, allow such DNS
traffic in the connectivity test policy yamls.

[1] https://docs.cilium.io/en/latest/gettingstarted/local-redirect-policy/#node-local-dns-cache

Reported-by: Julien Boulanger <julien.boulanger@fr.clara.net>
Signed-off-by: Aditi Ghag <aditi@cilium.io>

PR #997, opened 2022-07-28
fix(connectivity): Add node-local-dns entitiy match for local ip usage case

eminaktas:
When node-local-dns is deployed with local IP, 
policies fail to verify that the connection goes to 
node-local-dns. This change allows such DNS 
traffics in the connectivity test policy yamls.

Signed-off-by: eminaktas <eminaktas34@gmail.com>

### DIFF
--- a/connectivity/manifests/allow-all-except-world.yaml
+++ b/connectivity/manifests/allow-all-except-world.yaml
@@ -13,7 +13,17 @@ spec:
     - health
     - kube-apiserver
   - toEndpoints:
-    - {}
+      - {}
+  # When node-local-dns is deployed with local IP,
+  # Cilium labels its ip as world.
+  # This change prevents failing the connectivity
+  # test for such environments.
+  - toPorts:
+      - ports:
+          - port: "53"
+            protocol: UDP
+    toEntities:
+      - world
   ingress:
   - fromEntities:
     - host
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ cilium-sysdump-*.zip
 
 # Editor metas
 .vscode/
+.idea/
--- a/connectivity/manifests/allow-all-except-world-pre-v1.11.yaml
+++ b/connectivity/manifests/allow-all-except-world-pre-v1.11.yaml
@@ -13,6 +13,16 @@ spec:
     - health
   - toEndpoints:
     - {}
+  # When node-local-dns is deployed with local IP,
+  # Cilium labels its ip as world.
+  # This change prevents failing the connectivity
+  # test for such environments.
+  - toPorts:
+      - ports:
+          - port: "53"
+            protocol: UDP
+    toEntities:
+      - world
   ingress:
   - fromEntities:
     - host
--- a/connectivity/manifests/client-egress-only-dns.yaml
+++ b/connectivity/manifests/client-egress-only-dns.yaml
@@ -15,12 +15,16 @@ spec:
         dns:
         - matchPattern: "*"
     toEndpoints:
-    - matchLabels:
-        io.kubernetes.pod.namespace: kube-system
-        k8s-app: kube-dns
-    - matchLabels:
-        io.kubernetes.pod.namespace: kube-system
-        k8s-app: coredns
-    - matchLabels:
-        io.kubernetes.pod.namespace: kube-system
-        k8s-app: node-local-dns
+      - matchExpressions:
+          - { key: 'k8s-app', operator: In, values: [ "kube-dns", "coredns", "node-local-dns", "nodelocaldns" ] }
+          - { key: 'io.kubernetes.pod.namespace', operator: In, values: [ "kube-system" ] }
+  # When node-local-dns is deployed with local IP,
+  # Cilium labels its ip as world.
+  # This change prevents failing the connectivity
+  # test for such environments.
+  - toPorts:
+    - ports:
+      - port: "53"
+        protocol: UDP
+    toEntities:
+    - world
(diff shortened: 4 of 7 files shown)


eminaktas:
ping @aditighag

tklauser:
Thanks for the PR @eminaktas.

Could you please add a short description why the change is needed and/or an issue reference to the commit message?

eminaktas:
> Thanks for the PR @eminaktas.
> 
> Could you please add a short description why the change is needed and/or an issue reference to the commit message?

Thanks for the comment @tklauser . It's done.

aditighag (changes requested):
Barring the `connectivity/manifests/client-egress-to-entities-world.yaml`, I don't think enabling access to the `world` entity is recommended, it's too permissive.

squeed:
Does https://github.com/cilium/cilium-cli/pull/995 fix this?

eminaktas:
> Does #995 fix this?

Yes, it does. However, it might introduce a security issue because this commit uses the `world` label to allow the nodelocaldns requests. Like @aditighag mentions that it's too permissive. To prevent it, Cilium should label the nodelcaldns as `host`.

Kubespray's default nodelocaldns IP is `169.254.25.10`. It creates a network interface on the node to provide a connection to nodelocaldns.

squeed:
@eminaktas #995 no longer opens access to `world`. Does it fix your use case, or do we need further work? @aditighag ?

eminaktas:
> @eminaktas #995 no longer opens access to `world`. Does it fix your use case, or do we need further work? @aditighag ?

Oh I guess I misunderstood you. #995 fixes only when node-local-dns is deployed as k8s service. This commit is to fix when node-local-dns is deployed with local IP.

aditighag:
> @eminaktas #995 no longer opens access to `world`. Does it fix your use case, or do we need further work? @aditighag ?

#995 only extends the connectivity tests for [LRP](https://docs.cilium.io/en/v1.12/gettingstarted/local-redirect-policy.html) based node-local dns deployment. It doesn't have the identity issue because LRP works with pod ips, not the 169.254... address.

eminaktas:
I added a comment [here](https://github.com/cilium/cilium/issues/18644#issuecomment-1250903016) for something that I noticed.

eminaktas:
I am kindly pinging for this case @squeed @aditighag.

I updated the PR to prevent failing the connectivity test for the environment where nodelocaldns is configured with a local IP labelled world by Cilium. In addition, I updated the policies with the narrowest authorization, so the packages do not drop.

I am looking forward to your feedback 👀 
Thanks 🙏

squeed:
Rather than adding blanket world access, can you wait for #1267 then template the node local dns address in? And, is there an easy way for cilium-cli to autodetect the dns IP?

Or, are we just being too picky here, @aditighag?

squeed (approved):

eminaktas:
Hi, @aditighag 
Actually, there is no change Cilium still can't mark the local IPs with a `host` label. I think this is currently the only solution for the test cases. Also, this will save time for other people. Like @squeed asked, are we just being too picky on this?

connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -14,6 +14,16 @@ spec:
     - kube-apiserver
   - toEndpoints:
     - {}
+  # When node-local-dns is deployed with local IP,
```
aditighag, line 17: Why not narrow this down further  -
```
- matchExpressions:
          - { key: 'k8s-app', operator: In, values: [ "node-local-dns", "nodelocaldns" ] }
          - { key: 'io.kubernetes.pod.namespace', operator: In, values: [ "kube-system" ] }
```

aditighag (changes requested):
Hi @eminaktas! I went over all the policies that the PR is modifying, and it looks we seem to relax them only for DNS traffic going to node_local pods. I deployed a couple of policies, and validated that rest of the traffic to `world` is blocked. 
While it's not ideal to poke holes in the connectivity manifests as the `world` identity is fully permissive, until we have better support to add custom identities, we probably need this workaround.  We should follow up on -https://github.com/cilium/cilium/issues/18644#issuecomment-1062265283. This also includes making LRP as stable so that users won't have to deal with such IP addresses in their cluster. 
I've requested changes at one place, rest of the changes are fine.

connectivity/manifests/allow-all-except-world.yaml
eminaktas: Yeah, I added and tested there wasn't any issue. thanks.

eminaktas:
> Hi @eminaktas! I went over all the policies that the PR is modifying, and it looks we seem to relax them only for DNS traffic going to node_local pods. I deployed a couple of policies, and validated that rest of the traffic to `world` is blocked. While it's not ideal to poke holes in the connectivity manifests as the `world` identity is fully permissive, until we have better support to add custom identities, we probably need this workaround. We should follow up on -[cilium/cilium#18644 (comment)](https://github.com/cilium/cilium/issues/18644#issuecomment-1062265283). This also includes making LRP as stable so that users won't have to deal with such IP addresses in their cluster. I've requested changes at one place, rest of the changes are fine.

Thanks for the help @aditighag.

connectivity/manifests/allow-all-except-world.yaml
eminaktas: The new change caused a failure for `Multicluster` test. You can find the logs [here](https://github.com/cilium/cilium-cli/actions/runs/3882219651/jobs/6622177117#step:19:10935).

aditighag (approved):

PR #998, opened 2022-07-28
connectivity: Introduce Cilium feature detection

gandro:
This PR adds Cilium feature detection to the Cilium connectivity test suite. Cilium features are things like L7 proxy status, chaining
mode, flow aggregation mode etc. Because some features are dependent on run-time information like kernel versions, we prefer to collect the feature status from `cilium status` where available. Other features are currently detected by parsing the Cilium ConfigMap.

Some features are either simply turned on or off, while others additionally might include a `Mode` string which provides more information about in what mode a particular feature is running (e.g. when running with CNI chaining, `Enabled` will be true, and the `Mode` string will additionally contain the name of the chained CNI).

With these changes, tests can now require certain features to be enabled. This PR adds the `FeatureL7Proxy` requirement to all tests which depend on the L7 proxy being available. Subsequent PRs will also add more conditional connectivity test behavior based on the features introduced by this PR (e.g. host firewall and chaining mode).

**Review per commit**


connectivity/check/features.go
```diff
@@ -0,0 +1,217 @@
...
+	return nil
+}
+
+func (ct *ConnectivityTest) extractFeaturesFromCiliumStatus(ctx context.Context, ciliumPod Pod, result FeatureSet) error {
+	stdout, err := ciliumPod.K8sClient.ExecInPodWithTTY(ctx, ciliumPod.Pod.Namespace, ciliumPod.Pod.Name,
+		defaults.AgentContainerName, []string{"cilium", "status", "-o", "json"})
+	if err != nil {
+		return fmt.Errorf("failed to fetch cilium status: %w", err)
+
+	}
```
tklauser, line 141: nit: drop the empty line

```suggestion
		return fmt.Errorf("failed to fetch cilium status: %w", err)
	}
```

connectivity/check/features.go
```diff
@@ -0,0 +1,217 @@
...
+
+	// Monitor aggregation level defaults to none.
+	mode := defaults.ConfigMapValueMonitorAggregatonNone
+	if v, ok := cm.Data[defaults.ConfigMapKeyMonitorAggregation]; ok {
+		mode = strings.ToLower(v)
+	}
+	result[FeatureMonitorAggregation] = FeatureStatus{
+		Enabled: mode != defaults.ConfigMapValueMonitorAggregatonNone,
+		Mode:    mode,
+	}
```
tklauser, line 130: snit: given monitor aggregation mode detection is now nicely contained within this function, I wonder whether it's still worth to keep `defaults.ConfigMapKeyMonitorAggregation` and `defaults.ConfigMapValueMonitorAggregatonNone` or rather just inline these consts

```suggestion
	mode := "none"
	if v, ok := cm.Data["monitor-aggregation"]; ok {
		mode = strings.ToLower(v)
	}
	result[FeatureMonitorAggregation] = FeatureStatus{
		Enabled: mode != "none",
		Mode:    mode,
	}
```

IMO this makes it slighty easier to read.

connectivity/check/features.go
```diff
@@ -0,0 +1,217 @@
...
+			return false
+		}
+		if req.requiresMode && (req.mode != status.Mode) {
+			return false
+		}
+	}
+
+	// empty requirement list matches any feature set
+	return len(reqs) == 0
+}
```
tklauser, line 70: Will this ever return `true` for non-empty `reqs` where required features are enabled and in the required mode?

tklauser (commented):
Very nice work :rocket: Thanks for splitting this up into commits that are easy to review by themselves.

I like the approach of using per-node `cilium status` and falling back to ConfigMap in case information is not available there.

A few small nits inline. There's also a linter and an MLH warning about missing Signed-off-by. Other than that LGTM.

connectivity/check/features.go
gandro: No, it will not, that's a bug. Will address, thanks!

tklauser: Would be nice to add a unit test while at it.

tklauser (approved):
LGTM, thank you!

FWIW, I've pushed two additional commits: one to introduce a basic unit test for `(*FeatureSet).MatchRequirements`, adressing https://github.com/cilium/cilium-cli/pull/998#discussion_r932427303 and a second to add a debug log with all detected features which IMO might be helpful for Cilium CI to catch issues with feature detection.

brb:
Sweet! Can I provide a feature list for the feature detection? :sweat_smile:

tklauser:
> Sweet! Can I provide a feature list for the feature detection? sweat_smile

Please do. FWIW, it should be fairly easy to add new features to be detected here: https://github.com/cilium/cilium-cli/pull/998/files#diff-30fe8f14d3d70ed606b3e8a9ed2305d14bc2f7630530af495df364d8d6609e74R113-R172

PR #1012, opened 2022-08-04
connectivity: Add ICMP policy test from client to client

sayboras:
This commit is to add new connectivity test (e.g. client-ingress-icmp)
with allowed ICMP policy.

---

As highlighted below, CI is still with cilium 1.11, so testing was done locally with minikube cilium 1.12.0

### DIFF
--- a/connectivity/check/features.go
+++ b/connectivity/check/features.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/blang/semver/v4"
 	"github.com/cilium/cilium/api/v1/models"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -24,6 +25,7 @@ const (
 	FeatureMonitorAggregation Feature = "monitor-aggregation"
 	FeatureL7Proxy            Feature = "l7-proxy"
 	FeatureHostFirewall       Feature = "host-firewall"
+	FeatureICMPPolicy         Feature = "icmp-policy"
 
 	FeatureKPRMode                  Feature = "kpr-mode"
 	FeatureKPRExternalIPs           Feature = "kpr-external-ips"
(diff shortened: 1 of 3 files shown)


tklauser (changes requested):
CI failure looks legitimate: https://github.com/cilium/cilium-cli/runs/7674461089?check_suite_focus=true#step:11:1089

I think this would be an excellent use case for the feature detection added by @gandro in #998.

Also, it looks like `enable-icmp-rules` was only enabled by default for Cilium v1.12 and we're currently still using v1.11 in cilium-cli CI.

sayboras:
Thanks, it's making sense why the test was passed locally for me. Let me address the above comment.

connectivity/check/context.go
```diff
@@ -492,8 +492,8 @@ func (ct *ConnectivityTest) DetectMinimumCiliumVersion(ctx context.Context) (*se
 		if err != nil {
 			return nil, fmt.Errorf("unable to fetch cilium version on pod %q: %w", name, err)
 		}
-		v, _, _ := strings.Cut(stdout.String(), "-") // strips proprietary -releaseX suffix
-		podVersion, err := semver.Parse(v)
+		v, _, _ := strings.Cut(stdout.String(), "-")                      // strips proprietary -releaseX suffix
+		podVersion, err := semver.Parse(strings.Replace(v, "\n", "", -1)) // strips newline
```
aanm, line 496: I don't know where this is coming from but the remaining part looks great. Thanks!

aanm (approved):

connectivity/check/context.go
sayboras: it's coming from my installation with master, the cilium version std output is 1.12.90\n, and the above cut command didn't remove any prefix.

```
$ ./cilium connectivity test --force-deploy
ℹ️  Single-node environment detected, enabling single-node connectivity test
connectivity test failed: unable to parse cilium version on pod "cilium-prz84": Invalid character(s) found in patch number "90\n"
```

tklauser: I've sent #1016 which fixes this in a slightly more robust way.

sayboras: Thanks, I will rebase once #1016 lands in :100:

tklauser (approved):

PR #1016, opened 2022-08-05
connectivity: trim whitespaces from `cilium version` output

tklauser:
Currently, Cilium version might not be correctly detected in all cases
due to trailing whitespaces in the stdout of the `cilium version`
command, leading to the following warning:

    ⚠️  Unable to detect Cilium version, assuming v1.12.0 for connectivity tests: unable to parse cilium version on pod "cilium-b7tng": Invalid character(s) found in patch number "6\n"

Fix this by trimming all whitespaces from the command output.


sayboras (approved):
:shipit:

Issue #959, opened 2022-07-07
Cannot enable Hubble if Cilium is originally deployed with Hubble disabled


carloruiz:
## Bug report

**General Information**

- Cilium CLI version (run `cilium version`)
```
$ cilium version
cilium-cli: v0.11.4 compiled with go1.18.1 on darwin/amd64
cilium image (default): v1.11.3
cilium image (stable): v1.11.6
cilium image (running): v1.11.4
```

- Orchestration system version in use (e.g. `kubectl version`, ...)
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.15-gke.9900", GitCommit:"750002971a60d8a06e0a403c52724257f0f68481", GitTreeState:"clean", BuildDate:"2022-06-07T09:26:11Z", GoVersion:"go1.15.15b5", Compiler:"gc", Platform:"linux/amd64"}
```
- Platform: GKE 1.20.15

**How to reproduce the issue**

1. Deploy Cilium on GKE following the steps [here](https://docs.cilium.io/en/v1.11/gettingstarted/k8s-install-helm/). Add the additional flags ` --set operator.replicas=1 --set hubble.enabled=false`. Here's the full command.

```
helm template cilium cilium/cilium --version 1.11.4 \
 --namespace kube-system \
 --set nodeinit.enabled=true \
 --set nodeinit.reconfigureKubelet=true \
 --set nodeinit.removeCbrBridge=true \
 --set cni.binPath=/home/kubernetes/bin \
 --set gke.enabled=true \
 --set ipam.mode=kubernetes \
 --set ipv4NativeRoutingCIDR=$NATIVE_ROUTING_CIDR \
 --set operator.replicas=1 \
 --set hubble.enabled=false > cilium-resources.yaml
```

2. `kubectl apply -f cilium-resources.yaml`

4. Once Cilium is up and running, run `cilium hubble enable`. The following error will occur: 
```
🔑 Found CA in secret cilium-ca

[...]

PR #928, opened 2022-06-16
hubble port-forward: Don't generate Helm manifests

michi-covalent:
Get hubble-relay service port directly from Kubernetes APIs instead of
relying on Helm charts. This way the port-forward command works on
clusters in which Cilium was installed with --chart-directory flag.

Signed-off-by: Michi Mutsuzaki <michi@isovalent.com>

PR #1020, opened 2022-08-07
hubble: fix enable/disable command when `cilium-cli-helm-values` does…

olga-mir:
… not exist

This commit fixes hubble enable and disable commands which are broken
when cilium initially was not installed with cilium-cli. In these cases
`cilium-cli-helm-values` secret does not exist and all `cilium hubble`
commands fail.

Since hubble components are cherry-picked from the generated helm manifests,
it should be safe to proceed without the full helm values state.

We should not write the new helm state to the cluster so that if in the
future other components need to use it, we don't inadvertently
over shadow real installation parameters.

Fixes: #959

Signed-off-by: Olga Mirensky <5200844+olga-mir@users.noreply.github.com>

### DIFF
--- a/defaults/defaults.go
+++ b/defaults/defaults.go
@@ -101,3 +101,22 @@ var (
 	// CiliumPodSelector is the pod selector to be used for the Cilium agents.
 	CiliumPodSelector = "k8s-app=cilium"
 )
+
+// All hubble values from `cilium-config` configmap:
+// https://github.com/cilium/cilium/blob/d9a04be9d714e5f5544cbca7ef8db7a151bfce96/install/kubernetes/cilium/templates/cilium-configmap.yaml#L709-L750
+// this list is used to cherry-pick only hubble related values for configmap patch
+// when running in unknown install state (i.e. when `cilium-cli-helm-values` doesn't exist)
+var HubbleKeys = []string{
+	"enable-hubble",
+	"hubble-disable-tls",
+	"hubble-event-buffer-capacity",
+	"hubble-event-queue-size",
+	"hubble-flow-buffer-size",
+	"hubble-listen-address",
+	"hubble-metrics",
+	"hubble-metrics-server",
+	"hubble-socket-path",
+	"hubble-tls-cert-file",
+	"hubble-tls-client-ca-files",
+	"hubble-tls-key-file",
+}
--- a/go.mod
+++ b/go.mod
@@ -83,7 +83,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/envoyproxy/go-control-plane v0.10.2-0.20220325020618-49ff273808a1 // indirect
 	github.com/envoyproxy/protoc-gen-validate v0.6.2 // indirect
-	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
+	github.com/evanphx/json-patch v5.6.0+incompatible
 	github.com/exponent-io/jsonpath v0.0.0-20151013193312-d6023ce2651d // indirect
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -5,9 +5,11 @@ package hubble
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
@@ -22,6 +24,7 @@ import (
 	"github.com/cilium/cilium/api/v1/models"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	"github.com/cilium/cilium/pkg/versioncheck"
+	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/spf13/pflag"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/cli/values"
@@ -29,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
@@ -50,6 +54,7 @@ type k8sHubbleImplementation interface {
 	CreateConfigMap(ctx context.Context, namespace string, config *corev1.ConfigMap, opts metav1.CreateOptions) (*corev1.ConfigMap, error)
 	DeleteConfigMap(ctx context.Context, namespace, name string, opts metav1.DeleteOptions) error
 	GetConfigMap(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*corev1.ConfigMap, error)
+	PatchConfigMap(ctx context.Context, namespace, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions) (*corev1.ConfigMap, error)
 	UpdateConfigMap(ctx context.Context, configMap *corev1.ConfigMap, opts metav1.UpdateOptions) (*corev1.ConfigMap, error)
 	CreateDeployment(ctx context.Context, namespace string, deployment *appsv1.Deployment, opts metav1.CreateOptions) (*appsv1.Deployment, error)
 	GetDeployment(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*appsv1.Deployment, error)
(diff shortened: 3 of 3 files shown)


michi-covalent (approved):
LGTM! i'd like to get this reviewed by @gandro as well he's been looking into this area.

gandro (changes requested):
Thanks for the PR!

This PR falls back to using default configuration for the Cilium ConfigMap. That in turn will lead to unexpected results unfortunately: I tried this PR and it reset the whole Cilium configuration (e.g. disabling all custom features I enabled during install), which I don't think is what most users want. For example, I installed Cilium with PortMap chaining enabled (`cni.chainingMode=portmap` in Helm), and this was removed after I invoked `cilium hubble enable`.

The only way to support `cilium hubble enable|disable` against an unknown installation is that we ensure we do not replace any existing objects (such as the Cilium ConfigMap and DaemonSet), and instead only patch them to add whatever we need for Hubble. That requires a bit more work, but should be doable in practice.

olga-mir:
@gandro thanks for testing and for your input. I am keen to have a deeper look into this, unless someone else can provide a faster resolution. 
after last fixes in https://github.com/cilium/cilium-cli/pull/928 and https://github.com/cilium/cilium-cli/pull/964 hubble is the only place which still depends on the `cilium-cli-helm-values` secret, and even here it should be possible to work without it as you described above.  If we can implement these flows without the helm secret in hubble, then it is probably worth getting rid of the secret altogether (don't create it in new clusters and don't look it up for the existing ones). If it can't be done safely then I reckon this use case should be declared as unsupported. 
I'll keep working on this and should come back with my findings in a few days.

gandro:
@olga-mir Thanks, makes sense. Our long-term plan for the CLI is to migrate it to use fully use Helm in the backend, so instead of having `cilium-cli-helm-values`, it will rely on Helm's state to perform things like changing configuration or enabling Hubble.

Of course, that will require one to have Cilium installed via CLI or Helm (and not e.g. via a statically applied YAML), so the having a fallback which still works even if no Helm state is available would likely be nice to have. But I'm not sure how feasible that will be.

olga-mir:
Tested this with `kind` cluster, with custom fields in the configmap which are different from defaults. This PR works in this setup, but I have further questions/concerns, so it's still draft:
1. this implementation uses hard-coded list of values which is a bit brittle. We can assert minimum version, but how do we keep this updated for future versions? If you had another idea how to cherry-pick only hubble values let me know.
2. it will proceed with unknown state even if the secret exists but failed to parse, probably unlikely, but possible. We could enumerate possible errors in `getHelmState` and only act on an error where the secret did not exist.
3. Needs test and more testing.
4. I am not sure if in general it is right to support this use case as it still creates a drift. I think this is still a good idea - the drift is minimal required to enable/disable hubble and makes it easier for people to work with cilium (it is not a proper way way to manage hubble in production-like environments anyway). If this is not valuable then I am happy to close the PR and just add a user friendly error message that explains this behavior is intentional.

gandro:
Thanks a lot. The diff approach is pretty clever!

> 1. this implementation uses hard-coded list of values which is a bit brittle. We can assert minimum version, but how do we keep this updated for future versions? If you had another idea how to cherry-pick only hubble values let me know.

Yeah, it's an unfortunate situation. Overall, I'm not too worried about the set of values, they are somewhat stable between releases, and often fall-back to a sensible default if a value is missing. If more Hubble-related values get added that break things, we'll hopefully catch it in CI.

> 2. it will proceed with unknown state even if the secret exists but failed to parse, probably unlikely, but possible. We could enumerate possible errors in getHelmState and only act on an error where the secret did not exist.

If it's not too hard to implement, I think that would be ideal.

> 3. Needs test and more testing.

There is also an issue with `gofmt` according to CI. CI also seems temporarily broken at the moment, but I think that's unrelated to this PR (it's unable to fetch the Helm charts).

> 4. I am not sure if in general it is right to support this use case as it still creates a drift. 

There is one shortcoming which just patching the ConfigMap - the `cilium` DaemonSet also needs to be patched depending on the original configuration, namely it needs to mount the Hubble TLS secrets:

https://github.com/cilium/cilium/blob/6d920b25ae069f58c9f98ab87b54c94e7218da05/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L380-L384

But I'm fine to address that in a follow-up PR. Not even sure how to filter out that particular section when diffing the deployment. 

Maybe instead of automatically creating a patch based on the old and new Helm state, I would also be fine with having a hand-crafted JSON patch in case we don't have an old Helm state. And just try to apply that over the existing deployment. It would mean different paths depending on if we have the Helm state available or not, but might be more robust.

olga-mir:
I think this is fine to leave it without version check. Looking at file blame on master, the changes to this section were done 2 and 3 years ago (with one exception 5 months ago which was only a whitespace change).

I didn't dig into having a hand-crafted JSON patch implementation, but it feels that it will be more awkward as it will require passing a new parameter along few functions calls, from `generateManifestsEnable/Disable` along with `helmMapOpts` variable and passed on until configmap patch. It doesn't look like we can just define two const patches for enable and disable.

As for the daemonset bug, it should be ok to leave for a follow-up PR because it is not a regression introduced in this PR. By default if no hubble parameters are specified, hubble is enabled both in `helm install` and `cilium install` and hubble tls is mounted in `cilium` agent. Only if hubble was explicitly disabled at install time (e.g. `cilium install --helm-set hubble.enabled=false`) then with current cilium-cli `cilium hubble enable` will not patch the ds to mount the certs.

I did find another issue though which is a regression compared to the current behavior when cilium is installed with cilium-cli. 

On released cilium-cli:
```shell
$ cilium install
$ cilium hubble enable --ui # works
``` 

On this PR:
```shell
$ helm install ...
$ ./cilium hubble enable --ui
...
Error: Unable to enable Hubble: services "hubble-peer" already exists
```
This actually looks like `cilium install` bug, it should have created `hubble-peer` service, like helm install does.

Other than that it seems ok, the configmap values are preserved and `cilium hubble [enable|disable]` commands are working, also when invoked multiple times.

defaults/defaults.go
```diff
@@ -101,3 +101,20 @@ var (
 	// CiliumPodSelector is the pod selector to be used for the Cilium agents.
 	CiliumPodSelector = "k8s-app=cilium"
 )
+
+// all hubble values from `cilium-config“ configmap
+// https://github.com/cilium/cilium/blob/d9a04be9d714e5f5544cbca7ef8db7a151bfce96/install/kubernetes/cilium/templates/cilium-configmap.yaml#L709-L750
+var HubbleKeys = []string{
```
squeed, line 107: Can you add a quick comment as to what this array is for? And, perhaps, a follow-up PR in the helm chart that points users to this code if they add more hubble config keys?

olga-mir: Done. On second thought - a key is a hubble related key if and only if it has `hubble` substring in its name, at least up until now this was always true. Would it be less hacky if we just check if the string contains `hubble`?

gandro (approved):
Thanks a lot!

gandro:
> Commit [6a64764](https://github.com/cilium/cilium-cli/commit/6a64764eb1c4cc2c680a73881fb687a744467405) does not contain "Signed-off-by".
> 
> Please follow instructions provided in https://docs.cilium.io/en/stable/contributing/development/contributing_guide/#developer-s-certificate-of-origin

This needs to be addressed before the PR can be merged

olga-mir:
the newbie mistake - merge commit caused by “update branch” in UI. I’ve removed the commit and had to force push the branch.

defaults/defaults.go
squeed: Bah, I'm not sure. @chancez?

gandro: Yeah, I like that proposal. Feel free to implement. But the manually maintained list also is fine, I don't expect too much churn.

chancez: I'm not sure to be honest. I'd have to review all the flags in more detail. I think if your checking only for `hubble` then you will also catch the `hubble-relay` options, so that might be an issue?

gandro: Hubble Relay options are stored in a different configmap. This map here is only read by the Cilium Agent and the Operator (the latter does not parse any Hubble values).

chancez: Ah I didn't read the whole context, that makes sense.

gandro:
@olga-mir There seems to be an issue with vendoring:

```
please run 'go mod tidy && go mod vendor', and submit your changes
```

Might be due to jsonpatch? Or is this a bug in CI? :thinking:

gandro (commented):
Great, thanks! If there are no further changes to the code, could you squash all the commits into one?

olga-mir:
there is no further changes to this PR, if we can leave this https://github.com/cilium/cilium-cli/pull/1020#discussion_r961515555 improvement to a follow up PR.

gandro:
> there is no further changes to this PR, if we can leave this [#1020 (comment)](https://github.com/cilium/cilium-cli/pull/1020#discussion_r961515555) improvement to a follow up PR.

Agreed, that can be done in a follow-up. Let's run CI and then this should be ready to merge. Thanks again @olga-mir!

gandro (approved):

sayboras:
Thanks a lot for your fix 💯 , this issue was annoying as part of day to day cli usage 🎖️

PR #1023, opened 2022-08-08
Add stub for datapath connectivity tests

brb:
This PR:
- Extends `cilium install` to accept `--nodes-without-cilium` to denote nodes on which Cilium should not be installed.
- Adds `FeatureNodeWithoutCilium` to test for such nodes.
- Adds `outside-to-nodeport` as an example test case of the DP conformance suite.

See individual commit msgs for more details.

Ref: https://github.com/cilium/cilium/issues/20606

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -42,6 +42,9 @@ const (
 	kindClientName              = "client"
 	kindPerfName                = "perf"
 
+	hostNetNSDeploymentName = "host-netns"
+	kindHostNetNS           = "host-netns"
+
 	EchoServerHostPort = 40000
 )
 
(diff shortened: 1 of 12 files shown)


tklauser:
@brb this needs a rebase now that #1022 is merged. Also, it seems the Go workflow is failing because `connectivity/check/deployment.go` is not `gofmt`ed properly.

brb:
@tklauser Which Go vsn is used with the linter? I formated the code with the 1.18.4?

tklauser:
> @tklauser Which Go vsn is used with the linter? I formated the code with the 1.18.4?

It's 1.18.5, but that shouldn't make a difference. Did you use the `-s` option? There is the following diff when I run `gofmt -s -w` on this PR:

```diff
diff --git a/connectivity/check/deployment.go b/connectivity/check/deployment.go
index d6110069a7a6..79121c247ff2 100644
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -705,7 +705,7 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
                                        Command:     []string{"/bin/ash", "-c", "sleep 10000000"},
                                        HostNetwork: true,
                                        Tolerations: []corev1.Toleration{
-                                               corev1.Toleration{Operator: corev1.TolerationOpExists},
+                                               {Operator: corev1.TolerationOpExists},
                                        },
                                })
                                _, err = ct.clients.src.CreateDaemonSet(ctx, ct.params.TestNamespace, ds, metav1.CreateOptions{})
```

connectivity/check/deployment.go
```diff
@@ -620,6 +691,29 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+					Labels:      map[string]string{"other": "host-netns"},
+					Command:     []string{"/bin/ash", "-c", "sleep 10000000"},
+					HostNetwork: true,
+					Tolerations: []corev1.Toleration{
+						{Operator: corev1.TolerationOpExists},
+					},
+				})
+				_, err = ct.clients.src.CreateDaemonSet(ctx, ct.params.TestNamespace, ds, metav1.CreateOptions{})
+				if err != nil {
+					return fmt.Errorf("unable to create daemonset %s: %s", hostNetNSDeploymentName, err)
```
tklauser, line 713: Nit:
```suggestion
					return fmt.Errorf("unable to create daemonset %s: %w", hostNetNSDeploymentName, err)
```

internal/cli/cmd/connectivity.go
```diff
@@ -132,6 +132,7 @@ func newCmdConnectivityTest() *cobra.Command {
 	cmd.Flags().StringVar(&params.PerformanceImage, "performance-image", defaults.ConnectivityPerformanceImage, "Image path to use for performance")
 	cmd.Flags().StringVar(&params.JSONMockImage, "json-mock-image", defaults.ConnectivityCheckJSONMockImage, "Image path to use for json mock")
 	cmd.Flags().StringVar(&params.DNSTestServerImage, "dns-test-server-image", defaults.ConnectivityDNSTestServerImage, "Image path to use for CoreDNS")
+	cmd.Flags().BoolVar(&params.Datapath, "datapath", false, "Run datapath conformance tests")
```
tklauser, line 135: I wonder whether we should mark this as hidden for now, given the following in the commit message:

> For now, the test case is protected by the flag "--datapath". In the
future we might drop the flag. I.e., the test cases will be part of the
regular connectivity suite.

connectivity/check/deployment.go
```diff
@@ -844,6 +938,18 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
+	hostNetNSPods, err := ct.client.ListPods(ctx, ct.params.TestNamespace, metav1.ListOptions{LabelSelector: "kind=" + kindHostNetNS})
+	if err != nil {
+		return fmt.Errorf("unable to list host netns pods: %s", err)
```
tklauser, line 943: Nit:
```suggestion
		return fmt.Errorf("unable to list host netns pods: %w", err)
```

install/install.go
```diff
@@ -677,6 +681,16 @@ func (k *K8sInstaller) Install(ctx context.Context) error {
 		return err
 	}
 
+	for _, nodeName := range k.params.NodesWithoutCilium {
+		k.Log("🚀 Setting \"cilium.io/no-schedule=true\" label for %q node to prevent from scheduling Cilium on it...",
+			nodeName)
+		labelPatch := `[{"op":"add","path":"/metadata/labels/cilium.io~1no-schedule","value":"true"}]`
```
tklauser, line 687: Might be worth adding a comment why the special format with `~1` is needed?

internal/cli/cmd/connectivity.go
brb: SGTM!

tklauser (approved):

tommyp1ckles (approved):

PR #1031, opened 2022-08-10
sysdump: Trim ls output

michi-covalent:
- Trim the filename list string before splitting. If there is a trailing
  newline character, the last element in the filename becomes an empty
  string.
- Wrap the error to add some extra contexts on what failed.

Ref: https://github.com/cilium/cilium-cli/pull/827

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -1177,7 +1177,7 @@ func (c *Collector) SubmitCniConflistSubtask(ctx context.Context, pods []*corev1
 			if err != nil {
 				return err
 			}
-			cniConfigFileNames := strings.Split(outputStr.String(), "\n")
+			cniConfigFileNames := strings.Split(strings.TrimSpace(outputStr.String()), "\n")
 			for _, cniFileName := range cniConfigFileNames {
 				cniConfigPath := path.Join(c.Options.CNIConfigDirectory, cniFileName)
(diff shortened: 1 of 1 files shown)


sysdump/sysdump.go
```diff
@@ -1177,15 +1177,15 @@ func (c *Collector) SubmitCniConflistSubtask(ctx context.Context, pods []*corev1
...
+			cniConfigFileNames := strings.Split(strings.TrimSpace(outputStr.String()), "\n")
 			for _, cniFileName := range cniConfigFileNames {
 				cniConfigPath := path.Join(c.Options.CNIConfigDirectory, cniFileName)
 				cniConfigFileContent, err := c.Client.ExecInPod(ctx, p.GetNamespace(), p.GetName(), containerName, []string{
 					catCommand,
 					cniConfigPath,
 				})
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to copy CNI config file '%s' from directory '%s': %w", cniFileName, c.Options.CNIConfigDirectory, err)
```
tklauser, line 1188: Nit: let's use `%q` here for consistency:

```suggestion
					return fmt.Errorf("failed to copy CNI config file %q from directory %q: %w", cniFileName, c.Options.CNIConfigDirectory, err)
```

michi-covalent: makes sense, i keep forgetting `%q` is a thing. updated ✅

tklauser (approved):

Issue #1004, opened 2022-08-03
Enable kube-proxy replacement if no kube-proxy is detected in a cluster


brb:
When installing with `cilium install`, the cilium-cli could detect whether K8s' kube-proxy is present in a cluster. If not, then it could install Cilium with `--set kubeProxyReplacement=strict`.

ksankeerth:
Hi @brb,
I'm new to cilium, Can I work on this feature?

brb:
@ksankeerth Sure, thanks!

ksankeerth:
Hi @brb,

Small help on this issue. To detect kube-proxy has been installed or not, I'm thinking to use [ListDaemonSet](https://github.com/cilium/cilium-cli/blob/master/k8s/client.go#L425) function and doing a string match(kube-proxy/regex check).  Is there any better idea? Appreciate your feedback on this.

Thanks

brb:
@ksankeerth I think listing DaemonSets in the `kube-system` ns, and then matching for `kube-proxy` should work on > 80% clusters. So yeah, I suggest to take this approach.

Another problem which needs to be solved as part of this change is to detect the kube-apiserver IP and port (https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#quick-start). Not sure what's the most reliable strategy, but one way is to find a control plane node for IP and 6443 as port (default).

ksankeerth:
> @ksankeerth I think listing DaemonSets in the `kube-system` ns, and then matching for `kube-proxy` should work on > 80% clusters. So yeah, I suggest to take this approach.
> 
> Another problem which needs to be solved as part of this change is to detect the kube-apiserver IP and port (https://docs.cilium.io/en/v1.12/gettingstarted/kubeproxy-free/#quick-start). Not sure what's the most reliable strategy, but one way is to find a control plane node for IP and 6443 as port (default).

Thanks @brb.  

For the second problem, I used already available clients(similar client we used for sysdump), so I don't need to worry about the IP and port of the Master node.  I'm doing some tests and will send a PR soon.

ksankeerth:
Hi @brb,
I was able to test and created a PR.[1039](https://github.com/cilium/cilium-cli/pull/1039). Please check if possible. Thanks

```
root@kcp:/home/ubuntu/cilium-cli# ./cilium install
ℹ️  Using Cilium version 1.12.0
🔮 Auto-detected cluster name: kubernetes
[...]

PR #1039, opened 2022-08-19
Enable kube-proxy replacement if no kube-proxy is detected in a cluster

ksankeerth:
Previously, Cilium-CLI won’t enable kube-proxy by default automatically if no kube-proxy has been installed in the cluster. This patch will detect kube-proxy installation in the cluster automatically and it’ll set the helm value kubeProxyReplacement=strict if there were no kube-proxy.

Fixes: #1004

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 
 	"github.com/cilium/cilium-cli/k8s"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type validationCheck interface {
@@ -166,5 +168,33 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
 		return fmt.Errorf("invalid encryption mode")
 	}
 
+	k.autodetectKubeProxy(ctx)
+	return nil
+}
+
+func (k *K8sInstaller) autodetectKubeProxy(ctx context.Context) error {
+	kubeSysNameSpace := "kube-system"
+	dsList, err := k.client.ListDaemonSet(ctx, kubeSysNameSpace, metav1.ListOptions{})
+	if err != nil {
+		k.Log("⏭️ Skipping auto kube-proxy detction")
+		return nil
+	}
+
+	for _, ds := range dsList.Items {
+		if strings.Contains(ds.Name, "kube-proxy") {
+			k.Log("🔮 Auto-detected kube-proxy has been installed")
+			return nil
+		}
+	}
+	apiServerHost, apiServerPort := k.client.GetAPIServerHostAndPort()
+	if apiServerHost != "" && apiServerPort != "" {
+		k.Log("🔮 Auto-detected kube-proxy has not been installed")
+		k.Log("ℹ️ Cilium will fully replace all functionalities of kube-proxy")
+		// Use HelmOpts to set auto kube-proxy installation
+		k.params.HelmOpts.Values = append(k.params.HelmOpts.Values,
+			"kubeProxyReplacement=strict",
+			fmt.Sprintf("k8sServiceHost=%s", apiServerHost),
+			fmt.Sprintf("k8sServicePort=%s", apiServerPort))
+	}
 	return nil
 }
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -11,6 +11,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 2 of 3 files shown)


ksankeerth:
Hi Team,
Please check the changes and let me know if any improvements to be done for this PR. Thanks

install/autodetect.go
```diff
@@ -166,5 +168,27 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
...
+	for _, ds := range dsList.Items {
+		if strings.Contains(ds.Name, "kube-proxy") {
+			k.Log("🔮 Auto-detected kube-proxy has been installed")
+			return nil
+		}
+	}
+	k.Log("🔮 Auto-detected kube-proxy has not been installed")
+	k.Log("ℹ️ Cilium will fully replace all functionalities of kube-proxy")
+	// Use HelmOpts to set auto kube-proxy installation
+	k.params.HelmOpts.Values = append(k.params.HelmOpts.Values, "kubeProxyReplacement=strict")
```
brb, line 192: This needs to set the kube-apiserver endpoint addr too (please see my comment https://github.com/cilium/cilium-cli/issues/1004#issuecomment-1214813920).

brb (changes requested):
Thanks. Some comment re the kube-apiserver endpoint addr.

install/autodetect.go
ksankeerth: Thanks for the review. Sorry, I missed setting the API server and API server port. I fixed it and pushed it to the branch. I implemented a new method to get the server’s host and port in [k8s-client](https://github.com/ksankeerth/cilium-cli/blob/dev-1004/k8s/client.go#L126). Please check the changes.

k8s/client.go
```diff
@@ -121,6 +123,20 @@ func (c *Client) ClusterName() (name string) {
 	return
 }
 
+func (c *Client) GetAPIServerHostAndPort() (string, string) {
+	if context, ok := c.RawConfig.Contexts[c.ContextName()]; ok {
+		addr := c.RawConfig.Clusters[context.Cluster].Server
```
brb, line 128: I'm wondering whether this returns a ClusterIP or a real endpoint IP when kube-proxy is not running. Could you check? This is important, because if it returns the ClusterIP, then Cilium won't be able to connect to the kube-apiserver, as no ClusterIP => endpoint IP mapping exists in the Cilium's BPF LB (the chicken and egg problem).

ksankeerth: Your point makes sense. I checked without `kube-proxy` running(without kube-proxy add-ons), and it always returns the IP of the VM. Another thing, I noticed, It took the server value available in `/etc/kubernetes/admin.conf`.  

Already, We use the host returned by this value to predict the kind of clusters. [AKS/EKS](https://github.com/cilium/cilium-cli/blob/master/k8s/client.go#L523).  Do You think anything else we can check to confirm this?  

Further, for the command `cilium context`, We use this value and print it as `Host`.

```
root@kcp:/home/ubuntu/cilium-cli# ./cilium context
Context: kubernetes-admin@kubernetes
Cluster: kubernetes
Auth: kubernetes-admin
Host: https://192.168.64.1:6443
```

brb: Thanks for checking. I think we are good with this.

brb (approved):

install/autodetect.go
```diff
@@ -166,5 +168,30 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
...
+			k.Log("🔮 Auto-detected kube-proxy has been installed")
+			return nil
+		}
+	}
+	apiServerHost, apiServerPort := k.client.GetAPIServerHostAndPort()
+	if apiServerHost != "" && apiServerPort != "" {
+		k.Log("🔮 Auto-detected kube-proxy has not been installed")
+		k.Log("ℹ️ Cilium will fully replace all functionalities of kube-proxy")
+		// Use HelmOpts to set auto kube-proxy installation
+		k.params.HelmOpts.Values = append(k.params.HelmOpts.Values, "kubeProxyReplacement=strict", fmt.Sprintf("%s=%s", "k8sServiceHost", apiServerHost), fmt.Sprintf("%s=%s", "k8sServicePort", apiServerPort))
```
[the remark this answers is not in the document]
ksankeerth: Thanks for the suggestion. Done

tklauser (approved):
Thank you @ksankeerth!

install/autodetect.go
```diff
@@ -166,5 +168,33 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
...
+func (k *K8sInstaller) autodetectKubeProxy(ctx context.Context) error {
+	kubeSysNameSpace := "kube-system"
+	dsList, err := k.client.ListDaemonSet(ctx, kubeSysNameSpace, metav1.ListOptions{})
+	if err != nil {
+		k.Log("⏭️ Skipping auto kube-proxy detction")
+		return nil
+	}
+
+	for _, ds := range dsList.Items {
+		if strings.Contains(ds.Name, "kube-proxy") {
```
cmssczy, line 184: maybe we should use `if ds.Nmae == "kube-proxy"` instead of `strings.Contains`?

ksankeerth: thanks for the suggestion. Yes, the Kube-proxy auto-detection may take a different DaemonSet as Kube-Proxy if they also contain "kube-proxy". Just for curiosity, have you found such cases like that? @brb wdyt?

brb: I've only seen "kube-proxy", so the suggested change makes sense to me.

ksankeerth: Thanks @brb. If you think it's to better to change now, I can create a git Issue for that? (I can fix that too)

brb: I think it faster to create a PR than an issue, so I'd go with a fix. Thanks.

Issue #1026, opened 2022-08-09
Fetch tail previous logs when Cilium is in CrashLoopBackoff


joestringer:
During Cilium install with my own PR, I hit:

```
...
⌛ Waiting for Cilium to be installed and ready...

    /¯¯\
 /¯¯\__/¯¯\    Cilium:         3 errors
 \__/¯¯\__/    Operator:       OK
 /¯¯\__/¯¯\    Hubble:         disabled
 \__/¯¯\__/    ClusterMesh:    disabled
    \__/

DaemonSet         cilium             Desired: 2, Unavailable: 2/2
Deployment        cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
Containers:       cilium             Running: 2
                  cilium-operator    Running: 1
Cluster Pods:     0/3 managed by Cilium
Image versions    cilium             quay.io/cilium/cilium-ci:3378013ead0dca0263c1598ef8ede3dbe4f[25](https://github.com/cilium/cilium/runs/7735517531?check_suite_focus=true#step:7:26)318: 2
                  cilium-operator    quay.io/cilium/operator-generic-ci:3378013ead0dca0[26](https://github.com/cilium/cilium/runs/7735517531?check_suite_focus=true#step:7:27)3c1598ef8ede3dbe4f25318: 1
Errors:           cilium             cilium          2 pods of DaemonSet cilium are not ready
                  cilium             cilium-4j4r4    unable to retrieve cilium status: unable to upgrade connection: container not found ("cilium-agent")
                  cilium             cilium-nmzps    unable to retrieve cilium status: unable to upgrade connection: container not found ("cilium-agent")
ℹ️  Rollback disabled with '--rollback=false', leaving installed resources behind
Error: Unable to install Cilium: timeout while waiting for status to become successful: context deadline exceeded
```

Given `unable to retrieve cilium status: unable to upgrade connection: container not found ("cilium-agent")`, it may be worthwhile to attempt to gather the `--previous` logs via kubectl and print the tail. This way there should be a much more specific error condition to look deeper into.

squeed:
Taking a look at this.

squeed:
I also filed https://github.com/cilium/cilium/pull/21012 to set the TerminationMessagePolicy: FallbackToLogsOnError for most pods in the manifests; this will mean we capture the "dying gasp" in the Status object. It persists across node reboots.

squeed:
[...]

PR #1044, opened 2022-08-22
Status: generate a more meaningful message if agent not running

squeed:
If the agent container isn't running, we still tried to do a "kubectl exec", which is guaranteed to fail. Rather, let's generate a more useful error message.

This will now report something like this:

`unable to retrieve cilium status: container cilium-agent is in
CrashLoopBackOff, exited with code 0, try 'kubectl -n kube-system logs
-c cilium-agent cilium-wls67'`

Fixes: #1026

### DIFF
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -7,6 +7,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/cilium/cilium/api/v1/models"
@@ -54,6 +57,7 @@ type k8sImplementation interface {
 	GetDeployment(ctx context.Context, namespace, name string, options metav1.GetOptions) (*appsv1.Deployment, error)
 	ListPods(ctx context.Context, namespace string, options metav1.ListOptions) (*corev1.PodList, error)
 	ListCiliumEndpoints(ctx context.Context, namespace string, options metav1.ListOptions) (*ciliumv2.CiliumEndpointList, error)
+	CiliumLogs(ctx context.Context, namespace, pod string, since time.Time, filter *regexp.Regexp) (string, error)
 }
 
 func NewK8sStatusCollector(client k8sImplementation, params K8sStatusParameters) (*K8sStatusCollector, error) {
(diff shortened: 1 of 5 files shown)


status/k8s.go
```diff
@@ -479,10 +479,42 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
+						// determine CrashLoopBackOff status. Yes, it is this ugly.
+						if containerStatus != nil {
+							if containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == "CrashLoopBackOff" {
+								desc = "is in CrashLoopBackOff"
+							}
+							if containerStatus.LastTerminationState.Terminated != nil {
+								desc = fmt.Sprintf("%s, exited with code %d", desc, containerStatus.LastTerminationState.Terminated.ExitCode)
+							}
+						}
+						err = fmt.Errorf("container %s %s, try 'kubectl -n %s logs -c %s %s'", defaults.AgentContainerName, desc,
```
joestringer, line 514: Can we just do this for the user? Even if cilium-cli just does this for the first agent that has the problem, it would be an improvement. I wouldn't want it to do this N times for N nodes in the cluster.

Background: In Cilium CI like #1026, the cluster was already destroyed so I can't really run the command.

squeed: Sure, that's a good ideal

squeed:
@joestringer Updated, can you take a look?

status/k8s.go
```diff
@@ -477,12 +480,66 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
+								terminated := containerStatus.LastTerminationState.Terminated
+								desc = fmt.Sprintf("%s, exited with code %d", desc, terminated.ExitCode)
+
+								// capture final log line, maybe it's useful
+								// either from container message or
+								dyingGasp := ""
+								if terminated.Message != "" {
+									dyingGasp = terminated.Message
+								} else if !gotPodLogs {
+									gotPodLogs = true
```
joestringer, line 527: Are tasks parallelized? Does that mean we should be using something like a `sync.Once` rather than a bool?

joestringer (commented):
Logically looks good, but there's one point I'm not sure about wrt synchronization. Perhaps the regular cilium-cli reviewers could provide some insight there or you could dig into it to double-check.

status/k8s.go
```diff
@@ -477,12 +480,66 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
+									logs, err2 := k.client.CiliumLogs(ctx, pod.Namespace, pod.Name, terminated.FinishedAt.Time.Add(-2*time.Minute), nil)
```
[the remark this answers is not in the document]
joestringer: (Other option is limiting this to 10 lines or something and then we just "deal with it" if that's not enough)

status/k8s.go
kaworu: `statusTask` are [run concurrently using a workerpool](https://github.com/cilium/cilium-cli/blob/0d9bb5cacf64b162c70a78e5bd4b40d33fac5b4d/status/k8s.go#L565-L569) so I agree that synchronization is required here.

kaworu (changes requested):
Thanks @squeed overall LGTM but please take a look at Joe's comment about `statusTask` concurrency.

status/k8s.go
squeed: You caught me :-p. I figured the worst case was that we downloaded logs twice. But it could also make the race detector eventually scream, so I'll fix it.

squeed:
Added a sync.Once "gate"; PTAL

status/k8s.go
```diff
@@ -477,12 +481,68 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
+								// either from container message or a separate logs request
+								dyingGasp := ""
+								if terminated.Message != "" {
+									dyingGasp = strings.TrimSpace(terminated.Message)
+								} else {
+									agentLogsOnce.Do(func() { // in a sync.Once so we don't waste time retrieving lots of logs
+										logs, err := k.client.CiliumLogs(ctx, pod.Namespace, pod.Name, terminated.FinishedAt.Time.Add(-2*time.Minute), nil)
+										if err == nil && logs != "" {
+											dyingGasp = strings.TrimSpace(logs)
+										}
```
kaworu, line 532: If we fail to find logs the only time we enter this `func`, then we won't try another Pod correct? Not critical but not ideal either.

squeed: Once the TerminationMessagePolicy PR goes in (https://github.com/cilium/cilium/pull/21012), this code path will be very unlikely indeed. So I doubt it's a big deal.

status/k8s_test.go
```diff
@@ -140,6 +141,10 @@ func (c *k8sStatusMockClient) ListCiliumEndpoints(ctx context.Context, namespace
 	return c.ciliumEndpointList[options.LabelSelector], nil
 }
 
+func (c *k8sStatusMockClient) CiliumLogs(ctx context.Context, namespace, pod string, since time.Time, filter *regexp.Regexp) (string, error) {
+	return "[error] this is a funny thing, eh?", nil
```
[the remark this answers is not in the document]
squeed: hah, a bit sleep deprived ;-)

status/k8s.go
```diff
@@ -477,12 +481,68 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
-					s, err := k.client.CiliumStatus(ctx, k.params.Namespace, pod.Name)
+				task: func(ctx context.Context) error {
+					var s *models.StatusResponse
+					var err error
+
+					if containerStatus != nil && containerStatus.State.Running != nil {
+						// if container is running, execute "cilium status" in the container and parse the result
+
+						s, err = k.client.CiliumStatus(ctx, k.params.Namespace, pod.Name)
+
```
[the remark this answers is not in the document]
squeed: agreed.

squeed:
Update PTAL.

kaworu (approved):
:rocket:

status/k8s.go
```diff
@@ -477,12 +481,66 @@ func (k *K8sStatusCollector) status(ctx context.Context) *Status {
...
+							if containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == "CrashLoopBackOff" {
+								desc = "is in CrashLoopBackOff"
+							}
+							if containerStatus.LastTerminationState.Terminated != nil {
+								terminated := containerStatus.LastTerminationState.Terminated
+								desc = fmt.Sprintf("%s, exited with code %d", desc, terminated.ExitCode)
+
+								// capture final log line, maybe it's useful
+								// either from container message or a separate logs request
+								dyingGasp := ""
```
tklauser, line 522: Nice variable name :smile:

tklauser (approved):

PR #1050, opened 2022-08-23
Push workflow run results to Loki

michi-covalent:
Use michi-covalent/push-to-loki to push workflow run results to Loki
to make it easier to monitor the overall health of the CI. My ultimate
goal is to get rid of Slack notifications.

Ref: https://github.com/michi-covalent/push-to-loki

### DIFF
--- a/.github/workflows/loki.yaml
+++ b/.github/workflows/loki.yaml
@@ -1,0 +1,32 @@
+name: Push to Loki
+
+on:
+  workflow_run:
+    workflows:
+      - AKS (Azure IPAM)
+      - AKS (BYOCNI)
+      - EKS (tunnel)
+      - EKS (ENI)
+      - External Workloads
+      - GKE
+      - Go
+      - Image CI Build
+      - Kind
+      - Multicluster
+      - Create Release
+    types:
+      - completed
+
+permissions:
+  actions: read
+
+jobs:
+  push-to-loki:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Push to Loki
+        uses: michi-covalent/push-to-loki@v0.2.0
+        with:
+          endpoint: https://logs-prod3.grafana.net/loki/api/v1/push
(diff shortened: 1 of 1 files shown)


michi-covalent:
we need to create 2 secrets `LOKI_USERNAME` and `LOKI_PASSWORD` for this to work.

.github/workflows/loki.yaml
```diff
@@ -0,0 +1,32 @@
...
+  workflow_run:
+    workflows:
+      - AKS (Azure IPAM)
+      - AKS (BYOCNI)
+      - EKS (tunnel)
+      - EKS (ENI)
+      - External Workloads
+      - GKE
+      - Go
+      - Image CI Build
```
sayboras, line 13: Arguably, these two seems to be dev/compilation/rare issue, which PR author needs to address regardless 🤔

.github/workflows/loki.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+permissions:
+  actions: read
+
+jobs:
+  push-to-loki:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Push to Loki
+        uses: michi-covalent/push-to-loki@v0.2.0
```
sayboras, line 28: should we trust this random user workflow ? I bet we are 🥇

sayboras (approved):
I take a look at the python script to check which details this workflow is pushing to loki, seems legit 💯

.github/workflows/loki.yaml
michi-covalent: yeah if it fails it's likely an issue with the pull request itself, but i think it's still useful to push the data anyways. for example, we can see if the workflow gets slower over time 🔍

michi-covalent:
... accidentally clicked on the merge button. let me revert 🤦

michi-covalent:
https://github.com/cilium/cilium-cli/pull/1051

PR #1058, opened 2022-08-23
ci: Remove Slack notifications

michi-covalent:
We don't actively monitor the Slack channel, and we now send workflow
run results to Loki so I propose we stop sending Slack notifications.


nbusseneau (approved):
> We don't actively monitor the Slack channel

I do lol. But OK!

tklauser (approved):

PR #1063, opened 2022-08-26
Prepare for v0.12.2 release

tklauser:


michi-covalent (approved):

tklauser:
External workloads workflow timed out 2nd time in a row. I think we might want to bump the timeout there as well.

sayboras:
> External workloads workflow timed out 2nd time in a row. I think we might want to bump the timeout there as well.

The test seems to be completed, timeout happened in clean up step I reckon. Agreed that we should bump the timeout.

tklauser:
> > External workloads workflow timed out 2nd time in a row. I think we might want to bump the timeout there as well.
> 
> The test seems to be completed, timeout happened in clean up step I reckon. Agreed that we should bump the timeout.

Given the above and all other tests passed, I'm going to merge this PR.

PR #1066, opened 2022-08-28
connectivity: Add more connectivity tests

sayboras:
### Description

More connectivity tests as per below

- [x] default ingress denies
- [x] default egress denies
- [x] entity related tests


sayboras:
Pushed to forked repo instead of main one, closed this PR.

PR #1068, opened 2022-08-29
connectivity: Add more tests 

sayboras:
### Description

Couple of more connectivity tests, please refer to individual commit for more details.

I will remote temp commit after first round of review.

### DIFF
--- a/.github/workflows/gke.yaml
+++ b/.github/workflows/gke.yaml
@@ -36,7 +36,7 @@ jobs:
   installation-and-connectivity:
     if: ${{ github.repository == 'cilium/cilium-cli' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
     steps:
       - name: Checkout
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
@@ -128,7 +128,7 @@ jobs:
 
       - name: Wait for test job
         env:
-          timeout: 15m
+          timeout: 30m
         run: |
           # Background wait for job to complete or timeout
           kubectl -n kube-system wait job/cilium-cli --for=condition=complete --timeout=${{ env.timeout }} &
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -33,6 +33,12 @@ var (
 	//go:embed manifests/deny-all-entities.yaml
 	denyAllEntitiesPolicyYAML string
 
+	//go:embed manifests/allow-cluster-entity.yaml
+	allowClusterEntityPolicyYAML string
+
+	//go:embed manifests/allow-host-entity.yaml
+	allowHostEntityPolicyYAML string
+
 	//go:embed manifests/allow-all-except-world.yaml
 	allowAllExceptWorldPolicyYAML string
 	//go:embed manifests/allow-all-except-world-pre-v1.11.yaml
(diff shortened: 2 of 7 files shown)


.github/workflows/gke.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
   installation-and-connectivity:
     if: ${{ github.repository == 'cilium/cilium-cli' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
```
joamaki, line 38: commit message says increase timeout for GKE and External Workload, but only gke.yaml changed.

joamaki (approved):

.github/workflows/gke.yaml
sayboras: thanks, external workload was done in parrallel and got merged to master, I will update commit message along with temp commit later.

christarazi (approved):

sayboras:
All the tests are passed, waiting for reviews then I will remove temp commit.

.github/in-cluster-test-scripts/external-workloads.sh
```diff
@@ -4,7 +4,7 @@ set -x
 set -e
 
 # Run connectivity test
-cilium connectivity test --debug --all-flows
+cilium connectivity test --debug --all-flows --test '!host-entity.*'
```
joestringer, line 7: Can you document why these tests are being skipped in the code? If there is an outstanding issue that we need to solve in order to run these tests in this environment, can you file an issue or link it here?

.github/in-cluster-test-scripts/multicluster.sh
```diff
@@ -56,4 +56,4 @@ sleep 10s
 [[ $(pgrep -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay" | wc -l) == 2 ]]
 
 # Run connectivity test
-cilium --context "${CONTEXT1}" connectivity test --debug --multi-cluster "${CONTEXT2}" --test '!/*-deny,!/pod-to-.*-nodeport' --all-flows
+cilium --context "${CONTEXT1}" connectivity test --debug --multi-cluster "${CONTEXT2}" --test '!cluster-entity*,!/*-deny,!/pod-to-.*-nodeport' --all-flows
```
joestringer, line 59: Same here, could you document the reasons?

I'm not sure why we're skipping some of those other ones either, but it makes me wonder about them. I would have thought that multi-cluster should also test cluster entity to confirm the behaviour between clusters. Once we've added these tests into the CLI, any users may start running these tests, including the ones that we are skipping in CI. They won't have the benefit of the github action telling them which tests are good and which ones are bad.

joestringer (changes requested):
> waiting for reviews then I will remove temp commit.

I'm glad you mentioned it, I usually skip PRs with temp commits :sweat_smile: 

Couple of comments on skipped tests, please document why for future developers. The tests themselves LGTM though.

joestringer:
Just to help the protocol, I've added `dont-merge/preview-only` while there is a temporary commit. This way, even if you had all reviews :+1: and all tests passing, it's still communicating that this PR should not be merged in the current state. Feel free to remove the label when you have dropped the temporary commit.

.github/in-cluster-test-scripts/multicluster.sh
sayboras: not sure about other tests, but for cluster entity, it's mainly due to below the particular setting, which is to restrict only local cluster label.

https://github.com/cilium/cilium/blob/master/pkg/policy/api/entity.go#L126

.github/in-cluster-test-scripts/external-workloads.sh
sayboras: I have added remote-note in policy, so that we don't need to skip this test.

.github/in-cluster-test-scripts/multicluster.sh
sayboras: I have adjusted the test, so that we only verify for local node.

sayboras:
GKE was passed previously after timeout increase, all comments are incorporated, temp commit was removed. Marking this ready to review again.

connectivity/suite.go
```diff
@@ -197,6 +212,64 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
...
+			return check.ResultDrop, check.ResultNone
+		})
+
+	// This policy allows cluster entity
+	ct.NewTest("cluster-entity").
+		WithPolicy(allowClusterEntityPolicyYAML).
+		WithScenarios(
+			// Only enable to local cluster for now due to the below
+			// https://github.com/cilium/cilium/blob/88c4dddede2a3b5b9a7339c1316a0dedd7229a26/pkg/policy/api/entity.go#L126
+			tests.PodToPod(tests.WithDestinationLabelsOption(map[string]string{"name": "other-same-node"})),
```
joestringer, line 257: :+1: sorta makes sense but also I think this is effectively API at this point. Ie, we should test that "cluster" does not allow traffic from pods in one cluster to pods in another cluster.

connectivity/suite.go
```diff
@@ -197,6 +212,64 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 			return check.ResultOK, check.ResultDrop
 		})
 
+	// This policy denies all ingresses by default
+	ct.NewTest("all-ingress-deny").
+		WithPolicy(denyAllIngressPolicyYAML).
+		WithScenarios(
+			tests.PodToPod(),
+			tests.PodToCIDR(), // This should be allowed by the policy
```
joestringer, line 220: It would be more explicit to lay this out further:
- Pod to Pod fails because there is no egress policy (so that part works) but then at the destination there is ingress policy that denies the traffic.
- Egress to world works because there is no egress policy, then when replies come back, they are considered as "replies" to the outbound connection so they are not subject to ingress policy.

joestringer (commented):
Couple of minor comments but I think this makes sense :+1:

connectivity/suite.go
sayboras: good point, I have added the negative test for multi cluster case :100:

connectivity/suite.go
sayboras: agreed :100: 

There is a related PR to make sure that we will verify if the traffic is dropped due to which direction (e.g. Ingress or Egress)

https://github.com/cilium/cilium-cli/pull/1095

joestringer (approved):

PR #1069, opened 2022-08-30
Add L7 HTTP method test

nebril:


connectivity/tests/pod.go
```diff
@@ -108,7 +136,11 @@ func curlEndpoints(ctx context.Context, s check.Scenario, t *check.Test,
 				"X-Very-Secret-Token": "42",
 			})
 			t.NewAction(s, epName, client, ep).Run(func(a *check.Action) {
-				a.ExecInPod(ctx, curl(ep, "-H", "X-Very-Secret-Token: 42"))
+				opts := make([]string, 0, len(curlOpts)+2)
```
sayboras, line 139: +2 or +1 :thinking:

connectivity/manifests/client-egress-l7-http-method.yaml
```diff
@@ -0,0 +1,30 @@
+---
+# client2 is allowed to contact the echo Pod
+# on port 8080 via POST method. HTTP introspection is enabled for client2.
+# The toFQDNs section relies on DNS introspection being performed by
+# the client-egress-only-dns policy.
+apiVersion: "cilium.io/v2"
+kind: CiliumNetworkPolicy
+metadata:
+  namespace: cilium-test
+  name: client-egress-l7-http-method
```
sayboras, line 10: nit: test name could be same as policy name for faster reference (e.g. client-egress-l7-method)

sayboras (approved):
Small comments as per below, but not a blocker.

LGTM :100:

connectivity/tests/pod.go
nebril: `"-H", "X-Very-Secret-Token: 42"` are two args I think :thinking:

sayboras: and that is why there are three types of people in the world, those can count, and those can't :sweat:

PR #1078, opened 2022-09-04
helm: Add support for remote chart version

sayboras:
### Description

Currently, the cilium charts are embedded inside cilium cli binary,
which will require a new cilium/charts bump in go.mod and then a new
release for cilium-cli. It's making sense to do so if some other code
changes are required to make installation working for new char version,
but most of the time, only go.mod upgrade is required.

This commit is to add the capability to install new chart version without
the need of releasing or installing new cilium-cli.

### Testing

The changes are tested on top of commit 4fc8b226 (i.e. v1.13.0-rc0 is not
supported). 

```
# Trying to download invalid version v1.14.0
$ ./cilium install --version 1.14.0   
chart "cilium" version "1.14.0" not found in https://helm.cilium.io repository

# Trying to install v1.13.0-rc0
$ ./cilium install --version 1.13.0-rc0
🔮 Auto-detected Kubernetes kind: minikube
✨ Running "minikube" validation checks
✅ Detected minikube version "1.26.1"
ℹ️  Using Cilium version 1.13.0-rc0
..

$ ./cilium status
    /¯¯\
 /¯¯\__/¯¯\    Cilium:         OK
 \__/¯¯\__/    Operator:       OK
 /¯¯\__/¯¯\    Hubble:         disabled
 \__/¯¯\__/    ClusterMesh:    disabled
    \__/

Deployment        cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
DaemonSet         cilium             Desired: 1, Ready: 1/1, Available: 1/1
Containers:       cilium             Running: 1
                  cilium-operator    Running: 1
Cluster Pods:     1/1 managed by Cilium
Image versions    cilium             quay.io/cilium/cilium-ci:latest: 1
                  cilium-operator    quay.io/cilium/operator-generic-ci:latest: 1
                  
 $ ./cilium hubble enable           
🔑 Found CA in secret cilium-ca
...
✨ Patching ConfigMap cilium-config to enable Hubble...
🚀 Creating ConfigMap for Cilium version 1.13.0-rc0...
♻️  Restarted Cilium pods
               
```

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -183,7 +183,7 @@ func newClient(namespace, k8sVersion string) (*action.Install, error) {
 	return helmClient, nil
 }
 
-func newChartFromCiliumVersion(ciliumVersion semver2.Version) (*chart.Chart, error) {
+func newChartFromEmbeddedFile(ciliumVersion semver2.Version) (*chart.Chart, error) {
 	helmTgz, err := helm.HelmFS.ReadFile(fmt.Sprintf("cilium-%s.tgz", ciliumVersion))
 	if err != nil {
 		return nil, fmt.Errorf("cilium version not found: %s", err)
@@ -216,7 +216,7 @@ func GenManifests(
 			return nil, err
 		}
 	} else {
-		helmChart, err = newChartFromCiliumVersion(ciliumVer)
+		helmChart, err = newChartFromEmbeddedFile(ciliumVer)
 		if err != nil {
 			return nil, err
 		}
@@ -332,7 +332,7 @@ func ResolveHelmChartVersion(versionFlag, chartDirectoryFlag string) (semver2.Ve
 		if err != nil {
 			return semver2.Version{}, err
 		}
-		if _, err = newChartFromCiliumVersion(version); err != nil {
+		if _, err = newChartFromEmbeddedFile(version); err != nil {
 			return semver2.Version{}, err
 		}
 		return version, nil
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -7,7 +7,11 @@ package helm
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
+	"os"
+	"path"
 	"regexp"
 	"sort"
 	"strings"
@@ -29,6 +33,8 @@ import (
 	"helm.sh/helm/v3/pkg/strvals"
 	corev1 "k8s.io/api/core/v1"
 )
+
+const ciliumChart = "https://helm.cilium.io"
 
 var settings = cli.New()
 
@@ -186,7 +192,7 @@ func newClient(namespace, k8sVersion string) (*action.Install, error) {
 func newChartFromEmbeddedFile(ciliumVersion semver2.Version) (*chart.Chart, error) {
 	helmTgz, err := helm.HelmFS.ReadFile(fmt.Sprintf("cilium-%s.tgz", ciliumVersion))
 	if err != nil {
-		return nil, fmt.Errorf("cilium version not found: %s", err)
+		return nil, fmt.Errorf("cilium version not found: %w", err)
 	}
 
 	// Check chart dependencies to make sure all are present in /charts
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):

internal/helm/helm.go
```diff
@@ -197,6 +203,33 @@ func newChartFromDirectory(directory string) (*chart.Chart, error) {
...
+	}
+	file := path.Join(pull.DestDir, fmt.Sprintf("cilium-%s.tgz", ciliumVersion))
+	defer func() {
+		_ = os.Remove(file)
+	}()
+	helmTgz, err := os.ReadFile(file)
+	if err != nil {
+		return nil, err
+	}
+	return loader.LoadArchive(bytes.NewReader(helmTgz))
```
kaworu, line 230: Since [loader.LoadArchive](https://github.com/sayboras/cilium-cli/blob/2b9f0ead4fe18825429081e75659064593dd988b/vendor/helm.sh/helm/v3/pkg/chart/loader/archive.go#L189) needs a `io.Reader`, do we need to read the whole file at once here?

```suggestion
	fd, err := os.Open(file)
	if err != nil {
		return nil, err
	}
	defer fd.Close()
	return loader.LoadArchive(fd)
```

sayboras: Good point, I need to admit that I copied this piece from another loader 🥲

internal/helm/helm.go
```diff
@@ -197,6 +203,33 @@ func newChartFromDirectory(directory string) (*chart.Chart, error) {
...
+func newChartFromRemote(ciliumVersion semver2.Version) (*chart.Chart, error) {
+	pull := action.NewPullWithOpts(action.WithConfig(new(action.Configuration)))
+	pull.Settings = settings
+	pull.RepoURL = ciliumChart
+	pull.Version = ciliumVersion.String()
+
+	tempDir, err := os.MkdirTemp("", "cilium-cli")
+	if err != nil {
+		return nil, err
+	}
```
kaworu, line 215: from the [`os.MkdirTemp`](https://pkg.go.dev/os#MkdirTemp) doc:

> […] It is the caller's responsibility to remove the directory when it is no longer needed.

```suggestion
	}
	defer os.RemoveAll(tempDir)
```

internal/helm/helm.go
```diff
@@ -218,7 +251,13 @@ func GenManifests(
...
-			return nil, err
+			if !errors.Is(err, fs.ErrNotExist) {
+				return nil, err
+			}
+			helmChart, err = newChartFromRemote(ciliumVer)
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
```
kaworu, line 262: nit: I feel like this logic could be extracted in its own `func` to avoid further nesting.

internal/helm/helm.go
```diff
@@ -333,7 +372,13 @@ func ResolveHelmChartVersion(versionFlag, chartDirectoryFlag string) (semver2.Ve
 			return semver2.Version{}, err
 		}
 		if _, err = newChartFromCiliumVersion(version); err != nil {
-			return semver2.Version{}, err
+			if !errors.Is(err, fs.ErrNotExist) {
+				return semver2.Version{}, err
+			}
+			_, err = newChartFromRemote(version)
```
kaworu, line 378: If I understand correctly, when `cilium install --version $REMOTE_CILIUM_VERSION` is executed, we will fetch the charts _twice_ (once we[ setup the `k8sInstaller`](https://github.com/cilium/cilium-cli/blob/f5b079c42bc417541f8bb7c88ddc95b171728ee0/install/install.go#L365), and again when [performing the installation](https://github.com/cilium/cilium-cli/blob/f5b079c42bc417541f8bb7c88ddc95b171728ee0/install/install.go#L619)).

I think here we should be able to do something akin to `helm search repo cilium/cilium --version $REMOTE_CILIUM_VERSION`  (i.e. validating that the version is published) without having to fetch the whole Charts. Does that make sense?

kaworu (changes requested):
Thanks for the update @sayboras. I took a closer look, and I still have a question about double fetching the charts (couple of nits on the way).

internal/helm/helm.go
```diff
@@ -218,7 +251,13 @@ func GenManifests(
 	} else {
 		helmChart, err = newChartFromCiliumVersion(ciliumVer)
```
squeed, line 252: suggestion: rename this to `newChartFromLocal` or something similar.

sayboras:
> Does it make sense for `cilium-cli` to maintain a cache of downloaded charts? The standard is for it to live in `$XDG_CACHE_HOME`, defaulting to `~/.cache`.

Good point, this will help to avoid double download (into temp dir) as what Alex mentioned above. 

I will adjust the patch soon.

squeed (approved):
Looks great!

internal/helm/helm.go
```diff
@@ -197,6 +203,55 @@ func newChartFromDirectory(directory string) (*chart.Chart, error) {
...
+// newChartFromRemoteWithCache fetches the chart from remote repository, the chart file
+// is then stored in the local cache directory for future usage.
+func newChartFromRemoteWithCache(ciliumVersion semver2.Version) (*chart.Chart, error) {
+	cacheDir, err := ciliumCacheDir()
+	if err != nil {
+		return nil, err
+	}
+
+	file := path.Join(cacheDir, fmt.Sprintf("cilium-%s.tgz", ciliumVersion))
+	if _, err = os.Stat(file); err != nil {
```
kaworu, line 215: nit: I would directly try `os.Open` here and fallback to fetching on `fs.ErrNotExist`. I don't see an problem here, but this kind of pattern can run into TOCTOU issues.

kaworu (approved):
Thanks @sayboras!

internal/helm/helm.go
sayboras: I park this one right now, as we might eventually need to read the file one more time.

PR #1095, opened 2022-09-19
policy: Add support for drop direction and reason

### LINKED ISSUE #1046: Add traffic drop sub reasons and do validation for those
Right now, when doing flow validation, the result only checks if the traffic is dropped: https://github.com/cilium/cilium-cli/blob/173587006fb0760e259ca4296db62dd6ed6f1406/connectivity/check/policy.go#L197-L198. It is not checking why and where exactly the traffic was dropped.

However, traffic can be dropped for multiple reasons such as routing problems, iptables, etc. Thus, we should have sub reasons for why a traffic is dropped so that we can be sure that the end result of a test is actually the one that we should be expecting to see.

Some examples are:
- Traffic dropped by a deny policy
- Traffic dropped at ingress of a pod
- Traffic dropped at egress of a pod

sayboras:
Please refer to individual commit for more details.

Fixes #1046

### DIFF
--- a/connectivity/check/policy.go
+++ b/connectivity/check/policy.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	flowpb "github.com/cilium/cilium/api/v1/flow"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	"github.com/cilium/cilium/pkg/k8s/client/clientset/versioned/scheme"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


connectivity/suite.go
```diff
@@ -345,13 +350,13 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 		WithPolicy(echoIngressICMPDenyPolicyYAML). // Deny ICMP traffic from client to another client
 		WithFeatureRequirements(check.RequireFeatureEnabled(check.FeatureICMPPolicy)).
 		WithScenarios(
-			tests.PodToPod(),       // Client to echo traffic should be allowed
+			//tests.PodToPod(),       // Client to echo traffic should be allowed
```
tklauser, line 353: Why is this commented out?

connectivity/check/policy.go
```diff
@@ -138,6 +139,20 @@ func deleteCNP(ctx context.Context, client *k8s.Client, cnp *ciliumv2.CiliumNetw
...
+	}
+
+	policyDenyReason = func(flow *flowpb.Flow) bool {
+		return flow.GetDropReasonDesc() == flowpb.DropReason_POLICY_DENY
+	}
+
+	defaultDenyReason = func(flow *flowpb.Flow) bool {
+		return flow.GetDropReasonDesc() == flowpb.DropReason_POLICY_DENIED
+	}
+)
```
tklauser, line 154: Any reason to not just make these plain `func`s?

connectivity/suite.go
sayboras: sorry, it was unintentional as part of debugging :100:

connectivity/check/policy.go
sayboras: Good catch :100:

aanm (approved):

sayboras:
Force push to rebase with master and update drop result for newly added test (e.g. cluster-entity, host-identity, etc)

tklauser (approved):

Issue #1083, opened 2022-09-07
Reduce sysdump warnings


pchaigno:
Almost every time someone takes a sysdump on a Cilium-managed cluster, they end up with a bunch of warnings:
```
⚠️ Deployment "hubble-relay" not found in namespace "kube-system" - this is expected if Hubble is not enabled
⚠️ Deployment "hubble-ui" not found in namespace "kube-system" - this is expected if Hubble UI is not enabled
⚠️ Deployment "clustermesh-apiserver" not found in namespace "kube-system" - this is expected if 'clustermesh-apiserver' isn't enabled
⚠️ The following tasks failed, the sysdump may be incomplete:
⚠️ [11] Collecting Cilium egress NAT policies: failed to collect Cilium egress NAT policies: the server could not find the requested resource (get ciliumegressnatpolicies.cilium.io)
⚠️ [12] Collecting Cilium local redirect policies: failed to collect Cilium local redirect policies: the server could not find the requested resource (get ciliumlocalredirectpolicies.cilium.io)
⚠️ [17] Collecting CiliumClusterwideEnvoyConfigs: failed to collect CiliumClusterwideEnvoyConfigs: the server could not find the requested resource (get ciliumclusterwideenvoyconfigs.cilium.io)
⚠️ [18] Collecting CiliumEnvoyConfigs: failed to collect CiliumEnvoyConfigs: the server could not find the requested resource (get ciliumenvoyconfigs.cilium.io)
⚠️ [23] Collecting the Hubble Relay configuration: failed to collect the Hubble Relay configuration: configmaps "hubble-relay-config" not found
⚠️ Please note that depending on your Cilium version and installation options, this may be expected
```
Most of the time all of the warnings are fine to ignore. We even add one more warning at the end to say that all warnings may be expected :rofl: 

Instead, we could (should?) check the configuration to determine if there's really a need for a warning. For example, the warning below is not needed if the egress gateway is disabled (default):
```
⚠️ [11] Collecting Cilium egress NAT policies: failed to collect Cilium egress NAT policies: the server could not find the requested resource (get ciliumegressnatpolicies.cilium.io)
```

F1ko:
@pchaigno is there something already WIP?

otherwise I would like to work on the issue

pchaigno:
@F1ko Nothing in progress that I know of. Feel free to send a PR.

PR #1098, opened 2022-09-20
chore(sysdump): reduce sysdump warnings and unify output

f1ko:
Signed-off-by: Filip Nikolic

Fixes https://github.com/cilium/cilium-cli/issues/1083

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -246,288 +246,294 @@ func (c *Collector) WriteTable(filename string, value *metav1.Table) error {
 
 // Run performs the actual sysdump collection.
 func (c *Collector) Run() error {
+	taskDescription := func(desc string) string {
+		return fmt.Sprintf("%s %s", "Collecting", desc)
+	}
+	taskFailure := func(fail string) string {
+		return fmt.Sprintf("%s %s", "failed to collect", fail)
+	}
+
 	// tasks is the list of base tasks to be run.
 	tasks := []Task{
-
 		{
-			Description: "Collect Kubernetes nodes",
+			Description: taskDescription("Kubernetes nodes"),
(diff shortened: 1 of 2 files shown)


F1ko:
@ldelossa could it be that it is a flaky test? I cannot explain as to how my changes would impact a no-policy pod-to-service connectivity test.

tklauser (changes requested):
Thanks for the PR @F1ko.

It looks like the change is a mix of convenience refactoring and the actual fix for reducing the warnings. Could you please split up the change into individual commits, i.e. commits dedicated to refactoring and then the actual commit to reduce the warnings on top of that? That way it will be easier to review the change. Thanks

nbusseneau:
@F1ko I think there might have been a mistake during the rebase, as there are merge and revert commits in the log. Would it be possible to clean up the history? I can give pointers if you need help with git handling.

Also, as MLH is suggesting, you need to sign-off all commits. I see you have done so, but the format is incorrect: you need the name and address mail in there, as pointed in the docs: `Signed-off-by: Random J Developer <random@developer.example.org>`

F1ko:
Apologies for the delay.

Now everything should be as requested.

nbusseneau:
I am auto-acking on behalf of `github-sec` and `ci-structure`, since the changes don't require reviews from these team anymore (was only necessary due to the merge commits, now gone).

nbusseneau (approved):

joestringer:
@f1ko / @tklauser is this PR still being actively developed? It seems to be a bit stale and at least needs a rebase.

f1ko:
It is indeed stale @joestringer .

Will close it for now and open a new one incorporating the feedback when I get to it.

Issue #1111, opened 2022-09-26
Split Performance Tests into their Own Sub-Command


learnitall:
## Proposal / RFE

**Is your feature request related to a problem?**

When running performance tests through `cilium connectivity test --perf`, all of the deployments required for the non-perf connectivity checks are created and validated, giving some extra bloat and execution time:

```
❯ ./cilium connectivity test --perf
ℹ️  Monitor aggregation detected, will skip some flow validation steps
⚠️  Each zone only has a single node - could impact the performance test results
--> (Needed for perf)⌛ [kind-kind] Waiting for deployments [perf-client] to become ready...
--> (Needed for perf)⌛ [kind-kind] Waiting for deployments [perf-server] to become ready...
⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/client-668b56fb44-swmr9 to appear...
⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/client2-bf46d8f5d-sb49v to appear...
⌛ [kind-kind] Waiting for pod cilium-test/client-668b56fb44-swmr9 to reach DNS server on cilium-test/echo-same-node-5557788d8-pvtr5 pod...
⌛ [kind-kind] Waiting for pod cilium-test/client2-bf46d8f5d-sb49v to reach DNS server on cilium-test/echo-same-node-5557788d8-pvtr5 pod...
⌛ [kind-kind] Waiting for pod cilium-test/client-668b56fb44-swmr9 to reach DNS server on cilium-test/echo-other-node-7dc896756-jsgr2 pod...
⌛ [kind-kind] Waiting for pod cilium-test/client2-bf46d8f5d-sb49v to reach DNS server on cilium-test/echo-other-node-7dc896756-jsgr2 pod...
⌛ [kind-kind] Waiting for pod cilium-test/client-668b56fb44-swmr9 to reach default/kubernetes service...
⌛ [kind-kind] Waiting for pod cilium-test/client2-bf46d8f5d-sb49v to reach default/kubernetes service...
--> (Needed for perf)⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/perf-client-6b76f6645-jvnk4 to appear...
--> (Needed for perf)⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/perf-client-other-node-55c54fb756-qhnlv to appear...
[...]

PR #1115, opened 2022-09-29
test/connectivity: Add early return when deploying perf resources

learnitall:
This commit reorganizes ConnectivityTest.deploy and ConnectivityTest.validateDeployment to ensure that when the `--perf` flag is given from the user, only performance-related deployments are created and validated. Prior, all resources within the connectivity test are created and validated, even though they aren't used by the performance tests. This change should lead to faster startup times for performance tests.

This commit also adjusts how perf-related deployments are returned from `ConnectivityTest.deploymentList`. Prior, client deployments were returned in `srcList` and server deployments were returned in `dstList`, implying that `srcList` and `dstList` represent deployments belonging to clients and servers respectively. After review, this is an incorrect assumption that was made, as `srcList` and `dstList` split deployments for multi-cluster uses (ie source and destination cluster). Also, a deployment used during performance tests was missing from this list.

Acts as a quick way to address the issues discussed in https://github.com/cilium/cilium-cli/issues/1111 without a refactor.

Signed-off-by: Ryan Drew <ryan.drew@isovalent.com>

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -767,9 +767,11 @@ func (ct *ConnectivityTest) deploymentList() (srcList []string, dstList []string
 	if !ct.params.Perf {
 		srcList = []string{clientDeploymentName, client2DeploymentName, echoSameNodeDeploymentName}
 	} else {
-		perfDeploymentNameManager := newPerfDeploymentNameManager(&ct.params)
-		srcList = []string{perfDeploymentNameManager.ClientName()}
-		dstList = append(dstList, perfDeploymentNameManager.ServerName())
+		perfNm := newPerfDeploymentNameManager(&ct.params)
+		srcList = []string{perfNm.ClientName(), perfNm.ServerName()}
+		if !ct.params.SingleNode {
+			srcList = append(srcList, perfNm.ClientAcrossName())
+		}
 	}
 
 	if (ct.params.MultiCluster != "" || !ct.params.SingleNode) && !ct.params.Perf {
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -335,121 +335,6 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 		}
 	}
 
-	if ct.params.MultiCluster != "" {
-		if ct.params.ForceDeploy {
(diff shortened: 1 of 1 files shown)


connectivity/check/deployment.go
```diff
@@ -766,9 +767,11 @@ func (ct *ConnectivityTest) deploymentList() (srcList []string, dstList []string
 	if !ct.params.Perf {
 		srcList = []string{clientDeploymentName, client2DeploymentName, echoSameNodeDeploymentName}
 	} else {
-		perfDeploymentNameManager := newPerfDeploymentNameManager(&ct.params)
-		srcList = []string{perfDeploymentNameManager.ClientName()}
-		dstList = append(dstList, perfDeploymentNameManager.ServerName())
```
tklauser, line 771: Now that `dstList` is returned empty for performance tests, we should probably skip the `ct.waitForDeployments` call on empty `dstDeployments` here:

https://github.com/cilium/cilium-cli/blob/0b7737bce14fbc7ace66aefc9973013b19ed4e64/connectivity/check/deployment.go#L823

to avoid logging a message with an empty liste of deployments:

https://github.com/cilium/cilium-cli/blob/0b7737bce14fbc7ace66aefc9973013b19ed4e64/connectivity/check/deployment.go#L1153

tklauser (commented):
This change is rather hard to review as it moves large code blocks around but also makes some inline changes described in the commit message, i.e. the following bit:

> This commit also adjusts how perf-related deployments are returned
from `ConnectivityTest.deploymentList`. Prior, client deployments
were returned in `srcList` and server deployments were returned
in `dstList`, implying that `srcList` and `dstList` represent
deployments belonging to clients and servers respectively.
After review, this is an incorrect assumption that was made, as
`srcList` and `dstList` split deployments for multi-cluster uses
(ie source and destination cluster). Also, a deployment
used during performance tests was missing from this list.

Would it be possible to split these changes into a separate commit?

learnitall:
Ok did a force push to adjust the PR with your feedback @tklauser - thank you for taking a look! I split the origin commit of the PR into two and added a length check on lists returned from `ConnectivityTest.deploymentList` in `ConnectivityTest.validateDeployments`.

tklauser (approved):
LGTM, thanks Ryan!

PR #1126, opened 2022-10-05
[connectivity] Only wait for NodePort if the feature is enabled

jaredledvina:
Currently, we can get stuck waiting for the NodePort even when the feature is detected as disabled. This is a quick update to also check to see if the feature is enabled before we wait.


connectivity/check/deployment.go
```diff
@@ -983,7 +983,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
-	if ct.params.MultiCluster == "" {
+	if ct.params.MultiCluster == "" && ct.features[FeatureKPRNodePort].Enabled {
```
gandro, line 986: Could you please explain why this test should be skipped?

This check only checks for the kube-proxy-replacement NodePort. But regular old NodePort with kube-proxy should still pass this connectivity check, and this is something we want to test as well.

gandro (changes requested):
This

joestringer:
I've converted this to draft while the feedback is addressed. Once you're ready to continue review, click the "Ready for review" button at the bottom of the page. Thanks for contributing!

PR #1127, opened 2022-10-05
[connectivity] Only warn when failing to read features

jaredledvina:
If a Cilium installation doesn't have, for example the `cilium-config` configmap, this logic currently returns err all the way back and cancels the tests. I feel that a potentially easier setup here is to just warn when we fail but continue on with the other ways to detect features.

### DIFF
--- a/connectivity/check/features.go
+++ b/connectivity/check/features.go
@@ -310,19 +310,19 @@ func (ct *ConnectivityTest) detectFeatures(ctx context.Context) error {
 
 		err := ct.extractFeaturesFromConfigMap(ctx, ciliumPod.K8sClient, features)
 		if err != nil {
-			return err
+			ct.Warnf("Unable to read features from configmap: %v", err)
 		}
 		err = ct.extractFeaturesFromRuntimeConfig(ctx, ciliumPod, features)
 		if err != nil {
-			return err
+			ct.Warnf("Unable to read features from runtime config: %v", err)
 		}
 		err = ct.extractFeaturesFromNodes(ctx, ciliumPod.K8sClient, features)
 		if err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


ldelossa:
Hello @jaredledvina 

The overall intent here is for the install process not to continue if we **known** its going to fail. I'm not sure continuing the installation is the right thing to do. 

Is there a concrete reason where this is useful? I'd assume if you encounter any of these fatals, you'd want to fix the causes in your environment, not continue on with the installation.

jaredledvina:
Hi @ldelossa ,

Very sorry for the delay in my response, the Thanksgiving holiday got the best of me. 

>Is there a concrete reason where this is useful? I'd assume if you encounter any of these fatals, you'd want to fix the causes in your environment, not continue on with the installation.

Sure, in our environments, we do not use the `cilium-config` configmap and instead just pass through all of the configuration as `arg`'s directly to the cilium-agent. As such, we don't have the configmap in our environment. With this change, the tests fail completely even though later on the feature are properly looked up from `extractFeaturesFromRuntimeConfig`. 

My intent with this change is that there's many ways features can be detected and we shouldn't fail the entire process just because one of them has failed. That said, if folks are amendable, I'd be happy to just limit this to warning on the configmap only. That might be the only feature check in here that might fail due to customization in an environment. Thanks for the review!

squeed:
Given that this PR has been sitting for almost a year, I'm going to close it. Feel free to reopen when ready.

jaredledvina:
👋 @squeed - I was simply waiting for a maintainer to get back to me. If folks upstream do not want this change, that's alright, I can maintain it in our fork. Thanks for taking a look!

squeed:
@ldelossa thoughts?

Issue #1125, opened 2022-10-05
[connectivity tests] ciliumPods isn't set, failing to wait for revision increment


jaredledvina:
## Bug report

With the latest version of the cli and when using the connectivity tests feature, there appears to not be any `ciliumPods` configured when we get to https://github.com/cilium/cilium-cli/blob/3235080feaa36fe30dd65f42c4261239efca1890/connectivity/check/policy.go#L58 such that this returns an empty map all the way back to https://github.com/cilium/cilium-cli/blob/3235080feaa36fe30dd65f42c4261239efca1890/connectivity/check/policy.go#L382. Later on, `waitCiliumPolicyRevisions` ends up not actually waiting for anything as the revisions list is empty here https://github.com/cilium/cilium-cli/blob/3235080feaa36fe30dd65f42c4261239efca1890/connectivity/check/policy.go#L422 . The clearest indication this is happening is that when running with `--debug` we never actually log https://github.com/cilium/cilium-cli/blob/3235080feaa36fe30dd65f42c4261239efca1890/connectivity/check/policy.go#L388 

I'm not sure when this broke or if it ever worked but, it does lead to racy test conditions. This is particularly noticeable when there's tests that should fail but pass as the test command is run _right_ before the cilium-agent processes the new CNP. 

**General Information**

- Cilium CLI version (run `cilium version`) `cilium-cli: 0.12.4 compiled with go1.19.1 on darwin/amd64` 
- Orchestration system version in use (e.g. `kubectl version`, ...): v1.21.9
- Platform / infrastructure information (e.g. AWS / Azure / GCP, image / kernel versions): AWS
- Link to relevant artifacts (policies, deployments scripts, ...): N/A
- Generate and upload a system zip: `cilium sysdump`: N/A

**How to reproduce the issue**

Right now, I'm manually adding `t.Debug` logs to verify that there's no ciliumPods but, the tests are racy and I can easily reproduce the issue when running them on a small cluster with other activity.

jaredledvina:
Of course I see the issue _after_ opening this. We need to have `k8s-app=cilium` setup such that it's detected here: https://github.com/cilium/cilium-cli/blob/3235080feaa36fe30dd65f42c4261239efca1890/connectivity/check/context.go#L474

PR #1128, opened 2022-10-05
[connectivity] Allow configuring cilium-agent pod selector

jaredledvina:
Closes https://github.com/cilium/cilium-cli/issues/1125

Currently, if you have deployed the `cilium-agent` and the pods do not have the label `k8s-app=cilium`, the tests fail to find the agent pods which doesn't completely break the tests but, avoids a lot of the automatic feature detection and some additional features. 

This PR adds another parameter to allow using a custom selector while sticking with the default `k8s-app=cilium` label.


tklauser (approved):
Thanks

PR #1131, opened 2022-10-06
Allow to delete cilium pods before connectivity tests

tklauser:
Add a `--delete-cilium-pod-on-nodes` flag to `cilium connectivity test` which will delete the Cilium pods from the specified nodes before running tests. On the specified nodes this will keep only the datapath plumbed, but no Cilium agent running.

This is implemented similar to the existing `cilium install --nodes-without-cilium` flag which will set up pod's nodeAffinity such that Cilium pods won't be scheduled on nodes with label `cilium.io/no-schedule`.

This serves to migrate the "chaos tests" currently part of Cilium in-tree tests to cilium-cli.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -1212,7 +1212,7 @@ func (k *K8sClusterMesh) determineStatusConnectivity(ctx context.Context) (*Conn
 		Clusters:       map[string]*ClusterStats{},
 	}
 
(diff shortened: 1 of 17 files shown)


nathanjsweet (approved):

kaworu (approved):
Hubble changes LGTM

jrajahalme (approved):

PR #1137, opened 2022-10-10
Don't restart Cilium on cilium uninstall

michi-covalent:
cilium uninstall calls K8sHubble.Disable to delete resources created by Hubble. We don't need to update cilium-config and restart Cilium to actually disable Hubble since we are deleting Cilium right after.

Fixes: #255


kaworu (approved):

ldelossa (approved):
looks good and very innocent.

PR #1175, opened 2022-10-26
kpr: Improve k8s API server endpoint detection to handle Kind

YutaroHayakawa:
Please see the commit for more details.

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/cilium/cilium-cli/k8s"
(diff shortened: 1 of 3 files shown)


brb:
> This doesn't work in the Kind environment since the API server
endpoint on the config can only accessed through port-forwarding on the
host.

That's interesting. Are you running it on OS X? Asking, as on Linux no port-forwarding for kube-apiserver for Kind is required.

install/autodetect.go
```diff
@@ -187,6 +188,38 @@ func (k *K8sInstaller) autodetectKubeProxy(ctx context.Context) error {
 		}
 	}
 	apiServerHost, apiServerPort := k.client.GetAPIServerHostAndPort()
+	if k.flavor.Kind == k8s.KindKind {
+		k.Log("ℹ️ Detecting real kubernetes API server addr and port on Kind")
```
tommyp1ckles, line 192: nit: Kubernetes is a proper noun

tommyp1ckles:
lgtm but same question as @brb to make sure this isn't just compatible on one platform

YutaroHayakawa:
Nope, I'm using Linux with Docker. Yeah, agree. In theory, they don't need it, but in my environment they always setup port-forwarding.

```
$ kind --version
kind version 0.11.1
$ docker version
 Version:           20.10.18
 API version:       1.41
 Go version:        go1.18.6
 Git commit:        b40c2f6
 Built:             Thu Sep  8 23:11:45 2022
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.18
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.18.6
  Git commit:       e42327a
  Built:            Thu Sep  8 23:09:37 2022
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.8
  GitCommit:        9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

```
$ docker ps
<...>
481d0dbae644   kindest/node:v1.21.1   "/usr/local/bin/entr…"   2 hours ago    Up 2 hours   127.0.0.1:34755->6443/tcp                              cluster1-control-plane
```

YutaroHayakawa:
Fixed the capitalization mentioned by Tom.

Wrt the port-forwarding stuff, even if we don't have port-forwarding, this detection mechanism should work in theory because the API server's address anyways matches the Endpoints of Kubernetes service. @brb @tommyp1ckles, Could you quickly test this on your environment? If it works, then we can move forward and merge it.

squeed:
You can get the "internal" kubeconfig just by running `kind get kubeconfig --name $CLUSTER_NAME --internal`. That will have the real apiserver domain name -- does that help?

brb (approved):
Tested. Works like a charm!

PR #1203, opened 2022-11-09
test/connectivity: Add a bit more logging around Hubble connections

squeed:
When connections to hubble fail, there currently isn't enough information to debug the issue after-the-fact. So, add in a few more debug lines to try and get to the bottom of the problem.

Also, bump the connection timeout from 10 to 30 seconds.


squeed:
Needed to add a waitgroup just to stop some misleading log ordering. It doesn't have any correctness implications.

PR #1214, opened 2022-11-14
test/connectivity: fix goroutine leak

squeed:
I realized after merging that there was a goroutine leak added in commit 7644dcc362. Need to move the `wg.Done()` to a defer to ensure it always is called.

Fixes: 7644dcc362


tklauser (approved):
Oh, good catch 🎣

Issue #1235, opened 2022-11-28
cilium status shows completed pods as unmanaged


KODLQ:
## Bug report

**General Information**

- Cilium CLI version (run cilium version) v0.12.6
- Orchestration system version in use (e.g. `kubectl version`, ...) v1.25.4
- Platform / infrastructure information (e.g. AWS / Azure / GCP, image / kernel versions) GCP
- Link to relevant artifacts (policies, deployments scripts, ...) n/a
- Generate and upload a system zip: `cilium sysdump` n/a

**How to reproduce the issue**

1. Check cilium status
```cilium status```
2. Output will include pods in completed stage as unmanaged 
```Cluster Pods: CEPS / (running PODS + completed PODS)```

**Expected outcome**
* cilium status only shows amount of CEPS and running pods
```Cluster Pods: CEPS / running PODS```
* Does not add PODS in completed state.

PR #1236, opened 2022-11-28
status: Exclude non-running pods from total Cluster Pods count

felfa01:
Since CEPs are cleaned up for Completed Pods, these pods should not be considered for total Cluster Pods count.

Fixes: #1235

Signed-off-by: flxman <felix.farjsjo@gmail.com>

### DIFF
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -178,7 +178,7 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	if pods != nil && len(pods.Items) != 0 {
 		for _, pod := range pods.Items {
-			if !pod.Spec.HostNetwork {
+			if !pod.Spec.HostNetwork && pod.Status.Phase == corev1.PodRunning {
 				numberAllPod++
 			}
 		}


status/k8s.go
```diff
@@ -178,7 +178,7 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	if pods != nil && len(pods.Items) != 0 {
 		for _, pod := range pods.Items {
-			if !pod.Spec.HostNetwork {
+			if !pod.Spec.HostNetwork && pod.Status.Phase != corev1.PodSucceeded {
```
tklauser, line 181: I wonder whether we should exclude all but running pods. IMO including e.g. failed and pending pods doesn't make much sense as far as Cilium is concerned.

So how about:

```suggestion
			if !pod.Spec.HostNetwork && pod.Status.Phase == corev1.PodRunning` {
```

which would also exclude completed pods (among others)?

felfa01: @tklauser Yes, I agree with you. Commiting your suggestion, thanks!

tklauser: Thanks @felfa01. Could you please squash the two commits into one (i.e. dropping "my" commit and folding it into your change) and force-push the branch?

status/k8s.go
```diff
@@ -178,7 +178,7 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	if pods != nil && len(pods.Items) != 0 {
 		for _, pod := range pods.Items {
-			if !pod.Spec.HostNetwork {
+			if !pod.Spec.HostNetwork && pod.Status.Phase == corev1.PodRunning` {
```
felfa01, line 181:
```suggestion
			if !pod.Spec.HostNetwork && pod.Status.Phase == corev1.PodRunning {
```

status/k8s.go
felfa01: @tklauser Sure thing. Should be fine now.

tklauser (approved):
Thanks!

PR #1237, opened 2022-11-28
connectivity, install: delete pods before deleting namespace

tklauser:
This should help to work around an issue we're seeing in cilium/cilium CI where deleting the cilium-test namespace times out, see https://github.com/cilium/cilium/issues/22368

### DIFF
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -48,25 +48,8 @@ func (k *K8sUninstaller) Log(format string, a ...interface{}) {
 func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 	k.autodetect(ctx)
 
-	for _, n := range []string{k.params.TestNamespace, defaults.IngressSecretsNamespace} {
-		k.Log("🔥 Deleting %s namespace...", n)
-		k.client.DeleteNamespace(ctx, n, metav1.DeleteOptions{})
-	}
-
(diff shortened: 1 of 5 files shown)


pchaigno (approved):
I actually suggested not deleting the namespace itself at all :see_no_evil: but I think this is a good idea. If it works, it's definitely cleaner than leaving namespaces around :slightly_smiling_face:

install/uninstall.go
```diff
@@ -48,6 +49,16 @@ func (k *K8sUninstaller) Log(format string, a ...interface{}) {
 func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 	k.autodetect(ctx)
 
+	for _, kind := range []string{
+		check.KindEchoName,
+		check.KindClientName,
+		check.KindPerfName,
+		check.KindHostNetNS,
+	} {
+		k.Log("🔥 Deleting %s pods in %s namespace...", kind, k.params.TestNamespace)
```
jibi, line 58: Why not omit the `LabelSelector` altogether? That will cause the deletion of all pods under the namespace, without the need to explicitly list them (I fear that it's easy for this list to go out of sync in case we need to add some new pod in the future)

michi-covalent (approved):

pchaigno (approved):
LGTM assuming last commit is removed of course :grin:

PR #1241, opened 2022-11-29
connectivity: Add encryption tests

brb:
Please see commit msgs.

Example run - https://github.com/cilium/cilium/actions/runs/3573711857/jobs/6008107171.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -623,3 +623,8 @@ func (ct *ConnectivityTest) K8sClient() *k8s.Client {
 func (ct *ConnectivityTest) NodesWithoutCilium() []string {
 	return ct.nodesWithoutCilium
 }
+
+func (ct *ConnectivityTest) Feature(f Feature) (FeatureStatus, bool) {
+	s, ok := ct.features[f]
+	return s, ok
+}
--- a/connectivity/check/features.go
+++ b/connectivity/check/features.go
@@ -40,6 +40,8 @@ const (
 	FeatureNodeWithoutCilium Feature = "node-without-cilium"
 
 	FeatureHealthChecking Feature = "health-checking"
+
+	FeatureEncryption Feature = "encryption"
 )
 
 // FeatureStatus describes the status of a feature. Some features are either
@@ -281,6 +283,16 @@ func (ct *ConnectivityTest) extractFeaturesFromCiliumStatus(ctx context.Context,
 		}
 	}
 	result[FeatureKPRMode] = FeatureStatus{
+		Enabled: mode != "Disabled",
+		Mode:    mode,
+	}
+
+	// encryption
+	mode = "Disabled"
+	if enc := st.Encryption; enc != nil {
+		mode = enc.Mode
+	}
+	result[FeatureEncryption] = FeatureStatus{
 		Enabled: mode != "Disabled",
 		Mode:    mode,
 	}
--- a/connectivity/check/features.go
+++ b/connectivity/check/features.go
@@ -26,6 +26,7 @@ const (
 	FeatureL7Proxy            Feature = "l7-proxy"
 	FeatureHostFirewall       Feature = "host-firewall"
 	FeatureICMPPolicy         Feature = "icmp-policy"
+	FeatureTunnel             Feature = "tunnel"
 
 	FeatureKPRMode                Feature = "kpr-mode"
 	FeatureKPRExternalIPs         Feature = "kpr-external-ips"
@@ -159,6 +160,15 @@ func (ct *ConnectivityTest) extractFeaturesFromConfigMap(ctx context.Context, cl
 	}
 	result[FeatureCNIChaining] = FeatureStatus{
 		Enabled: mode != "none",
+		Mode:    mode,
+	}
+
+	mode = "Disabled"
+	if v, ok := cm.Data["tunnel"]; ok {
+		mode = v
+	}
+	result[FeatureTunnel] = FeatureStatus{
+		Enabled: mode != "Disabled",
 		Mode:    mode,
 	}
 
--- a/connectivity/tests/encryption.go
+++ b/connectivity/tests/encryption.go
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package tests
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/cilium/cilium-cli/connectivity/check"
+)
+
+// PodToPodEncryption is a test case which checks the following:
+//   - There is a connectivity between pods on different nodes when any
+//     encryption mode is on (either WireGuard or IPsec).
+//   - No unencrypted packet is leaked.
+//
+// The checks are implemented by curl'ing a server pod from a client pod, and
+// then inspecting tcpdump captures from the client pod's node.
+func PodToPodEncryption() check.Scenario {
+	return &podToPodEncryption{}
+}
+
+type podToPodEncryption struct{}
+
+func (s *podToPodEncryption) Name() string {
+	return "pod-to-pod-encryption"
+}
+
+func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
+	client := t.Context().RandomClientPod()
+
+	var server check.Pod
+	for _, pod := range t.Context().EchoPods() {
+		// Make sure that the server pod is on another node than client
+		if pod.Pod.Status.HostIP != client.Pod.Status.HostIP {
+			server = pod
+			break
+		}
+	}
+
+	bgCtx, bgCancel := context.WithCancel(ctx)
+
+	// clientHost is a pod running on the same node as the client pod, just in
+	// the host netns.
+	clientHost := t.Context().HostNetNSPodsByNode()[client.Pod.Spec.NodeName]
+
+	// Determine on which netdev iface to capture pkts. In the case of tunneling,
+	// we don't expect to see unencrypted pkts on a corresponding tunneling iface,
+	// so the choice is obvious. In the native routing mode, we run
+	// "ip route get $SERVER_POD_IP" from the client pod's node.
+	iface := ""
+	tunnelFeat, ok := t.Context().Feature(check.FeatureTunnel)
+	if ok && tunnelFeat.Enabled {
+		iface = "cilium_" + tunnelFeat.Mode // E.g. cilium_vxlan
+	} else {
+		cmd := []string{"/bin/sh", "-c",
+			fmt.Sprintf("ip -o r g %s from %s | grep -oP '(?<=dev )[^ ]+'",
+				server.Pod.Status.PodIP, client.Pod.Status.PodIP)}
+		t.Debugf("Running %s", strings.Join(cmd, " "))
+		dev, err := clientHost.K8sClient.ExecInPod(ctx, clientHost.Pod.Namespace,
+			clientHost.Pod.Name, "", cmd)
+		if err != nil {
+			t.Fatalf("Failed to get IP route: %s", err)
+		}
+		iface = strings.TrimRight(dev.String(), "\n\r")
+	}
+
+	t.Debugf("Detected %s iface for communication among client and server nodes",
+		iface)
+
+	bgStdout := &safeBuffer{}
+	bgStderr := &safeBuffer{}
+	bgExited := make(chan struct{})
+	// Start kubectl exec in bg (=goroutine)
+	go func() {
+		// Run tcpdump with -w instead of directly printing captured pkts. This
+		// is to avoid a race after sending ^C (triggered by bgCancel()) which
+		// might terminate the tcpdump process before it gets a chance to dump
+		// its captures.
+		cmd := []string{
+			"tcpdump", "-i", iface, "--immediate-mode", "-w", "/tmp/foo.pcap",
+			// Capture pod egress traffic.
+			// Unfortunately, we cannot use "host %s and host %s" filter here,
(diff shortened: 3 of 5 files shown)


connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+		// Run tcpdump with -w instead of directly printing captured pkts. This
+		// is to avoid a race after sending ^C (triggered by bgCancel()) which
+		// might terminate the tcpdump process before it gets a chance to dump
+		// its captures.
+		cmd := []string{
+			"tcpdump", "-i", iface, "--immediate-mode", "-w", "/tmp/foo.pcap",
+			// Unfortunately, we cannot use "host %s and host %s" filter here,
+			// as IPsec recirculates replies to the iface netdev, which would
+			// make tcpdump to capture the pkts (false positive)
+			fmt.Sprintf("src host %s and dst host %s", client.Pod.Status.PodIP, server.Pod.Status.PodIP),
```
pchaigno, line 94: I'd add in the above comment that we are capturing on ingress only. That's what the comment ends up saying but would be good to make it explicit.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+	// Redirect stderr to /dev/null, as tcpdump logs to stderr, and ExecInPod
+	// will return an error if any char is written to stderr. Anyway, the count
+	// is written to stdout.
+	cmd := []string{"/bin/sh", "-c", "tcpdump -r /tmp/foo.pcap --count 2>/dev/null"}
+	count, err := clientHost.K8sClient.ExecInPod(ctx, clientHost.Pod.Namespace, clientHost.Pod.Name, "", cmd)
+	if err != nil {
+		t.Fatalf("Failed to retrieve tcpdump pkt count: %s", err)
+	}
+	if !strings.HasPrefix(count.String(), "0 packets") {
+		t.Fatalf("Captured unencrypted pkt (count=%s)", count.String())
```
pchaigno, line 144: I can already predict the count here. It's going to be 1 :grin:

pchaigno (approved):
Nice commit history and comments! :heart: 

LGTM, only nits below.

Out of scope, but would be nice to check the same with a service translation on the path.

connectivity/tests/encryption.go
brb: It's also a test for tcpdump ;-)

connectivity/tests/encryption.go
brb: Makes sense, just that it's an egress traffic.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+	// "ip route get $SERVER_POD_IP" from the client pod's node. This is non-ideal
+	// when running on some managed K8s cloud envs. E.g., in a case of EKS, there
+	// might be multiple outgoing ifaces, and the cmd might yield wrong iface.
+	iface := ""
+	tunnelFeat, ok := t.Context().Feature(check.FeatureTunnel)
+	if ok && tunnelFeat.Enabled {
+		iface = "cilium_" + tunnelFeat.Mode // E.g. cilium_vxlan
+	} else {
+		cmd := []string{"/bin/sh", "-c",
+			fmt.Sprintf("ip -o r g %s | grep -oE 'dev [^ ]*' | cut -d' ' -f2", server.Pod.Status.PodIP)}
```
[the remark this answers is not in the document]
brb: Aye, good advice! The `from` part should work in the tunnel mode and the direct routing mode.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+
+	bgCtx, bgCancel := context.WithCancel(ctx)
+
+	// clientHost is a pod running on the same node as the client pod, just in
+	// the host netns.
+	clientHost := t.Context().HostNetNSPodsByNode()[client.Pod.Spec.NodeName]
+
+	// Determine on which netdev iface to capture pkts. In the case of tunneling,
+	// we don't expect to see unencrypted pkts on a corresponding tunneling iface,
+	// so the choice is obvious. While, in the native routing mode we run
```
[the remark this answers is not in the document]
brb: Not sure what I was thinking about when I added `While, `.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+func (s *podToPodEncryption) Name() string {
+	return "pod-to-pod-encryption"
+}
+
+func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
+	var client check.Pod
+	for _, pod := range t.Context().ClientPods() {
+		client = pod
+		break
+	}
```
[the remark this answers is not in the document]
brb: Sweet.

brb:
> Out of scope, but would be nice to check the same with a service translation on the path.

Yep, it's planned with https://github.com/cilium/cilium/pull/19401.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,179 @@
...
+
+	var server check.Pod
+	for _, pod := range t.Context().EchoPods() {
+		// Make sure that the server pod is on another node than client
+		if pod.Pod.Status.HostIP != client.Pod.Status.HostIP {
+			server = pod
+			break
+		}
+	}
```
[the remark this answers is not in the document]
brb: Once we have more users, we can move to the helper.

connectivity/check/features.go
```diff
@@ -285,6 +287,16 @@ func (ct *ConnectivityTest) extractFeaturesFromCiliumStatus(ctx context.Context,
 		Mode:    mode,
 	}
 
+	// encryption
+	mode = "Disabled"
```
tommyp1ckles, line 301: nit: this is probably worth making a const at this point

connectivity/check/features.go
```diff
@@ -162,6 +163,15 @@ func (ct *ConnectivityTest) extractFeaturesFromConfigMap(ctx context.Context, cl
 		Mode:    mode,
 	}
 
+	mode = "disabled"
```
tommyp1ckles, line 166: Why is this not capitalized, when all the other ones are `"Disabled"`

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// clientHost is a pod running on the same node as the client pod, just in
+	// the host netns.
+	clientHost := t.Context().HostNetNSPodsByNode()[client.Pod.Spec.NodeName]
+
+	// Determine on which netdev iface to capture pkts. In the case of tunneling,
+	// we don't expect to see unencrypted pkts on a corresponding tunneling iface,
+	// so the choice is obvious. In the native routing mode, we run
+	// "ip route get $SERVER_POD_IP" from the client pod's node. This is non-ideal
+	// when running on some managed K8s cloud envs. E.g., in a case of EKS, there
```
tommyp1ckles, line 58: In this case, might be worth logging out whatever interface we end up choosing to help with debugging?

tommyp1ckles (approved):

connectivity/check/features.go
brb: Good catch, probably it's time for a const.

brb:
@tklauser from @cilium/cli will do a review before merging.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,178 @@
...
+			// Unfortunately, we cannot use "host %s and host %s" filter here,
+			// as IPsec recirculates replies to the iface netdev, which would
+			// make tcpdump to capture the pkts (false positive).
+			fmt.Sprintf("src host %s and dst host %s", client.Pod.Status.PodIP, server.Pod.Status.PodIP),
+			// Only one pkt is enough, as we don't expect any unencrypted pkt
+			// to be captured
+			"-c", "1"}
+		t.Debugf("Running in bg: %s", strings.Join(cmd, " "))
+		err := clientHost.K8sClient.ExecInPodWithWriters(bgCtx, clientHost.Pod.Namespace,
+			clientHost.Pod.Name, "", cmd, bgStdout, bgStderr)
```
tklauser, line 99: `bgStderr` is only written to but never read. Is this intended? Otherwise we could just use

```suggestion
			clientHost.Pod.Name, "", cmd, bgStdout, io.Discard)
```

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,178 @@
...
+	}()
+
+	// Wait until tcpdump is ready to capture pkts
+	timeout := time.After(5 * time.Second)
+	for found := false; !found; {
+		select {
+		case <-timeout:
+			t.Fatalf("Failed to wait for tcpdump to be ready")
+		default:
+			line, err := bgStdout.ReadString('\n')
```
tklauser, line 113: Instead of concurrently reading from `bgStdout`, couldn't the bg goroutine check for tcpdump to be read and then just signal to this gorouting using a channel? This would probably allow to get rid of `safeBuffer` altogether. Or am I missing something?

brb: Good idea, but unfortunately the `ExecInPod*` blocks, and returns only after either a packet has been captured or ^C.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,178 @@
...
+	bgStderr := &safeBuffer{}
+	bgExited := make(chan struct{})
+	// Start kubectl exec in bg (=goroutine)
+	go func() {
+		// Run tcpdump with -w instead of directly printing captured pkts. This
+		// is to avoid a race after sending ^C (triggered by bgCancel()) which
+		// might terminate the tcpdump process before it gets a chance to dump
+		// its captures.
+		cmd := []string{
+			"tcpdump", "-i", iface, "--immediate-mode", "-w", "/tmp/foo.pcap",
```
joamaki, line 88: Would be better to generate a unique(ish) filename for this and mention that this is part of the cilium-cli encryption test. "cilium-test-encryption-XXX.pcap" where XXX is unix time?

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,178 @@
...
+	})
+
+	// Wait until tcpdump has exited
+	bgCancel()
+	<-bgExited
+
+	// Redirect stderr to /dev/null, as tcpdump logs to stderr, and ExecInPod
+	// will return an error if any char is written to stderr. Anyway, the count
+	// is written to stdout.
+	cmd := []string{"/bin/sh", "-c", "tcpdump -r /tmp/foo.pcap --count 2>/dev/null"}
```
joamaki, line 137: add `&& rm -f /tmp/foo.pcap` to not leave the file around.

joamaki (approved):

connectivity/tests/encryption.go
tklauser: Makes sense. But in that case we should probably check (or at least debug log) `bgStderr` contents somewhere after executing tcpdump?

brb: My reading of the `Exec*` code path is that the stderr is being pushed into the stdout later on.

connectivity/tests/encryption.go
```diff
@@ -0,0 +1,178 @@
...
+	// we don't expect to see unencrypted pkts on a corresponding tunneling iface,
+	// so the choice is obvious. In the native routing mode, we run
+	// "ip route get $SERVER_POD_IP" from the client pod's node.
+	iface := ""
+	tunnelFeat, ok := t.Context().Feature(check.FeatureTunnel)
+	if ok && tunnelFeat.Enabled {
+		iface = "cilium_" + tunnelFeat.Mode // E.g. cilium_vxlan
+	} else {
+		cmd := []string{"/bin/sh", "-c",
+			fmt.Sprintf("ip -o r g %s from %s | grep -oP '(?<=dev )[^ ]+'",
```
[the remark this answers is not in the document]
brb: Yep, but unfortunately this brings the runtime dependency for `jq`.

connectivity/tests/encryption.go
tklauser: Yes, so why would we need to provide `bgStderr` then? I think the stderr->stdout merge happens inside client-go if `PodExecOptions.TTY` is set, so I don't think we need to in provide an `io.Writer` for `stderr`?

In any case, with #1276 the use of `PodExecOptions.TTY` will hopefully go away soon and we'd want to populate an `io.Writer` for `stderr` (and check it at the `ExecInPod*` callsite). That being said, I'm ok to leave that as is and will clean up as part of #1276 after rebase.

Thanks!

connectivity/tests/encryption.go
pchaigno: Hm, don't we want to keep it around so that the workflows can collect it as part of the artifacts? It might be useful to debug failures.

connectivity/tests/encryption.go
brb: > That being said, I'm ok to leave that as is and will clean up as part of https://github.com/cilium/cilium-cli/pull/1276 after rebase.

Yeah, let's do that!

tklauser (approved):

connectivity/check/context.go
```diff
@@ -623,3 +623,8 @@ func (ct *ConnectivityTest) K8sClient() *k8s.Client {
 func (ct *ConnectivityTest) NodesWithoutCilium() []string {
 	return ct.nodesWithoutCilium
 }
+
+func (ct *ConnectivityTest) Feature(f Feature) (FeatureStatus, bool) {
+	s, ok := ct.features[f]
```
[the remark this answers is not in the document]
tklauser: Unfortunately, this won't work:

```
# github.com/cilium/cilium-cli/connectivity/check
connectivity/check/context.go:628:9: not enough return values
	have (FeatureStatus)
	want (FeatureStatus, bool)
```

Also see https://github.com/golang/go/issues/6230.

PR #1242, opened 2022-11-29
ci, make: allow to build release as root user

tklauser:
In local environments we usually want to keep the current behavior (i.e. release artifacts owned by the uid/gid of the user invoking `make release`). However in GitHub actions, the command recently started failing with:

    addgroup: gid '123' in use

(see https://github.com/cilium/cilium-cli/actions/runs/3572725726/jobs/6006032172)

In GitHub actions, the ownership of the release artifacts shouldn't matter since they're ephemeral anyway. We can now use the following command in the release action:

    RELEASE_UID=0 RELEASE_GID=0 RELEASER_USER= RELEASER_GROUP= make

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,7 +21,7 @@ jobs:
           go-version: 1.19.3
 
       - name: Generate the artifacts
-        run: make release
+        run: RELEASE_UID=0 RELEASE_GID=0 RELEASE_USER= RELEASE_GROUP= make release
 
       - name: Create Release
         id: create_release
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ VERSION=$(shell git describe --tags --always)
 TEST_TIMEOUT ?= 5s
 RELEASE_UID ?= $(shell id -u)
 RELEASE_GID ?= $(shell id -g)
+RELEASE_USER ?= release
+RELEASE_GROUP ?= release
 
 GOLANGCILINT_WANT_VERSION = 1.50.1
 GOLANGCILINT_VERSION = $(shell golangci-lint version 2>/dev/null)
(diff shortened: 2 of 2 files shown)


gandro (commented):
In the past, we've had problems with the release artifacts owned by root, so they couldn't be deleted anymore without sudo. But maybe that only applies to `local-release` which does a `rm` on them?

nbusseneau (approved):

tklauser:
> In the past, we've had problems with the release artifacts owned by root, so they couldn't be deleted anymore without sudo. But maybe that only applies to `local-release` which does a `rm` on them?

I think this only applies to `make local-release`. It worked on a test tag I pushed on this PR's branch: https://github.com/cilium/cilium-cli/actions/runs/3574633584

gandro:
> > In the past, we've had problems with the release artifacts owned by root, so they couldn't be deleted anymore without sudo. But maybe that only applies to `local-release` which does a `rm` on them?
> 
> I think this only applies to `make local-release`. It worked on a test tag I pushed on this PR's branch: https://github.com/cilium/cilium-cli/actions/runs/3574633584

Yes, I just double checked as well, the problem only exists if the user changes between the first and second call to `make release` (which isn't an issue on GHActions).

However, I don't fully understand how building as root is not triggering [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/), which was the whole reason we added the release user in the first place :thinking:

gandro:
> However, I don't fully understand how building as root is not triggering the protection against [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/), which was the whole reason we added the release user in the first place thinking

Ah! I guess because you added `-buildvcs=false` to go build. That explains it.

tklauser:
> > However, I don't fully understand how building as root is not triggering the protection against [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/), which was the whole reason we added the release user in the first place thinking
> 
> Ah! I guess because you added `-buildvcs=false` to go build. That explains it.

Yeah, that's the downside to it :disappointed: I couldn't find another way unfortunately.

gandro (approved):
Alternative solution: https://github.com/cilium/hubble/pull/821

michi-covalent (approved):

tklauser:
> Alternative solution: [cilium/hubble#821](https://github.com/cilium/hubble/pull/821)

Nice, thanks! TIL about `setpriv`. I'll adjust cilium-cli to that solution as well.

tklauser:
> > Alternative solution: [cilium/hubble#821](https://github.com/cilium/hubble/pull/821)
> 
> Nice, thanks! TIL about `setpriv`. I'll adjust cilium-cli to that solution as well.

Addressed in #1244

Issue #940, opened 2022-06-22
CI: GKE-based: network "default" does not have available private IP space in 10.0.0.0/8 to reserve a /14 block


tklauser:
GKE-based workflows (GKE, external workloads, multicluster) occasionally fail during cluster creation with the following error:

```
The network "default" does not have available private IP space in 10.0.0.0/9 to reserve a /14 block for pods for cluster 
``` 

Examples:
* Scheduled run on `master`: https://github.com/cilium/cilium-cli/runs/7000764590?check_suite_focus=true
* PR #933: https://github.com/cilium/cilium-cli/runs/6993592366?check_suite_focus=true

I assume this can happen if too many clusters are created simultaneously. One option to work around this would be to specify `--cluster-ipv4-cider` on `gcloud container clusters create`. Also see https://stackoverflow.com/questions/42129327/gke-cluster-creation-fails-because-the-network-default-does-not-have-available

margamanterola:
The runs linked by Tobias have now expired, but here's a recent one (from the cilium repo) that has the same error message: https://github.com/cilium/cilium/actions/runs/3464773077/jobs/5786716112

https://github.com/cilium/cilium/actions/workflows/conformance-gke.yaml is currently disabled because how flaky it is.

CC: @brlbil

PR #1251, opened 2022-12-01
workflows: add wait for no operation for cleaning up GKE

brlbil:
When a workflow cancels, the clean-up job cannot delete clusters in the provisioning state. With this commit, the clean-up job will wait for no operation on clusters before deleting the resources.

Fixes: #940


brlbil:
I did not add any run example, because this PR is basically copy of [another](https://github.com/cilium/cilium/pull/22350) on Cilium

nbusseneau (approved):

PR #1258, opened 2022-12-02
sysdump: exclude endpoint object files by default

tbalthazar:
Set the `--exclude-object-files` flag that was added in https://github.com/cilium/cilium/pull/22370 as a default flag when running sysdump.


tbalthazar:
@pchaigno @joamaki the follow-up PR to https://github.com/cilium/cilium/pull/22370

pchaigno (approved):

christarazi (approved):
Thanks for the PRs!

sysdump/sysdump.go
```diff
@@ -1186,7 +1186,7 @@ func (c *Collector) submitCiliumBugtoolTasks(ctx context.Context, pods []*corev1
 			}()
 
 			// Run 'cilium-bugtool' in the pod.
-			command := append([]string{ciliumBugtoolCommand, "--archiveType=gz"}, c.Options.CiliumBugtoolFlags...)
+			command := append([]string{ciliumBugtoolCommand, "--archiveType=gz", "--exclude-object-files"}, c.Options.CiliumBugtoolFlags...)
```
tklauser, line 1189: AFAICS, https://github.com/cilium/cilium/pull/22370 is only available in current Cilium `master` branch. However, `cilium sysdump` should also be usable with all stable releases still (currently v1.10.x, v1.11.x and v1.12.x). I think we'd need to guard the use of the `--exclude-object-files` flag on the Cilium version running in the pod.

tklauser (changes requested):

sysdump/sysdump.go
pchaigno: Oh, right :facepalm: It doesn't just ignore unknown flags.

tbalthazar: @tklauser @pchaigno That's a good point!
An approach could be to verify the version of the image in the loop, something like this:

```go
// get cilium agent version from container image
var version string
for _, container := range p.Spec.Containers {
	if container.Name == defaults.AgentContainerName {
                // e.g: quay.io/cilium/cilium:v1.12.2
		subs := strings.Split(container.Image, ":")
		version = subs[len(subs)-1]
	}
}

// pseudo-code here
flags := []string{"--archiveType=gz"}
if weHaveAVersion && semver.GTE(hardCodedNextVersionWithNewFlag) {
    flags = append(flags, "--exclude-object-files")
}

command := append([]string{ciliumBugtoolCommand}, flags, c.Options.CiliumBugtoolFlags...)
```

Would that make sense or is there a _more sophisticated_ way you already handle versions somewhere?

tbalthazar: That makes sense @tklauser, I updated the PR and left a comment Cc @pchaigno.

sysdump/sysdump.go
```diff
@@ -1170,6 +1171,27 @@ func (c *Collector) submitTetragonBugtoolTasks(ctx context.Context, pods []*core
 	return nil
 }
 
+func (c *Collector) getCiliumVersion(ctx context.Context, p *corev1.Pod) (*semver.Version, error) {
```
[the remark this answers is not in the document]
tklauser: Thanks @tbalthazar, good suggestion.

FWIW, `ExecInPodWithTTY` will probably go away soon: #1276. I think it'd make sense to use `ExecInPodWithStderr` in both places and move the implementation to a common package.  I'd suggest to move it to the `k8s` package as a `(*Client)` method instead of to the `util` package, though. We already have other helper methods which exec into pods to execute some command, e.g. https://github.com/cilium/cilium-cli/blob/43a7e471eb153942842ab41ee828d03ede7a5c81/k8s/client.go#L410 so this package seems to be a good fit.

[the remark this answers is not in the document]
tbalthazar: Hey @tklauser, I extracted the helper and rebased on main. Please let me know what you think. Thanks!

tklauser:
The staticcheck failures on the latest version of the branch look related: https://github.com/cilium/cilium-cli/actions/runs/3683477845

tbalthazar:
@tklauser lint issue should be fixed now

sysdump/sysdump.go
```diff
@@ -1194,8 +1195,23 @@ func (c *Collector) submitCiliumBugtoolTasks(ctx context.Context, pods []*corev1
...
+			bugtoolFlags := []string{"--archiveType=gz"}
+			ciliumVersion, err := c.Client.GetCiliumVersion(ctx, p)
+			if err == nil {
+				// We want to run cilium-bugtool with this flag by default, but it was
+				// introduced after this version of cilium
+				if minimumVersion, err := semver.Parse("1.12.4"); err == nil {
+					if ciliumVersion.GT(minimumVersion) {
+						bugtoolFlags = append(bugtoolFlags, "--exclude-object-files")
+					}
+				}
```
tklauser, line 1208: Let's use the `github.com/cilium/cilium/pkg/versioncheck` package here to parse  and check the Cilium version, like it is done in other places in the code base where we check Cilium version. Also, I think the `--exclude-object-files` flag will only be available starting with the upcoming Cilium 1.13 release, so we should check for a version ≥ 1.13.0 instead: 

```suggestion
				if versioncheck.MustCompile(">=1.13.0")(ciliumVersion) {
					bugtoolFlags = append(bugtoolFlags, "--exclude-object-files")
				}
```

k8s/client.go
```diff
@@ -808,6 +808,29 @@ func getCiliumVersionFromImage(image string) (string, error) {
 	return version, nil
 }
 
+// GetCiliumVersion returns a semver.Version representing the version of cilium
+// running in the cilium-agent pod
+func (c *Client) GetCiliumVersion(ctx context.Context, p *corev1.Pod) (*semver.Version, error) {
```
tklauser, line 813: Could you please split the changes factoring out `GetCiliumVersion` into a separate commit? This will improve bisectability of the commit history.

tklauser (changes requested):
Thanks @tbalthazar! I think we're almost there, only two minor remarks left.

tbalthazar:
@tklauser updated!

tklauser (approved):
LGTM, thanks @tbalthazar!

PR #1263, opened 2022-12-04
workflows: Collect sysdump immediately after test failures

pchaigno:
A new `-collect-sysdump-on-failure` option was recently introduced in the cilium connectivity test command by Jibi to collect a sysdump every time a connectivity test case fails, immediately after the failure. That ensures we get the state as it was at the time of the failure and not as it is only at the end of the full workflows.

This flag was tested on the cilium/cilium AKS workflows and this commit now extends it to all workflows in cilium/cilium-cli.

### DIFF
--- a/.github/in-cluster-test-scripts/aks.sh
+++ b/.github/in-cluster-test-scripts/aks.sh
@@ -16,7 +16,7 @@ sleep 10s
 [[ $(pgrep -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay" | wc -l) == 2 ]]
 
 # Run connectivity test
-cilium connectivity test --debug --all-flows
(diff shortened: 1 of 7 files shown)


tklauser (approved):

nbusseneau:
Shouldn't we remove the sysdump captured in the post-test step then?

https://github.com/cilium/cilium-cli/blob/a4c8c5ea0759c19310ad488555fa454fa0a115a5/.github/workflows/eks.yaml#L172

nbusseneau:
OK, I see https://github.com/cilium/cilium/pull/22537, I'll be making comments there to understand how this works.

nbusseneau:
OK, I see https://github.com/cilium/cilium/pull/22537, so I suppose we should not remove but instead rename the sysdump and make sure both sysdumps are collected by the arfitacts upload (using a wildcard).

PR #1268, opened 2022-12-06
workflows: Bump timeout for External Workload workflow

pchaigno:
We [sometimes reach the 30min timeout](https://github.com/cilium/cilium-cli/actions/runs/3580207167) on the External Workload workflow. This commit bumps the timeout to 45min.


pchaigno:
I've cancelled all workflows in this PR given the change is trivial and the CI job queue is already struggling.

tklauser (approved):

PR #1279, opened 2022-12-08
add help message for in-daemonset cilium CLI subcommands

### LINKED ISSUE #1229: Display help message if `cilium` per-cluster command is invoked with sub-command for per-node `cilium` command
Unfortunately, both the per-cluster CLI (i.e. the CLI maintained in this repository) and the per-node CLI (maintained in the `cilium/cilium` repo at https://github.com/cilium/cilium/tree/master/cilium) are named `cilium` which leads to confusion from time to time, see e.g. https://github.com/cilium/cilium-cli/issues/1206.

@aanm suggested to add a help message to the per-cluster `cilium` command which is displayed in case they are invoked with a sub-command only supported by the per-node `cilium` command:

> ... add a message, in addition to the helper message, asking if the user wanted to try out the cilium cli inside the container and giving a command example on how to do it

Additionally we could/should also update the Cilium docs to clarify the difference between the two commands.

kranurag7:
Signed-off-by: Anurag <contact.anurag7@gmail.com>
- Print a message whenever in-daemonset CLI subcommads are called for. 
```bash
$ cilium bpf 
Exec into the daemonset to use this subcommand.
To exec into the daemonset use the following command:
kubectl exec -it -n kube-system ds/cilium -- bash
$ cilium help bpf 
Exec into the daemonset to use this subcommand.
To exec into the daemonset use the following command:
kubectl exec -it -n kube-system ds/cilium -- bash
```

Fixes #1229

### DIFF
--- a/internal/cli/cmd/cmd.go
+++ b/internal/cli/cmd/cmd.go
@@ -84,5 +84,14 @@ cilium connectivity test`,
 	cmd.SetOut(os.Stdout)
 	cmd.SetErr(os.Stderr)
 
+	cmd.AddCommand(&cobra.Command{
+		Use:     "bpf",
+		Aliases: []string{"cleanup", "debuginfo", "endpoint", "fqdn", "identity", "ip", "kvstore", "lrp", "map", "metrics", "monitor", "node", "policy", "prefilter", "preflight", "recorder", "service"},
+		Hidden:  true,
+		Long: `This command is only available in cilium pods.
+Exec into a running cilium pod to use it. 
+kubectl -n kube-system exec -it ds/cilium -- bash`,
+	})
+
 	return cmd
 }


internal/cli/cmd/cmd.go
```diff
@@ -84,5 +84,14 @@ cilium connectivity test`,
 	cmd.SetOut(os.Stdout)
 	cmd.SetErr(os.Stderr)
 
+	cmd.AddCommand(&cobra.Command{
+		Use:     "bpf",
+		Aliases: []string{"cleanup", "debuginfo", "endpoint", "fqdn", "identity", "ip", "kvstore", "lrp", "map", "metrics", "monitor", "node", "policy", "prefilter", "preflight", "recorder", "service"},
+		Hidden:  true,
+		Long: `Exec into the daemonset to use this subcommand.
```
squeed, line 91: Wording suggestion:
"This command is only available in the Cilium agent.
Exec in to a running cilium pod to use it.
kubectl exec -it -n kube-system ds/cilium -- bash`,

squeed:
Brilliant idea :-). Just one wording suggestion.

internal/cli/cmd/cmd.go
squeed, line 91: Wording suggestion:
"This command is only available in the Cilium agent.
Exec in to a running cilium pod to use it.
kubectl exec -it -n kube-system ds/cilium -- bash`,

squeed (changes requested):

internal/cli/cmd/cmd.go
```diff
@@ -84,5 +84,14 @@ cilium connectivity test`,
 	cmd.SetOut(os.Stdout)
 	cmd.SetErr(os.Stderr)
 
+	cmd.AddCommand(&cobra.Command{
+		Use:     "bpf",
+		Aliases: []string{"cleanup", "debuginfo", "endpoint", "fqdn", "identity", "ip", "kvstore", "lrp", "map", "metrics", "monitor", "node", "policy", "prefilter", "preflight", "recorder", "service"},
```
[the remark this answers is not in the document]
kranurag7: I agree, if there's a `-o list` or `--output list` flag that will list all the commands of per node cilium CLI then we can use that.

joestringer:
@tklauser @kranurag7 can we help move this PR forward? What's the next step?

kranurag7:
@joestringer thanks for https://github.com/cilium/cilium/pull/28085 👍🏻 With that PR being merged, I think we can close this one. I'll still wait for any comments that need to be addressed w.r.t. this PR, but if not, then please feel free to close this pull request.

joestringer:
I think that this proposal may still provide some value, and potentially fits in future with the proposal from https://github.com/cilium/cilium-cli/issues/1930 . I'll defer to @tklauser and @michi-covalent whether they think it's worth trying to revisit this PR, rebase & get it into the tree.

As of now, the CI is reporting failures but I guess that's because it was from a while ago so if we wanted to merge this then it at least would need a rebase against the tip of the tree.  I'm not sure what the solution is for the help issue in the thread above, but I think that's a proposal for an extra change on top? We can always iterate and discuss on that in a subsequent follow-up PR.

squeed:
@kranurag7 It's been quite some time that this PR has sat; if it's still open in a week it's probably time to close.

Especially now that we've renamed cilium.

kranurag7:
Hey @squeed 
I was waiting on other maintainers Joe maintained to have a look again. 
I think given the binary is already renamed it's safe to close this. 

I'll leave it upto you unless you or others have some comment left here. 
Thank you. :)

joestringer:
I think there's been plenty of time for the other folks to chime in and there's no sign of interest. Regardless, thank you @kranurag7 for your efforts investigating this and making the proposal.

PR #1285, opened 2022-12-09
README, dependabot: mark v0.10 branch as unmaintained

tklauser:
With the upcoming Cilium v1.13 release, Cilium v1.10 will no longer be maintained per the Cilium release policy [1]. cilium-cli v0.10 is only maintained for Cilium v1.10 compatibility, so mark it as unmaintained too.

[1] https://docs.cilium.io/en/v1.12/contributing/release/organization/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,30 +24,10 @@ updates:
     labels:
     - kind/enhancement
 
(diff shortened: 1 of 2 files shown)


michi-covalent (approved):
the end of an era

PR #1286, opened 2022-12-09
install: update existing roles, bindings if already extant

squeed:
There are certain cases where we fail to clean up Roles and RoleBindings on uninstall. So, if we encounter them on reinstall, update them instead.


tklauser (approved):
Thanks

PR #1299, opened 2022-12-14
k8s: Bring back TTY to Exec

brb:
This is a partial revert of commit 30310b18 ("connectivity, k8s, internal/utils: drop ExecInPodWithTTY and CtrlCReader").

Unfortunately, using the same context as StreamWithContext() to indicate a cancellation of a program running via Exec doesn't work as expected. From quick experimentation, it only closes a connection which doesn't necessarily send SIGTERM to a remote program. As a result, the tcpdump kept dangling in the encryption tests.

Bring back the ctrlreader, and use a separate context for it.

Signed-off-by: Martynas Pumputis <m@lambda.lt>

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -403,13 +403,17 @@ func (c *Client) ExecInPod(ctx context.Context, namespace, pod, container string
 	return result.Stdout, nil
 }
 
-func (c *Client) ExecInPodWithWriters(ctx context.Context, namespace, pod, container string, command []string, stdout, stderr io.Writer) error {
-	err := c.execInPodWithWriters(ctx, ExecParameters{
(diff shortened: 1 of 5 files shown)


brb:
Successful run - https://github.com/cilium/cilium/actions/runs/3693474477/jobs/6253507992.

k8s/client.go
```diff
@@ -385,6 +385,18 @@ func (c *Client) ExecInPodWithStderr(ctx context.Context, namespace, pod, contai
...
+		Namespace: namespace,
+		Pod:       pod,
+		Container: container,
+		Command:   command,
+		TTY:       true,
+	})
+	// Using TTY (for context cancellation support) fuses stderr into stdout
+	return result.Stdout, err
+}
+
```
tklauser, line 399: This is not used anywhere, so I think we can drop it from the revert.

tklauser (approved):
One nit, otherwise LGTM.

Issue #1079, opened 2022-09-05
bug: cilium sysdump predates global `--namespace` flag


ernado:
That is, currently you should do:
```console
cilium sysdump --cilium-namespace=cni
```
But this probably should also work (but it fails, defaulting to `kube-system`):
```console
cilium --namespace=cni sysdump
```

ChrsMark:
@tklauser @qmonnet I have pushed a patch for this in https://github.com/cilium/cilium-cli/pull/1302. Raising this here too for awareness :).

PR #1302, opened 2022-12-18
Honor global namespace flag in sysdump command

ChrsMark:
As reported at https://github.com/cilium/cilium-cli/issues/1079 the global `--namespace` flag is not honored in `sysdump` command. 

This PR honors the global `--namespace` flag in `sysdump` commad by making `--namespace` flag equivalent to the `--cilium-namespace` flag.

If both flags are set, then the  `--namespace` global flag comes with priority.

Closes https://github.com/cilium/cilium-cli/issues/1079

Signed-off-by: ChrsMark <chrismarkou92@gmail.com>

### DIFF
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -29,6 +29,10 @@ func newCmdSysdump() *cobra.Command {
 		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Honor --namespace global flag in case it is set and --cilium-namespace is not set
+			if sysdumpOptions.CiliumNamespace == "" && cmd.Flags().Changed("namespace") {
+				sysdumpOptions.CiliumNamespace = namespace
+			}
 			// Silence klog to avoid displaying "throttling" messages - those are expected.
 			klog.SetOutput(io.Discard)
 			// Collect the sysdump.
@@ -54,7 +58,7 @@ func initSysdumpFlags(cmd *cobra.Command, options *sysdump.Options, optionPrefix
 		"The labels used to target Cilium pods")
 	cmd.Flags().StringVar(&options.CiliumNamespace,
 		optionPrefix+"cilium-namespace", "",
-		"The namespace Cilium is running in")
+		"The namespace Cilium is running in. If not provided then the --namespace global flag is used (if provided)")
 	cmd.Flags().StringVar(&options.CiliumOperatorNamespace,
 		optionPrefix+"cilium-operator-namespace", "",
 		"The namespace Cilium operator is running in")
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -161,6 +161,8 @@ func NewCollector(k KubernetesClient, o Options, startTime time.Time, cliVersion
 		}
 		c.log("🔮 Detected Cilium installation in namespace %q", ns)
 		c.Options.CiliumNamespace = ns
+	} else {
+		c.log("ℹ️  Cilium namespace: %s", c.Options.CiliumNamespace)
 	}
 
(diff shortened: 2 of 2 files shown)


ChrsMark:
Hey @joamaki! Any ideas if the failing test is related? I cannot understand a lot of things from the output 😞

internal/cli/cmd/sysdump.go
```diff
@@ -29,6 +29,7 @@ func newCmdSysdump() *cobra.Command {
 		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			sysdumpOptions.CiliumNamespace = namespace
```
tklauser, line 32: > If both flags are set, then the `--namespace` global flag comes with priority.

I don't think this change works as expected. When specifying only `--cilium-namespace`, the given namespace is no longer used but instead the default for `--namespace` (i.e. `kube-system`) is used.

tklauser (changes requested):

internal/cli/cmd/sysdump.go
ChrsMark: You are right! 

How about following an order like:

1. if `-cilium-namespace` is provided then use it
2. if `-cilium-namespace` is not provided, then use the `--namespace` (if provided) instead 
3. if both are provided `-cilium-namespace` comes with priority
4. if none is provided then we fall back to the default of `--namespace` which is `kube-system`.

Would that work?

tklauser: Sounds good to me, thanks @ChrsMark. We should probably also document this order of precedence in these flags' help messages.

ChrsMark:
Hey @tklauser, I tried to address the comment(s) here so feel free to have another look when you get the chance.

internal/cli/cmd/sysdump.go
```diff
@@ -29,6 +29,9 @@ func newCmdSysdump() *cobra.Command {
 		Short: "Collects information required to troubleshoot issues with Cilium and Hubble",
 		Long:  ``,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if sysdumpOptions.CiliumNamespace == "" {
+				sysdumpOptions.CiliumNamespace = namespace
```
tklauser, line 33: I think we only want to set this if the `--namespace` flag was explicitly specified on the command line. Currently, it is set to `kube-system` by default if `--namespace` is not specified: https://github.com/cilium/cilium-cli/blob/d6447fb5055448eb73776b1a0dc08b83c74506ea/internal/cli/cmd/cmd.go#L69

Otherwise we'd break the auto-detection logic here: https://github.com/cilium/cilium-cli/blob/d6447fb5055448eb73776b1a0dc08b83c74506ea/sysdump/sysdump.go#L155-L162

I think we could probably use https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Changed to check whether the `--namespace` flag was explicitly set and only override `sysdumpOptions.CiliumNamespace` in case it was changed.

tklauser (changes requested):
Thanks for your patience @ChrsMark. I think this is still not completely correct as it would now break the namespace auto-detection, please see my comment inline.

ChrsMark:
Hey @tklauser and thanks for the review again! I agree, what your proposed here looks more robust. 
Summarizing on my own benefit with some examples:

1. `./cilium --namespace foo sysdump --cilium-namespace bar` -> namespace value will be `bar` 
2. `./cilium sysdump --cilium-namespace bar` -> namespace value will be `bar`
3. `./cilium --namespace foo sysdump ` -> namespace value will be `foo`
4. `./cilium sysdump` -> nothing will be set and the auto-detection logic will take place.

Based on this, feel free to have another look and let me know what you think :).

tklauser (approved):
Looks good to me, thanks a lot @ChrsMark.

PR #1304, opened 2022-12-20
connectivity: Fix iface derivation in encrypt tests

brb:
This commit fixes two issues:

* grep's "-P" is not available on busybox.
* In the direct routing mode, "ip r g $DST_POD_IP from $SRC_POD_IP" fails with "RTNETLINK answers: Network unreachable".

Fixes: 998ef8ab15e ("connectivity: Add encryption test")
Signed-off-by: Martynas Pumputis <m@lambda.lt>

### DIFF
--- a/connectivity/tests/encryption.go
+++ b/connectivity/tests/encryption.go
@@ -59,8 +59,8 @@ func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
 		iface = "cilium_" + tunnelFeat.Mode // E.g. cilium_vxlan
 	} else {
 		cmd := []string{"/bin/sh", "-c",
-			fmt.Sprintf("ip -o r g %s from %s | grep -oP '(?<=dev )[^ ]+'",
-				server.Pod.Status.PodIP, client.Pod.Status.PodIP)}
+			fmt.Sprintf("ip -o r g %s | grep -oE 'dev [^ ]*' | cut -d' ' -f2",
+				server.Pod.Status.PodIP)}
 		t.Debugf("Running %s", strings.Join(cmd, " "))
 		dev, err := clientHost.K8sClient.ExecInPod(ctx, clientHost.Pod.Namespace,
 			clientHost.Pod.Name, "", cmd)


connectivity/tests/encryption.go
```diff
@@ -59,8 +59,8 @@ func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
 		iface = "cilium_" + tunnelFeat.Mode // E.g. cilium_vxlan
 	} else {
 		cmd := []string{"/bin/sh", "-c",
-			fmt.Sprintf("ip -o r g %s from %s | grep -oP '(?<=dev )[^ ]+'",
-				server.Pod.Status.PodIP, client.Pod.Status.PodIP)}
```
pchaigno, line 63: How did the tests pass?

pchaigno: > In the direct routing mode, "ip r g $DST_POD_IP from $SRC_POD_IP"
  fails with "RTNETLINK answers: Network unreachable".

Do you know why? The revert means it may break on EKS.

brb: > How did the tests pass?

The direct routing was never tested. However, it's going to change in https://github.com/cilium/cilium/pull/22803.

> Do you know why? The revert means it may break on EKS.

My initial thought was that the rp_filter is to blame, but setting it to 0 didn't resolve the issue. For now my priority is to get many different DP configurations tested "on prem" (=Kind on LVH VMs). Once it's been done - managed K8s. However, it doesn't mean that others cannot pick it meanwhile.

pchaigno: Got it. Might be worth creating an issue for that on this repository, just so we don't waste time when we start running this test on EKS in > 6months.

pchaigno (approved):

joamaki (approved):

brb:
The CI with the changes passed - https://github.com/cilium/cilium/actions/runs/3739715885/jobs/6347256459.

connectivity/tests/encryption.go
brb: > Might be worth creating an issue for that on this repository

Yep, I'm going to do in a bit.

brb:
Multicluster hit the infamous 1.1.1.1 flake.

connectivity/tests/encryption.go
brb: Just for posterity - https://github.com/cilium/cilium-cli/issues/1306.

PR #1315, opened 2023-01-05
sysdump: Collect new CR introduced as part of 1.13

sayboras:
This commit is to make sure that we have enough details for debugging
and troubleshooting.


nathanjsweet (approved):

sysdump/sysdump.go
```diff
@@ -424,20 +424,6 @@ func (c *Collector) Run() error {
...
-				v, err := c.Client.ListCiliumEgressNATPolicies(ctx, metav1.ListOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to collect Cilium egress NAT policies: %w", err)
-				}
-				if err := c.WriteYAML(ciliumEgressNATPoliciesFileName, v); err != nil {
-					return fmt.Errorf("failed to collect Cilium egress NAT policies: %w", err)
-				}
-				return nil
-			},
-		},
```
tklauser, line 440: I think we still want to collect `CiliumEgressNATPolicies` when running `cilium sysdump` against Cilium v1.12 or earlier.

tklauser (changes requested):

sysdump/sysdump.go
sayboras: the generated client was also removed from cilium v1.13.x, so if we want to to the above, we might need to use other kind of client (e.g. dynamic client), what do you think ?

tklauser: Sounds good. Or just copy the generated client to cilium-cli code base - whatever is easier to implement.

sayboras: thanks, let me give it a crack and get back to you.

tklauser (approved):

PR #1320, opened 2023-01-10
Increase copy retries

nebril:
If a user wants to get bugtool output while they are on unreliable network, it makes sense to increase retry count to make their experience better.


sayboras (approved):
nit: I didn't think about having CLI flag for maximum number of retries before, WDYT?

nebril:
@sayboras I think it would make sense if it would allow the user to set it to infinite to make sure that the data gets downloaded properly. I will add it to the PR.

tommyp1ckles (approved):

Issue #1314, opened 2023-01-04
Cancelling failing cilium-health probes results in a tight loop which spams the terminal.


bimmlerd:
## Bug report

Persistenly failing cilium-health probes are in a loop until some general timeout: https://github.com/cilium/cilium-cli/blob/30310b18ae7751beb43744691c2e68c589b65861/connectivity/tests/health.go#L43-L65

Typical log would be
```
⚠️  cilium-health validation failed: "cilium-agent 'kube-system/cilium-wgcls': connectivity to path 'kind-worker.health-endpoint.primary-address.http.status' is unhealthy: 'Get \"http://10.0.0.12:4240/hello\": read tcp 10.0.1.135:50650->10.0.0.12:4240: read: connection reset by peer'", retrying...
```

As a result, one might want to stop the health check prematurely, e.g. with a Ctrl+C. This seems to cancel `ctx`, upon which https://github.com/cilium/cilium-cli/blob/30310b18ae7751beb43744691c2e68c589b65861/connectivity/tests/health.go#L46-L50 fails with 

```
  cilium-health probe failed: "error sending request: Post \"https://127.0.0.1:53497/api/v1/namespaces/kube-system/pods/cilium-wgcls/exec?command=cilium-health&command=status&command=--probe&command=-o%3Djson&stderr=true&stdout=true\": dial tcp 127.0.0.1:53497: operation was canceled", stdout: {"" '\x00' '\x00'}, retrying...
```

Because of the `continue`, this just leads to a tight loop, spamming the above warning into the terminal at a high rate.

**General Information**

- Cilium CLI version (run `cilium version`)
```
cilium-cli: v0.12.11-40-g2fc0835f compiled with go1.19.4 on darwin/arm64
cilium image (default): v1.12.4
cilium image (stable): v1.12.5
cilium image (running): v1.12.5
```
- Orchestration system version in use (e.g. `kubectl version`, ...)
```
Client Version: v1.26.0
Kustomize Version: v4.5.7
Server Version: v1.25.3
```
- Platform / infrastructure information: Kind

**How to reproduce the issue**

~I'm not sure why the health probes ended up failing in the first place, but once in that state, it's easy 😅~

1. Start kind _without_ specifying `disableDefaultCNI: true`, but with multiple nodes. The following works with `kind create cluster --config ...`:
```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
- role: worker
[...]

PR #1353, opened 2023-01-25
Check if context is cancelled before next iteration

dharmicksai:
When the health check fails, the test keeps on retrying.
If the user tries to stop the test prematurely by sending a SIGINT signal,
the pod.K8sClient.ExecInPod function returns an error, as the context is canceled.
due to this, the test would get stuck in an infinite loop.

Now the context is checked for cancellation before continuing to the next loop

Fixes: #1314

Signed-off-by: K Dharmick Sai <dharmicksaik@gmail.com>

### DIFF
--- a/connectivity/tests/health.go
+++ b/connectivity/tests/health.go
@@ -47,15 +47,13 @@ func runHealthProbe(ctx context.Context, t *check.ConnectivityTest, pod *check.P
 		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, defaults.AgentContainerName, cmd)
 		if err != nil {
 			t.Warnf("cilium-health probe failed: %q, stdout: %q, retrying...", err, stdout)
-			continue
+		} else {
+			err = validateHealthStatus(t, pod, stdout)
+			if err == nil {
+				return
+			}
+			t.Warnf("cilium-health validation failed: %q, retrying...", err)
 		}
-
-		err = validateHealthStatus(t, pod, stdout)
-		if err == nil {
-			return
-		}
-		t.Warnf("cilium-health validation failed: %q, retrying...", err)
-
 		// Wait until it's time to retry or context is cancelled.
 		select {
 		case <-done:


connectivity/tests/health.go
```diff
@@ -45,8 +45,13 @@ func runHealthProbe(ctx context.Context, t *check.ConnectivityTest, pod *check.P
 
 		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, "", cmd)
 		if err != nil {
-			t.Warnf("cilium-health probe failed: %q, stdout: %q, retrying...", err, stdout)
-			continue
+			select {
+			case <-done:
```
bimmlerd, line 49: this fixes the immediate bug regarding ctx cancellation through Ctrl+C, but as mentioned in https://github.com/cilium/cilium-cli/issues/1314#issuecomment-1376929199 we should also check the retryTimer here.

bimmlerd (changes requested):
Thanks for your contribution!
Approach looks good, just need to also check the retryTimer.

connectivity/tests/health.go
dharmicksai: Oh, I will update the pr

dharmicksai:
Multicluster / installation-and-connectivity (pull_request_target)  check is failing with the following error messages:

`❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 5 --max-time 20 http://echo-other-node:8080" failed: command terminated with exit code 28`
`❌ Aborting flow matching: context deadline exceeded` 
`❌ Flow validation failed for peer cilium-test/client-755fb678bd-wzxzh: 1 failures (first: 0, last: 8, matched: 2)`

The errors are being logged in the [ValidateFlows function](https://github.com/cilium/cilium-cli/blob/v0.12.12/connectivity/check/action.go#L825)
 How can these errors be resolved?

connectivity/tests/health.go
```diff
@@ -45,21 +45,14 @@ func runHealthProbe(ctx context.Context, t *check.ConnectivityTest, pod *check.P
 
 		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, "", cmd)
 		if err != nil {
-			select {
-			case <-done:
+			t.Warnf("cilium-health probe failed: %q, stdout: %q, retrying...", err, stdout)
+		}else{
```
bimmlerd, line 49: nit: looks like you didn't run gofmt, please do.

bimmlerd (changes requested):
Changes LGTM (except for gofmt nit). To get it merged, please squash the two commits and add a brief commit message describing the bug this fixes.

Regarding the CI, seems like a flake. Squashing the commits will retrigger it anyway, let's see whether it works ok then.

connectivity/tests/health.go
dharmicksai: I ran gofmt and squashed the commits

bimmlerd:
Change now lgtm. CI bombed pretty hard, but I'll retrigger those which I can see are unrelated:
- rerunning EKS (tunnel): https://github.com/cilium/cilium-cli/actions/runs/4044003310/jobs/6953620920#step:12:34 (something about helm.cilium.io)
- rerunning EKS (ENI): https://github.com/cilium/cilium-cli/actions/runs/4044003304/jobs/6953620932#step:12:33 (something about helm.cilium.io)\
- the others failed with "no agent available" which also sounds like an infra issue.

will approve once CI is happier.

bimmlerd (approved):
Had to rebase on latest master to get CI to pass.
(Seems like you rebased on `master` instead of `origin/master` without pulling `origin/master` into your local `master` first, thus resetting the commits back a couple of weeks, apparently breaking CI)
Thanks for sticking with this, looking forward to more contributions :)

dharmicksai:
Thank you @bimmlerd, for guiding me through my first issue.

PR #1354, opened 2023-01-25
[connectivity] add container name for health tests

fl64:
In some environments it is possible to use ds cilium with multiple containers (e.g. kube-rbac-proxy) in these cases the health tests do not work correctly.
For example:

```
⚠️ cilium-health probe failed: "a container name must be specified for pod agent-gtrg7, choose one of: [check-linux-kernel mount-cgroup clean-cilium-state cilium-agent kube-rbac-proxy]", stdout: {" '\x00' '\x00'}, retrying...
```

### DIFF
--- a/connectivity/tests/health.go
+++ b/connectivity/tests/health.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/util/jsonpath"
 
 	"github.com/cilium/cilium-cli/connectivity/check"
+	"github.com/cilium/cilium-cli/defaults"
 )
 
 func CiliumHealth() check.Scenario {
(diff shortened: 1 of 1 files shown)


connectivity/tests/health.go
```diff
@@ -43,7 +43,7 @@ func runHealthProbe(ctx context.Context, t *check.ConnectivityTest, pod *check.P
 	for {
 		retryTimer := time.After(time.Second)
 
-		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, "", cmd)
+		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, "cilium-agent", cmd)
```
tklauser, line 46: Let's use the pre-defined const for the container name here:

```suggestion
		stdout, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, defaults.AgentContainerName, cmd)
```

tklauser (changes requested):

tklauser:
@fl64 You'll need to import the defaults package as well, it seems. Please also squash the commits into one.

fl64:
@tklauser @tk Sorry, my fault, I missed it.

tklauser (approved):
Thanks

PR #1362, opened 2023-02-01
clustermesh: Add missing brackets of etcd option

giorio94:
When clustermesh-apiserver Pod is deployed on the IPv6 single-stack cluster, etcd fails to startup with the error like this.

```
invalid value "https://127.0.0.1:2379,https://a:1:0:3::fc75:2379" for flag -listen-client-urls: URL address does not have the form "host:port": https://a:1:0:3::fc75:2379
```

This happens because we don't put brackets around the IPv6 address. Fix deployment template to correctly handle that.

Related: cilium/cilium#22962

Signed-off-by: Marco Iorio <marco.iorio@isovalent.com>


YutaroHayakawa (approved):
Good catch! LGTM 👍

PR #1369, opened 2023-02-03
ci: use cloud provider specific external target for pod-to-world tests

tklauser:
Follows https://github.com/cilium/cilium/pull/23103, quoting its commit message:

> With the configurable --external-target flag we are configuring different targets per cloud provider. Targets were chosen to be the least hops from the clusters. They were determined by experimenting with connection times from cloud providers. Splitting targets also has the advantage of not overloading a single target.

/cc @brlbil

### DIFF
--- a/.github/in-cluster-test-scripts/aks.sh
+++ b/.github/in-cluster-test-scripts/aks.sh
@@ -16,7 +16,7 @@ sleep 10s
 [[ $(pgrep -f "cilium.*hubble.*port-forward|kubectl.*port-forward.*hubble-relay" | wc -l) == 2 ]]
 
 # Run connectivity test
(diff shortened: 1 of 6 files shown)


brlbil (commented):
Just curious, is there a reason --external-target is not set in `external-workloads.sh`?
If it is not set, it will default to `one.one.one.one`.
@tklauser

tklauser:
> Just curious, is there a reason --external-target is not set in `external-workloads.sh`? If it is not set, it will default to `one.one.one.one`. @tklauser

No reason other than me forgetting to update that workflow :wink: thanks for spotting it, now updated.

brlbil (approved):
LGTM

PR #1392, opened 2023-02-15
Wait for deletion of agent DaemonSet before deleting other resources.

marseel:
Wait for the deletion of agent DaemonSet before deleting other resources.
This allows the agent to clean up cilium annotations added to the node.

Related PR: https://github.com/cilium/cilium/pull/23772

### DIFF
--- a/install/uninstall.go
+++ b/install/uninstall.go
@@ -48,12 +48,23 @@ func (k *K8sUninstaller) Log(format string, a ...interface{}) {
 func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
 	k.autodetect(ctx)
 
+	k.Log("🔥 Deleting agent DaemonSet...")
+	k.client.DeleteDaemonSet(ctx, k.params.Namespace, defaults.AgentDaemonSetName, metav1.DeleteOptions{})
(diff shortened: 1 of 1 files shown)


install/uninstall.go
```diff
@@ -94,25 +99,25 @@ func (k *K8sUninstaller) Uninstall(ctx context.Context) error {
...
 
-	retry:
+func (k *K8sUninstaller) waitForPodsToBeDeleted(ctx context.Context) error {
+	for {
 		pods, err := k.client.ListPods(ctx, k.params.Namespace, metav1.ListOptions{LabelSelector: defaults.AgentPodSelector})
 		if err != nil {
 			return err
 		}
 
 		if len(pods.Items) > 0 {
```
squeed, line 117: unrelated nit: if the context is cancelled while we're sleeping, we won't notice it. Can we fix that?

squeed (approved):
Looks good, but there's one more cleanup we might as well do while we're here.

marseel:
PTAL, I've also added some logging when error happens or when we successfully delete daemonset.

ldelossa (approved):
nice add :100:

PR #1399, opened 2023-02-16
defaults: Lower connectivity timeouts

brb:
The timeouts are extensively used in connectivity tests which checks whether a policy has denied a connection. Obviously, this leads to waiting 20s.

Before (with the upcoming IPv6 connectivity tests):

    ./cilium connectivity test  6.99s user 1.42s system 1% cpu 11:48.72
    total

After:

    ./cilium connectivity test  7.06s user 1.40s system 2% cpu 6:14.97
    total

### DIFF
--- a/defaults/defaults.go
+++ b/defaults/defaults.go
@@ -85,8 +85,8 @@ const (
 
 	PolicyWaitTimeout = 15 * time.Second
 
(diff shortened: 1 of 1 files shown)


tklauser (approved):
:rocket:

PR #1402, opened 2023-02-17
fix(connectivity): add dns rules and change protocol to ANY

eminaktas:
Connectivity tests fail with the `node-local-dns` environment. This PR adds missing `DNS` rules and changes protocol from `UDP` to `ANY`

### DIFF
--- a/connectivity/manifests/allow-all-except-world.yaml
+++ b/connectivity/manifests/allow-all-except-world.yaml
@@ -19,9 +19,11 @@ spec:
   # This change prevents failing the connectivity
   # test for such environments.
   - toPorts:
-      - ports:
-          - port: "53"
-            protocol: UDP
+    - ports:
+      - port: "53"
+        protocol: UDP
+      - port: "53"
+        protocol: TCP
     toEntities:
       - world
   ingress:
--- a/connectivity/manifests/client-egress-l7-tls.yaml
+++ b/connectivity/manifests/client-egress-l7-tls.yaml
@@ -12,7 +12,9 @@ specs:
   - toPorts:
     - ports:
       - port: "53"
-        protocol: ANY
+        protocol: UDP
+      - port: "53"
+        protocol: TCP
       rules:
         dns:
           - matchPattern: "*"
--- a/connectivity/manifests/client-egress-only-dns.yaml
+++ b/connectivity/manifests/client-egress-only-dns.yaml
@@ -10,7 +10,9 @@ spec:
   - toPorts:
     - ports:
       - port: "53"
-        protocol: ANY
+        protocol: UDP
+      - port: "53"
+        protocol: TCP
       rules:
         dns:
         - matchPattern: "*"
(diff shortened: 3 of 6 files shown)


connectivity/manifests/allow-all-except-world.yaml
```diff
@@ -19,9 +19,9 @@ spec:
   # This change prevents failing the connectivity
   # test for such environments.
   - toPorts:
-      - ports:
-          - port: "53"
-            protocol: UDP
+    - ports:
+      - port: "53"
+        protocol: ANY
```
aditighag, line 24: Why are we opening the policy rules for all protocols if the intention is to only allow DNS traffic? Did you try adding a similar rule for TCP?

aditighag (changes requested):
See the inline comment.

connectivity/manifests/allow-all-except-world.yaml
eminaktas: Thanks for asking. I should have mentioned it already. I was searching for the root cause. I did a test with TCP it didn't affect the result. The problem is that the policies with `FQDNS` do not work. Of course, I am talking about the environment with the `node-local-dns` It looks like it cannot detect the protocol when only the `client-egress-to-fqdns-one-one-one-one` or the `client-egress-only-dns` policies are applied.

I added the DNS rule (the part below) for the above policies. But the problem wasn't that something was missing.

```yaml
      rules:
        dns:
        - matchPattern: "*"
```

Then, I noticed that Cilium couldn't validate the protocols.

We might add a test environment with `node-local-dns` deployed. I can help with it.

I noticed this problem while I was testing the new release of Cilium.

joestringer (approved):
Considering the Cilium policy behaviour for how it handles the node local DNS IP and the way that these manifests are currently working around the issue, I think that this proposal is accurate.

Switching from `UDP` to `ANY` is also effectively an alias for TCP and UDP anyway, and DNS can commonly go over both UDP and TCP.

connectivity/manifests/allow-all-except-world.yaml
joestringer: Worth noting that if this is set to `ANY` and there is a `rules` stanza, Cilium only supports redirecting traffic for TCP and UDP so I believe this is effectively the same as having another rule for TCP specifically.

tklauser (approved):

aditighag (approved):

aditighag:
Sorry for the delay.

aditighag:
Test failures - 
>AKS: no-policies/pod-to-world/https-to-bing.com-1: cilium-test/client2-5c6c769648-7zd4n (10.0.1.164) -> bing.com-https (bing.com:443)

We have similar flakes every now and then where tests involve connecting to an external entity. 

I've restarted the external workloads test as there were some infrastructure failures in the earlier run.

squeed:
This flaked on #2070 -- marking ready to merge.

PR #1404, opened 2023-02-20
ingress: Create Ingress Service if required

sayboras:
This is to make sure that the shared ingress service can be created if
required.

Reported-by: Xiao Song


michi-covalent (approved):
let's roll

PR #1414, opened 2023-02-24
connectivity tests: fix issue on IPv6-only clusters

giorio94:
This PR changes the behavior of `IPFamilyNone` (now renamed to `IPFamilyAny` for consistency) to allow both IPv4 and IPv6 addresses, and switches the "waitForPodDNS" check to use it is order to prevent issues on IPv6-only clusters.


brb (approved):
Thanks!

squeed (approved):

PR #1418, opened 2023-02-27
uninstall: set "cni.uninstall" to true on uninstall

squeed:
There is a Helm variable, `cni.uninstall`, that tells the agent to clean up its CNI configuration file and plugin binaries. This behavior was disabled by default in PR https://github.com/cilium/cilium/pull/24009. However, when we *know* we're uninstalling Cilium, we should re-enable the old behavior.

So, manually touch the key file in the configmap to enable uninstall. Then, the pre-stop hook will be activated on shutdown and clean up as usual.


rolinh (approved):

install/uninstall.go
```diff
@@ -141,3 +145,29 @@ func (k *K8sUninstaller) waitForPodsToBeDeleted(ctx context.Context) error {
...
+func (k *K8sUninstaller) enableCNIUninstall(ctx context.Context) {
+	pods, err := k.client.ListPods(ctx, k.params.Namespace, metav1.ListOptions{LabelSelector: defaults.AgentPodSelector})
+	if err != nil {
+		k.Log("❌ Failed to enable cni cleanup: %v", err)
+		return
+	}
+	wp := workerpool.New(k.params.WorkerCount)
+
+	for i := range pods.Items {
+		pod := &(pods.Items[i]) // go iterator silliness; can't use range variable
```
kaworu, line 158: How about something like
```suggestion
	for _, pod := range pods.Items {
		pod := pod // capture range variable
```

install/uninstall.go
```diff
@@ -141,3 +145,29 @@ func (k *K8sUninstaller) waitForPodsToBeDeleted(ctx context.Context) error {
...
 	}
 }
+
+func (k *K8sUninstaller) enableCNIUninstall(ctx context.Context) {
+	pods, err := k.client.ListPods(ctx, k.params.Namespace, metav1.ListOptions{LabelSelector: defaults.AgentPodSelector})
+	if err != nil {
+		k.Log("❌ Failed to enable cni cleanup: %v", err)
+		return
+	}
+	wp := workerpool.New(k.params.WorkerCount)
```
kaworu, line 155: We need to shutdown the workerpool eventually to avoid leaking its "run" goroutine. Also we can provide the base context since we have been given one
```suggestion
	wp := workerpool.NewWithContext(ctx, k.params.WorkerCount)
	defer wp.Close()
```

kaworu (changes requested):
Thanks @squeed! We need to call `wp.Close()` but other than that and some nitpicking LGTM.

install/uninstall.go
squeed: oh right, that's the trick for that.

install/uninstall.go
squeed: aha, good point!

squeed:
@kaworu thanks for the review, updated!

kaworu (approved):
Thanks!

PR #1419, opened 2023-02-27
enabled l7Proxy & removed err.log message for WireGuard

### LINKED ISSUE #1405: install: Don't disable L7 proxy when installing with WireGuard enabled
Since https://github.com/cilium/cilium/pull/19401 (to be released in v1.14) we no longer need to disabled the L7 proxy when WireGuard is enabled on the > v1.13 Cilium (the relevant log message `L7 proxy disabled due to Wireguard encryption`).

As a bonus, we can remove the `Wireguard does not support node encryption yet` warning too.

Neilblaze:
Since https://github.com/cilium/cilium/pull/19401 (to be released in v1.14) there's no need to **_disable_** the `L7 proxy` when WireGuard is **_enabled_** on the > v1.13 Cilium (the relevant log message L7 proxy disabled due to Wireguard encryption).

> As a bonus, the unnecessary log/WARN `Wireguard does not support node encryption yet` has also been removed.

Fixes #1405 
cc: @brb 
Signed-off-by: [Pratyay Banerjee](mailto:putubanerjee23@gmail.com)

### DIFF
--- a/install/helm.go
+++ b/install/helm.go
@@ -125,11 +125,13 @@ func (k *K8sInstaller) generateManifests(ctx context.Context) error {
 			// TODO(gandro): Future versions of Cilium will remove the following
 			// two limitations, we will need to have set the config map values
 			// based on the installed Cilium version
-			helmMapOpts["l7Proxy"] = "false"
-			k.Log("ℹ️  L7 proxy disabled due to Wireguard encryption")
+			if versioncheck.MustCompile("<=1.13.0")(k.chartVersion) {
+				helmMapOpts["l7Proxy"] = "false"
+				k.Log("ℹ️  L7 proxy disabled due to Wireguard encryption")
 
-			if k.params.NodeEncryption {
-				k.Log("⚠️️  Wireguard does not support node encryption yet")
+				if k.params.NodeEncryption {
+					k.Log("⚠️️  Wireguard does not support node encryption yet")
+				}
 			}
 		}
 


install/helm.go
```diff
@@ -125,12 +125,8 @@ func (k *K8sInstaller) generateManifests(ctx context.Context) error {
 			// TODO(gandro): Future versions of Cilium will remove the following
 			// two limitations, we will need to have set the config map values
 			// based on the installed Cilium version
-			helmMapOpts["l7Proxy"] = "false"
-			k.Log("ℹ️  L7 proxy disabled due to Wireguard encryption")
-
-			if k.params.NodeEncryption {
-				k.Log("⚠️️  Wireguard does not support node encryption yet")
-			}
+			helmMapOpts["l7Proxy"] = "true"
```
brb, line 128: I think we don't need to set any `l7Proxy` here, as it should be enabled by default.

However, we need to figure out which version of Cilium we are trying to install. I.e., we need to keep the existing code for <1.14 versions.

brb (changes requested):
Thanks! Some comment.

install/helm.go
Neilblaze: @brb Thanks for the review. That means I just need to implement a simple `versioncheck ("<=1.13.0")` to restrict, right? 👇🏼 

```go
switch k.params.Encryption {
case encryptionIPsec:
	helmMapOpts["encryption.enabled"] = "true"
	helmMapOpts["encryption.type"] = "ipsec"
	if k.params.NodeEncryption {
		helmMapOpts["encryption.nodeEncryption"] = "true"
	}
case encryptionWireguard:
	helmMapOpts["encryption.enabled"] = "true"
	helmMapOpts["encryption.type"] = "wireguard"
	// TODO(gandro): Future versions of Cilium will remove the following
	// two limitations, we will need to have set the config map values
	// based on the installed Cilium version
	versioncheck.MustCompile("<=1.13.0")(k.chartVersion):
		helmMapOpts["l7Proxy"] = "false"
		k.Log("ℹ️  L7 proxy disabled due to Wireguard encryption")

		if k.params.NodeEncryption {
			k.Log("⚠️️  Wireguard does not support node encryption yet")
		}
}
```

brb: Yep :+1:

brb (changes requested):
Thanks! Could you squash all commit messages into a single one? Also, please no emojis in commit messages :sweat_smile:

Neilblaze:
@brb done ✔️

brb (changes requested):
Close to merging. One more nit - could you remove the duplicates from your commit message?
```
enabled l7Proxy & removed err.log message for WireGuard

Signed-off-by: Pratyay Banerjee <putubanerjee23@gmail.com>

enabled l7Proxy & removed err.log message for WireGuard

Signed-off-by: Pratyay Banerjee <putubanerjee23@gmail.com>

enabled l7Proxy & removed err.log message for WireGuard

Signed-off-by: Pratyay Banerjee <putubanerjee23@gmail.com>
```

Neilblaze:
> One more nit - could you remove the duplicates from your commit message?

@brb Thanks for the catch. Everything should be fine now! 😄

brb (approved):

tklauser (approved):

ShiroDN:
There is a problem now, condition `versioncheck.MustCompile("<=1.13.0")` does not match patch versions of cilium ex. 0.13.x so now when I tried to install cilium with `--encryption wireguard` it keeps restarting with error:

`level=fatal msg="Wireguard (--enable-wireguard) is not compatible with L7 proxy (--enable-l7-proxy)" subsys=daemon`

Issue #1396, opened 2023-02-16
Seamless Cilium CLI + Helm


christarazi:
Allow Cilium CLI to work seamlessly with Helm by using Helm as the backend for installation.

There are a number of issues opened already which reference this, but no one issue to describe the entirety of the problem. This is an umbrella issue for

* https://github.com/cilium/cilium-cli/issues/515
* https://github.com/cilium/cilium-cli/issues/751
* https://github.com/cilium/cilium-cli/issues/1106

giorio94:
Related: https://github.com/cilium/cilium-cli/issues/1391

lambdanis:
Another issue that would benefit from this: #1422

Also, there were problems when creating 3rd party custom resources (e.g. #1107) as `helm template` commands run by Cilium CLI don't interact with k8s API. A proper Helm integration would solve it too.

xiaods:
any update on it?

lambdanis:
@xiaods Yes. The README documents the [helm installation mode](https://github.com/cilium/cilium-cli/tree/v0.15.8#helm-installation-mode), introduced in v0.14 and used by default since v0.15.

christarazi:
@xiaods With the latest CLI, does it resolve your problems?

xiaods:
> @xiaods With the latest CLI, does it resolve your problems?

not yet.

```
[root@ip-172-31-0-254 ~]# cilium version
cilium-cli: v0.15.7 compiled with go1.21.0 on linux/amd64
cilium image (default): v1.14.1
cilium image (stable): v1.14.2
cilium image (running): 1.14.1


[root@ip-172-31-0-254 ~]# cilium status
    /¯¯\
 /¯¯\__/¯¯\    Cilium:             OK
 \__/¯¯\__/    Operator:           OK
 /¯¯\__/¯¯\    Envoy DaemonSet:    disabled (using embedded mode)
 \__/¯¯\__/    Hubble Relay:       disabled
    \__/       ClusterMesh:        disabled

DaemonSet              cilium             Desired: 1, Ready: 1/1, Available: 1/1
Deployment             cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
Containers:            cilium             Running: 1
                       cilium-operator    Running: 1
Cluster Pods:          3/3 managed by Cilium
Helm chart version:    1.14.1
[...]

PR #1423, opened 2023-02-28
hubble: Patch Cilium DaemonSet with hubble-metrics port

### LINKED ISSUE #1422: `hubble enable` doesn't set hubble-metrics port in Cilium agent
## Bug report

Running a command enabling Hubble and Hubble metrics, e.g.:
```
cilium hubble enable --helm-set-string=hubble.metrics.enabled="{http,dns,drop,tcp,flow,icmp,port-distribution}"
```
results in a `hubble-metrics` service being created:
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    prometheus.io/port: "9965"
    prometheus.io/scrape: "true"
  labels:
    app.kubernetes.io/name: hubble
    app.kubernetes.io/part-of: cilium
    k8s-app: hubble
  name: hubble-metrics
  namespace: kube-system
spec:
  clusterIP: None
  ports:
  - name: hubble-metrics
    port: 9965
    protocol: TCP
    targetPort: hubble-metrics
  selector:
    k8s-app: cilium
  type: ClusterIP
```
but doesn't set `hubble-metrics` port in Cilium daemonset. That means that `hubble-metrics` Endpoints object doesn't get populated with Cilium pods IPs. Prometheus service discovery often relies on Endpoints objects, so Prometheus won't scrape Hubble metrics in that case.

If you enable Hubble metrics on `cilium install`:
```
cilium install --version="1.13.0" --helm-set-string=hubble.metrics.enabled="{http,dns,drop,tcp,flow,icmp,port-distribution}"
```
then `hubble-metrics` port is configured. But from the user perspective it makes sense to enable Hubble metrics when enabling Hubble:
```
cilium install
cilium hubble enable --helm-set-string=hubble.metrics.enabled="{http,dns,drop,tcp,flow,icmp,port-distribution}"
```
and then metrics are exposed, but `hubble-metrics` port is not configured.

A "proper" solution for this issue would be using Helm commands for all installs/updates (cilium/cilium-cli#1396). Meanwhile, a "hack" solution would be updating cilium daemonset in addition to other components on `hubble enable`.

**General Information**

- Cilium CLI version: v0.12.13
- Orchestration system version in use: k8s 1.24.7
- Platform / infrastructure information: Kind

```
cilium install --version="1.13.0"
cilium hubble enable --helm-set-string=hubble.metrics.enabled="{http,dns,drop,tcp,flow,icmp,port-distribution}"
```

lambdanis:
So far when enabling Hubble metrics in `hubble enable` command, e.g.:
```
cilium hubble enable --helm-set-string=hubble.metrics.enabled="{http,dns,drop,tcp,flow,icmp,port-distribution}"
```
the CLI created a hubble-metrics service, but didn't add a hubble-metrics port to Cilium daemonset. Then Hubble metrics were exposed by Cilium, but the hubble-metrics Endpoints object wasn't populated with the Cilium pod IPs. Prometheus scrape configs often rely on Endpoints objects, so this results in Hubble metrics not being scraped.

This patch fixes that by patching the Cilium agent daemonset with a hubble-metrics port if a metrics service is created in `hubble enable` command.

It's a rough patch and it assumes using the default Hubble metrics port (9965), and Cilium agent name ("cilium"). A proper solution to the problems would be using Helm as backend for all install/update commands (#1396).

Fixes #1422

### DIFF
--- a/hubble/hubble.go
+++ b/hubble/hubble.go
@@ -64,6 +64,7 @@ type k8sHubbleImplementation interface {
 	DeletePodCollection(ctx context.Context, namespace string, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error
 	ListPods(ctx context.Context, namespace string, options metav1.ListOptions) (*corev1.PodList, error)
 	GetDaemonSet(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*appsv1.DaemonSet, error)
+	PatchDaemonSet(ctx context.Context, namespace, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions) (*appsv1.DaemonSet, error)
 	CiliumStatus(ctx context.Context, namespace, pod string) (*models.StatusResponse, error)
 	ListCiliumEndpoints(ctx context.Context, namespace string, opts metav1.ListOptions) (*ciliumv2.CiliumEndpointList, error)
 	GetServerVersion() (*semver.Version, error)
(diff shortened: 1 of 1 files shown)


hubble/hubble.go
```diff
@@ -670,6 +671,25 @@ func (k *K8sHubble) Enable(ctx context.Context) error {
 		if _, err := k.client.CreateService(ctx, k.params.Namespace, metricsSvc, metav1.CreateOptions{}); err != nil {
 			return err
 		}
+
+		// Add hubble-metrics port to Cilium agent daemonset
+		k.Log("🔥 Patching Agent DaemonSet...")
+		patch := []byte(fmt.Sprintf(`{
```
chancez, line 677: Shouldn't we only do this if `hubble.metrics.enabled` is true?

lambdanis: It's inside the `if metricsSvc := k.generateMetricsService(); metricsSvc != nil` block - so it will happen only if `hubble-metrics` service is created, so only if `hubble.metrics.enabled` is non-empty. I think tying this patch together with the service creation makes sense.

chancez (approved):

PR #1424, opened 2023-03-01
Add profiling data(cpu/heap) of cilium agent to sysdump.

marseel:
Scrape cpu and heap profiles from the Cilium agents by default when gathering sysdump.


tklauser (approved):
Thanks

nathanjsweet (approved):

PR #1438, opened 2023-03-06
experiment: Don't pay attention to yaml files anymore

squeed:
Rather than referencing individual template files, set up install such that we pass around rendered objects and reference them by name and kind.

This means we don't have to worry about template structure changing when using cilium-cli.


squeed:
This was a fun experiment, but we should probably just use helm directly.

PR #1440, opened 2023-03-07
connectivity: Report IP families that are being tested

### LINKED ISSUE #1410: connectivity: Report which IP families are tested
We should improve the connectivity reporting saying which IP families (i.e., v4 or v6) are being tested. 

Also, each Action description could contain the family too. E.g., we could report `curl-v4-1`, `curl-v6-2`, etc.

Neilblaze:
The objective is to *improve the connectivity reporting* by displaying which *IP families* (i.e., v4 or v6) are being tested.
Additionally, each Action description should contain the family too. E.g., we could report `curl-v4-1`, `curl-v6-2`, etc.

Fixes #1410 
cc: @brb, @gandro 
Signed-off-by: [Pratyay Banerjee](mailto:putubanerjee23@gmail.com)


Neilblaze:
@brb Can you please provide some feedback on where I'm going wrong? 😅 

> Even one of the error log state this? → `could not import github.com/cilium/cilium-cli/connectivity/tests`

connectivity/tests/host.go
```diff
@@ -89,15 +96,25 @@ func (s *podToHostPort) Run(ctx context.Context, t *check.Test) {
 			echo := echo // copy to avoid memory aliasing when using reference
 
 			baseURL := fmt.Sprintf("%s://%s:%d%s", echo.Scheme(), echo.Pod.Status.HostIP, check.EchoServerHostPort, echo.Path())
+
+			var ipFamStr string
+
+			t.ForEachIPFamily(func(ipFam check.IPFamily) {
```
gandro, line 102: Thanks for the PR. You have a syntax error here. This func here is never closed.

I'd recommend testing the code out locally too, that will make it easier to spot defects.

Neilblaze: Ah silly 😭  
Thank you so much @gandro for the catch! Because of the fact that I work on two different PC's (`Ubuntu 22.04 LTS` + `Windows 10`), the mistake happened while I was shifting the code from one system to another. 😅 

> And thanks for the advice, I'll make sure to test the code out locally too to avoid such silly mistakes. 😄

Neilblaze: Also, @gandro this → [`github.com/cilium/cilium-cli/connectivity/check`](github.com/cilium/cilium-cli/connectivity/check) is throwing 404.

Additionally, how can I check the IP then? (refer to [this](https://github.com/cilium/cilium-cli/pull/1440/commits/8bef04a0cf94c8904322524bcaf811d35bd0aae4#diff-ad5bbb6bf8573892c9d19da501655d19000d19a334327b2790e5d45af02cb4ecR41))

gandro: The proper import path should be `"github.com/cilium/cilium-cli/connectivity/check"`. This is a Go *import path* and therefore not a valid GitHub URL. The GitHub URL for that file would be https://github.com/cilium/cilium-cli/tree/master/connectivity/check

> Should I have to use [this](https://github.com/cilium/cilium-cli/blob/master/connectivity/check/utils.go#LL14-L15C12)?

Yes, that's the one to use.

Neilblaze:
@gandro I think it's fine now. Phew!
The implemented code could be optimized using the ternary operator. I tried this:

```go
t.ForEachIPFamily(func(ipFam check.IPFamily) {
    ipFamStr = ipFam == check.IPFamilyV4 ? "v4" : "v6"
})
```

instead of

```go
t.ForEachIPFamily(func(ipFam check.IPFamily) {
    if ipFam == check.IPFamilyV4 {
        ipFamStr = "v4"
} else if ipFam == check.IPFamilyV6 {
    ipFamStr = "v6"
}
//...
```

It should work, but apparently, I was getting an error log stating _**"?" as illegal character**_. On doing some research, I found [this](https://stackoverflow.com/questions/38293759/hello-go11-illegal-character-u0023) on StackOverflow. Hence, I proceeded with the previous one. Also, instead of performing two checks, we can simply check whether it's `ipv4` or not, else it should be `ipv6`. 

If the above one sounds fine to you, then this can be implemented:

```go
t.ForEachIPFamily(func(ipFam check.IPFamily) {
    if ipFam == check.IPFamilyV4 {
        ipFamStr = "v4"
    } else {
        ipFamStr = "v6"
    }
// ...
```

Once you confirm, then I'll format the indentation & push the final changes.
Refer to [this](https://github.com/cilium/cilium-cli/pull/1440/commits/a248a047d9f18839e313aeeb48cab2afccfa8192#:~:text=Check%20failure%20on%20line%20124%20in%20connectivity/tests/host.go). And apologies for creating this simple thing a mess. 😅

connectivity/tests/host.go
```diff
@@ -35,7 +35,15 @@ func (s *podToHost) Run(ctx context.Context, t *check.Test) {
...
 			node := node
 
+			var ipFamStr string
+
 			t.ForEachIPFamily(func(ipFam check.IPFamily) {
+				if ipFam == check.IPFamilyV4 {
+					ipFamStr = "v4"
+				} else if ipFam == check.IPFamilyV6 {
+					ipFamStr = "v6"
+				}
```
gandro, line 45: Thanks! Golang does not support the ternary operator, that's why you get syntax errors.

What I would recommend instead however is either move this into a `ShortString()` method, or (my preference) just use the existing String() method. It will print `ipv4` instead of `v4`, but I think that's actually fine. Using `String()`, you can just pass the IPFamily directly to `Sprintf` with the `%s` specifier, no manual conversion to a string needed.

connectivity/tests/host.go
```diff
@@ -88,21 +96,31 @@ func (s *podToHostPort) Run(ctx context.Context, t *check.Test) {
...
+			t.ForEachIPFamily(func(ipFam check.IPFamily) {
+				if ipFam == check.IPFamilyV4 {
+					ipFamStr = "v4"
+				} else if ipFam == check.IPFamilyV6 {
+					ipFamStr = "v6"
+				}
+
+				baseURL := fmt.Sprintf("%s://%s:%d%s", echo.Scheme(), echo.Pod.Status.HostIP, check.EchoServerHostPort, echo.Path())
+				ep := check.HTTPEndpoint(echo.Name(), baseURL)
+				t.NewAction(s, fmt.Sprintf("curl-%s-%d", ipFamStr, i), &client, ep, check.IPFamilyAny).Run(func(a *check.Action) {
```
gandro, line 110: There are many other places in the code were we construct the `curl-%d` string that also need to be extended with the IP family (for example, pod.go).

connectivity/tests/host.go
```diff
@@ -88,21 +96,31 @@ func (s *podToHostPort) Run(ctx context.Context, t *check.Test) {
...
-					// observe flows after DNAT has been applied (e.g. by
-					// HostReachableServices),
-					AltDstIP:   echo.Address(check.IPFamilyAny),
-					AltDstPort: echo.Port(),
-				}))
-			})
+			var ipFamStr string
 
-			i++
+			t.ForEachIPFamily(func(ipFam check.IPFamily) {
```
gandro, line 101: The change which adds `t.ForEachIPFamily(func(ipFam check.IPFamily)` here does not make sense. The code within the func  does not actually use a different IP (it still calls the CurlCommand with IPFamilyAny).

I would suggest do revert this change here and only modify tests which currently already do call `ForEachIPFamily` and use different IPs in each invocation.

gandro (changes requested):
Thanks for the work so far. There are a few things that need to be addressed still. I've left some feedback below.

connectivity/tests/host.go
Neilblaze: @gandro Gotcha! So, it should be like this, right? 👇🏼 

```go
func (s *podToHost) Run(ctx context.Context, t *check.Test) {
	ct := t.Context()
	// Construct a unique list of all nodes in the cluster running workloads.

	var i int

	for _, pod := range ct.ClientPods() {
		pod := pod // copy to avoid memory aliasing when using reference

		for _, node := range ct.Nodes() {
			node := node

			t.ForEachIPFamily(func(ipFam check.IPFamily) {
				dstAddrs := []string{}
				for _, addr := range node.Status.Addresses {
					if check.GetIPFamily(addr.Address) != ipFam {
						continue
					}
					dstAddrs = append(dstAddrs, addr.Address)
				}
				if len(dstAddrs) == 0 {
					return
				}

				dst := check.ICMPEndpoint("", dstAddrs...)
				t.NewAction(s, fmt.Sprintf("ping-%s-%d", ipFam, i), &pod, dst, ipFam).Run(func(a *check.Action) {
					a.ExecInPod(ctx, ct.PingCommand(dst, ipFam))

					a.ValidateFlows(ctx, pod, a.GetEgressRequirements(check.FlowParameters{
						Protocol: check.ICMP,
					}))
				})

				i++
			})
		}
	}
}
```

If yes, then I got the rest of the part.

connectivity/tests/host.go
Neilblaze: Okay, that means, I need to make the following changes to [`pod.go`](https://github.com/cilium/cilium-cli/blob/master/connectivity/tests/pod.go) & [`service.go`](https://github.com/cilium/cilium-cli/blob/master/connectivity/tests/service.go) too, correct?

connectivity/tests/host.go
Neilblaze: Makes sense. 
Okay so confirming once again, I need to revert the changes done in `podToHostPort` since it still calls the CurlCommand with IPFamilyAny.

gandro: Correct!

connectivity/tests/host.go
gandro: Yes, basically all occurrences of ForEachIPFamily should be checked if we have different code based on the IPFamily and annotate the action name accordingly.

connectivity/tests/host.go
gandro: Yes, I think `fmt.Sprintf("ping-%s-%d", ipFam, i)` looks correct to me.

Neilblaze:
@gandro Tests are passing successfully! Once you approve the changes, I'll squash all the commits together along with the DCO.

Neilblaze:
@gandro All the tests ~are~ were passing (including `EKS(ENI)` & `Multicluster`). But after I squashed all the commits together (after adding the DCO), those two are specifically failing. Might be a [fluke](https://cilium.slack.com/archives/C7PE7V806/p1678705158483459?thread_ts=1678565058.006269&cid=C7PE7V806), would you mind re-triggering this job again? And if everything checks out, please merge :)

connectivity/tests/host.go
```diff
@@ -36,24 +36,28 @@ func (s *podToHost) Run(ctx context.Context, t *check.Test) {
 			node := node
 
 			t.ForEachIPFamily(func(ipFam check.IPFamily) {
+				dstAddrs := []string{}
 				for _, addr := range node.Status.Addresses {
 					if check.GetIPFamily(addr.Address) != ipFam {
 						continue
 					}
+					dstAddrs = append(dstAddrs, addr.Address)
```
gandro, line 44: Why this change? Let's not change any of the logic in this commit please. If you think this fixes a bug, we can change this in a separate commit or PR.

connectivity/tests/pod.go
```diff
@@ -50,7 +50,8 @@ func (s *podToPod) Run(ctx context.Context, t *check.Test) {
 				continue
 			}
 			t.ForEachIPFamily(func(ipFam check.IPFamily) {
-				t.NewAction(s, fmt.Sprintf("curl-%d", i), &client, echo, ipFam).Run(func(a *check.Action) {
+				actionDesc := fmt.Sprintf("curl-%s-%d", ipFam.String(), i)
```
gandro, line 53: nit, the following should work too:
```suggestion
				actionDesc := fmt.Sprintf("curl-%s-%d", ipFam, i)
```

connectivity/tests/service.go
```diff
@@ -49,7 +49,9 @@ func (s *podToService) Run(ctx context.Context, t *check.Test) {
 				continue
 			}
 
-			t.NewAction(s, fmt.Sprintf("curl-%d", i), &pod, svc, check.IPFamilyAny).Run(func(a *check.Action) {
+			actionDesc := fmt.Sprintf("curl-%s-%d", check.IPFamilyAny, i)
```
gandro, line 52: For `IPFamilyAny`, I would leave it as is at the moment and not change the output. I personally don't think adding `any` to the name adds value.

gandro (changes requested):
Thanks for the update. Yes, those look like unrelated flakes (it seems the github container registry has been struggling since yesterday, which is used for the actions too).

But before we validate CI, I think there are still a few things to address in the code. I've left some inline feedback.

In addition, there are more usages of ForEachIPFamily to patch (I recommend searchg for ForEachIPFamily in your IDE to make sure you get them all):

https://github.com/cilium/cilium-cli/blob/338836044b294ceeba7f72841e0c4b2042e6b16c/connectivity/tests/client.go#L42
https://github.com/cilium/cilium-cli/blob/acebe2761286a4f30f492e5baf5e83b1e33464cf/connectivity/tests/pod.go#L108
https://github.com/cilium/cilium-cli/blob/338836044b294ceeba7f72841e0c4b2042e6b16c/connectivity/tests/encryption.go#L155
https://github.com/cilium/cilium-cli/blob/338836044b294ceeba7f72841e0c4b2042e6b16c/connectivity/tests/encryption.go#L160

connectivity/tests/host.go
Neilblaze: Okay, I'll revert it.

connectivity/tests/service.go
Neilblaze: Okay, I'll fix this.

Neilblaze: Makes sense. I'll fix this.

Neilblaze:
@gandro I underestimated how simple it was, and as a result, I delved deeply into every aspect, leading me into all this unnecessary trouble lol. Hopefully this time it should be fine.

gandro:
GKE failed connecting to google.com - possibly due to external factors. Restarting

https://github.com/cilium/cilium-cli/actions/runs/4449884381/jobs/7814654455?pr=1440

Neilblaze:
> Please do not remove automatically assigned code-owners, since not every reviewer (like myself) belongs to all codeowner teams.

@gandro Sorry about the confusion, but as far as I can remember, I'm pretty sure I didn't unassign anyone 😅, but it seems like GitHub auto-removes any other reviewers when we select this 👇🏼 
image

nathanjsweet (approved):

gandro:
> > Please do not remove automatically assigned code-owners, since not every reviewer (like myself) belongs to all codeowner teams.
> 
> @gandro Sorry about the confusion, but as far as I can remember, I'm pretty sure I didn't unassign anyone sweat_smile, but it seems like GitHub auto-removes any other reviewers when we select this 👇🏼 

Oh, interesting, thanks for the clarification!

PR #1447, opened 2023-03-09
bug: "allow-all-egress.yaml" Manifest Traffic Fix

nathanjsweet:
The "allow-all" egress policy manifest file for the connectivity tests incorrectly "allows all" traffic with the policy of allowing all IPv4 traffic. This works because of a bug in Cilium which incorrectly translates the "0.0.0.0/0" CIDR policy as being synonymous with the "world" entity. This is incorrect and is being fixed. The policy now correctly refers to all external traffic.

### DIFF
--- a/connectivity/manifests/allow-all-egress.yaml
+++ b/connectivity/manifests/allow-all-egress.yaml
@@ -7,5 +7,5 @@ spec:
   egress:
   - toEndpoints:
     - {}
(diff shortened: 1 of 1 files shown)


joestringer (approved):

nathanjsweet:
I don't think this change is necessary now. It's cilium that's wrong.

PR #1457, opened 2023-03-16
clustermesh: correctly mount the etcd data dir in the init container

giorio94:
This PR fixes an incorrect volume mount specification for the init container of the clustermesh-apiserver deployment created through `cilium clustermesh enable`.


tklauser (approved):
Thanks

PR #1462, opened 2023-03-19
Update default Cilium version to v1.13.1

gaffneyd4:
Now that v1.13.1 [is available](https://github.com/cilium/cilium/releases/tag/v1.13.1), bumping this to the default version for the CLI to install.


tklauser:
Hi @gaffneyd4, thanks for the PR. It looks like the version was already bumped as part of https://github.com/cilium/cilium-cli/pull/1458 which was opened slightly earlier and now merged. That PR also bumped in the corresponding Helm charts for the newly released versions, which is required in order for installation to work correctly for these versions.

PR #1463, opened 2023-03-20
Add a l7 header replace test

meyskens:
This adds a test for the L7 policy to add/replace a HTTP header in a request. It does this by sending a request to a new endpoint that required an auth header, which will succeed if injected from a secret.

This came up while needing a test for https://github.com/cilium/cilium/issues/24020 and it will work against the current Cilium version where it was previously not tested in Ginkgo.

### DIFF
--- a/connectivity/check/policy.go
+++ b/connectivity/check/policy.go
@@ -182,6 +182,14 @@ var (
 		ExitCode:       ExitCurlHTTPError,
 	}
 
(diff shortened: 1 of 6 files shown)


jrajahalme (approved):
Looks good, thanks for the added coverage!

squeed (approved):

squeed:
This looks ready to merge - is the comment above still relevant?

meyskens:
The release got cut yesterday and PR updated

PR #1464, opened 2023-03-20
vendor: Pick up the latest cilium

michi-covalent:
Need https://github.com/cilium/cilium/pull/24470 to fix this error that happens during cilium connectivity test setup:

    connectivity test failed: unmarshaling cilium runtime config json: json: cannot unmarshal array into Go struct field DaemonConfig.NodeEncryptionOptOutLabels of type labels.Selector


sayboras (approved):
:checkered_flag:

PR #1474, opened 2023-03-25
Replace deprecated flags with --helm-set

michi-covalent:
- [x] Remove https://github.com/cilium/cilium-cli/pull/1474/commits/d401a18403c118a92fb86c3d69b0ba51ed8f066a before merging.


aanm (approved):
LGTM. I've only added the label so that we don't merge this PR accidentally before removing the last commit.

christarazi (approved):
Very :innocent:

PR #1475, opened 2023-03-26
Add Helm-based install / uninstall commands

michi-covalent:
please review commit by commit. there are 2 commits:

- refactoring a bit to support helm-based install / uninstall commands.
- implement install / uninstall commands.

let's implement other commands in separate PRs to keep the PR size reasonable.

ref: https://github.com/cilium/cilium-cli/issues/1396

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -22,7 +22,13 @@ jobs:
   installation-and-connectivity:
     runs-on: ubuntu-22.04
     timeout-minutes: 40
+    strategy:
+      matrix:
+        mode: ["classic", "helm"]
     steps:
+      - name: Set mode
+        run: |
+          echo "CILIUM_CLI_MODE=${{ matrix.mode }}" >> $GITHUB_ENV
       - name: Checkout
         uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3
 
@@ -65,8 +71,18 @@ jobs:
             --helm-set tls.secretsBackend=k8s
 
       - name: Enable Relay
+        if: ${{ matrix.mode == 'classic' }}
         run: |
           cilium hubble enable --ui
+
+      # TODO: Implement 'cilium hubble enable' command
+      - name: Enable Relay
+        if: ${{ matrix.mode == 'helm' }}
+        run: |
+          helm repo add cilium https://helm.cilium.io
+          helm upgrade -n kube-system cilium cilium/cilium --version ${{ env.cilium_version }} --set hubble.relay.enabled=true --set hubble.ui.enabled=true --reuse-values
+          helm get values -n kube-system cilium
+          cilium status --wait
 
       - name: Relay Port Forward
         run: |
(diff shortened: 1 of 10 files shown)


defaults/defaults.go
```diff
@@ -93,6 +93,8 @@ const (
 	IngressControllerName   = "cilium.io/ingress-controller"
 	IngressSecretsNamespace = "cilium-secrets"
 
+	// HelmReleaseName is the efault Helm release name for Cilium.
```
tklauser, line 96:
```suggestion
	// HelmReleaseName is the default Helm release name for Cilium.
```

install/install.go
```diff
@@ -965,3 +971,50 @@ func (k *K8sInstaller) RollbackInstallation(ctx context.Context) {
...
+		if helmValues["ipv4NativeRoutingCIDR"] == nil {
+			cidr, err := k.gkeNativeRoutingCIDR(ctx, k.client.ContextName())
+			if err != nil {
+				k.Log("❌ Unable to auto-detect GKE native routing CIDR. Is \"gcloud\" installed?")
+				k.Log("ℹ️  You can set the native routing CIDR manually with --helm-set ipv4NativeRoutingCIDR=x.x.x.x/x")
+				return err
+			}
+			k.params.IPv4NativeRoutingCIDR = cidr
+		}
+	}
```
tklauser, line 1000: Instead of duplicating this logic here and in `(*K8sInstaller).Install`, maybe move it to a common function. Or alternatively, drop this method altogether and move the helm-specific logic into `(*K8sInstaller).Install` behind a newly introduced `withHelm bool` parameter?

internal/cli/cmd/install.go
```diff
@@ -233,3 +233,85 @@ cilium upgrade --version %s
...
+	cmd.Flags().StringVar(&params.HelmChartDirectory, "chart-directory", "", "Helm chart directory")
+	cmd.Flags().StringSliceVar(&params.HelmOpts.ValueFiles, "helm-values", []string{}, "Specify helm values in a YAML file or a URL (can specify multiple)")
+	cmd.Flags().StringArrayVar(&params.HelmOpts.Values, "helm-set", []string{}, "Set helm values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
+	cmd.Flags().StringArrayVar(&params.HelmOpts.StringValues, "helm-set-string", []string{}, "Set helm STRING values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)")
+	cmd.Flags().StringArrayVar(&params.HelmOpts.FileValues, "helm-set-file", []string{}, "Set helm values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)")
+	cmd.Flags().BoolVar(&params.ListVersions, "list-versions", false, "List all the available versions without actually installing")
+	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait for helm install to finish")
```
tklauser, line 278: It's unfortunate these flags are duplicated between this function and `newCmdInstall`. Maybe define them in a common func that is called from both and `newCmdInstall` just adds its additional (deprecated) flags on top?

michi-covalent:
thank you for your review @aanm & @tklauser ! tobias, all of your suggestions make sense. i'm updating the PR now 🚀

tklauser (approved):
:sparkles:

.github/workflows/kind.yaml
```diff
@@ -22,7 +22,13 @@ jobs:
   installation-and-connectivity:
     runs-on: ubuntu-22.04
     timeout-minutes: 40
+    strategy:
+      matrix:
+        mode: ["classic", "helm"]
```
nbusseneau, line 27: Am I correct in that later on we will drop the classic mode and only have Helm available?

.github/workflows/kind.yaml
```diff
@@ -86,16 +102,37 @@ jobs:
...
         run: |
           cilium install \
           --version=${{ env.cilium_version}} \
           --encryption=ipsec \
           --helm-set kubeProxyReplacement=disabled
 
+      - name: Install Cilium with IPsec Encryption
+        if: ${{ matrix.mode == 'helm' }}
+        run: |
+          kubectl create -n kube-system secret generic cilium-ipsec-keys \
```
nbusseneau, line 115: Can't we still have this handled by the CLI even in Helm mode?

nbusseneau (approved):
This first stage LGTM, some questions below.

.github/workflows/kind.yaml
michi-covalent: yeah we'll gradually transition to the helm mode, and then deprecate the classic mode ✨

.github/workflows/kind.yaml
michi-covalent: yeah we could. i do want to have a discussion first though and come up with a general guideline on how cilium-cli should handle resources that are not part of the helm chart.

on the one hand, these flags are very convenient. on the other hand, i think it's a very slippery slope. we might end up adding more and more logic to cilium-cli without carefully considering:

- what resources should be managed by cilium, and what resources should be managed by the user.
- if cilium should be managing the resource, why it's not part of the helm chart.

.github/workflows/kind.yaml
```diff
@@ -65,9 +71,19 @@ jobs:
             --helm-set tls.secretsBackend=k8s
 
       - name: Enable Relay
+        if: ${{ matrix.mode == 'classic' }}
         run: |
           cilium hubble enable --ui
 
+      # TODO: Implement 'cilium hubble enable' command
```
christarazi, line 78: How difficult is it to do this?

christarazi (approved):

.github/workflows/kind.yaml
michi-covalent: extremely easy. just waiting for this PR to get merged before marking https://github.com/cilium/cilium-cli/pull/1479 ready for review 😊

Issue #1204, opened 2022-11-09
Connection to Hubble during flow validation is unreliable (i.e. `Timeout waiting for flow listener to become ready`)


gandro:
In multiple tests, we've observed that the connection between the Cilium CLI and Hubble is unreliable. This typically looks as follows:

```
 📄 Following flows...
  [.] Action [no-policies/client-to-client/ping-1: cilium-test/client2-5998d566b4-xsbdg (10.84.0.195) -> cilium-test/client-6488dcf5d4-wbs57 (10.84.0.66:0)]
  🟥 Timeout waiting for flow listener to become ready

[=] Test [allow-all-except-world]
  🟥 Receiving flows from Hubble Relay: hubble server status failure: context canceled
```

It's not clear why the flow listener times out. There are multiple potential explanations which may apply:

- The port-forward between the Cilium CLI and Hubble Relay is unreliable. This has been reported previously on AKS (for any port forward). For CI purposes, we could potentially rely on a service type LoadBalancer instead, which might be more stable. There are authentication concerns however, since anyone could access type LoadBalancer services.
- The connection between Hubble Relay and the Hubble Observers (inside cilium-agent) is unreliable, this could indicate a pod2host connectivity issue. This needs to be investigated as it likely would not only affect Hubble.
- No reconnection attempt is is made. The Cilium CLI should ideally try to reconnect to deal with ephemeral connectivity issues between it and the cluster.

Related https://github.com/cilium/cilium-cli/issues/1202

squeed:
We may wish to consider passing `--debug` to `cilium connectivity` in all the CI tests.

soggiest:
I've observed this as well  when running the `cilium connectivity test` command from my laptop to clusters running in EKS.

michi-covalent:
we've added retries to connect to hubble relay, but it still fails sometimes:
```
2023-06-06T14:31:20.206828192Z   🐛 hubble relay server status failed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:4245: connect: connection refused"
2023-06-06T14:31:21.207331072Z   🐛 retrying hubble relay server status request
2023-06-06T14:31:21.208022384Z   🐛 hubble relay server status failed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:4245: connect: connection refused"
[...]

PR #1494, opened 2023-04-08
Implement FromCIDR ingress tests

christarazi:
Requires https://github.com/cilium/cilium-cli/pull/1498 and I pulled out https://github.com/cilium/cilium-cli/pull/1501 to a separate PR because it broke CI on cilium/cilium.

Review commit by commit.

- connectivity/check: Deploy host-netns daemonset by default
- internal/cli/cmd: Clarify --nodes-without-cilium
- .github/workflows: Label nodes for --nodes-without-cilium
- .github/workflows: Install static routes for nodes without Cilium
- connectivity/check: Validate --external-from-cidrs
- Rename current CIDR tests to to-cidr.go
- connectivity: Refactor scenarios passed to tests
- connectivity: Implement FromCIDR tests
- .github: Prepare workflows for tests with a 3rd node

---

Providing commit msg of the most important commit for convenience:

> connectivity: Implement FromCIDR tests
>
> These should cover the tests from the Cilium (github.com/cilium/cilium)
Ginkgo suite for FromCIDRs, see [1].

[1]:
https://github.com/cilium/cilium/blob/8f9daa8c4e9ccf17c4badaa28aa6c1b0b51924a8/test/k8s/net_policies.go#L488-L557.


michi-covalent (approved):
lgtm on https://github.com/cilium/cilium-cli/pull/1494/commits/23a8dc868eeb3b8da4071f9f5ce89caea04eaf59. extra credit if you add a unit test.

.github/workflows/kind.yaml
```diff
@@ -135,6 +145,13 @@ jobs:
           cilium sysdump --output-filename cilium-sysdump-out --hubble-flows-count 10000
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
+      # This needs to be in sync with the flag --nodes-without-cilium below
+      # during installation.
+      - name: Unlabel nodes
+        if: ${{ matrix.mode == 'helm' }}
+        run: |
+          kubectl label nodes chart-testing-worker2 cilium.io/no-schedule=true
```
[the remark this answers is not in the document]
christarazi: Ah, that's a copy-pasta mistake. It should be fixed now.

christarazi:
* Kind hit https://github.com/cilium/cilium-cli/issues/1204
* External workloads hit https://github.com/cilium/cilium-cli/issues/1503

michi-covalent (approved):

michi-covalent (approved):

michi-covalent (approved):

.github/kind-config.yaml
```diff
@@ -13,6 +13,8 @@ nodes:
           taints: []
   - role: worker
     image: kindest/node:v1.19.11
+  - role: worker
+    image: kindest/node:v1.19.11
```
michi-covalent, line 17: looks like a duplicate

michi-covalent: oh no you are just adding another node

connectivity/check/deployment.go
```diff
@@ -257,7 +257,7 @@ func newDaemonSet(p daemonSetParameters) *appsv1.DaemonSet {
 							ReadinessProbe:  p.ReadinessProbe,
 							SecurityContext: &corev1.SecurityContext{
 								Capabilities: &corev1.Capabilities{
-									Add: []corev1.Capability{"NET_RAW"},
+									Add: []corev1.Capability{"NET_ADMIN", "NET_RAW"},
```
joestringer, line 260: Will this affect all connectivity tests any user runs in their environment? It seems to me that "configuring routes on the node" is very much outside the realm of stuff that `cilium connectivity test` should do in a production environment :thinking:

connectivity/tests/from-cidr.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (f *fromCIDRToPod) Run(ctx context.Context, t *check.Test) {
+	clientPod := t.Context().HostNetNSPodsByNode()[t.NodesWithoutCilium()[0]]
+	i := 0
+
+	for _, pod := range t.Context().EchoPods() {
+		ep := check.HTTPEndpoint(
+			"http-endpoint",
+			fmt.Sprintf("%s://%s:%d%s", pod.Scheme(), pod.Address(check.IPFamilyAny), pod.Port(), pod.Path()),
```
aanm, line 33: will not work for IPv6 because of the `:`, use https://pkg.go.dev/net#JoinHostPort

connectivity/manifests/echo-ingress-from-cidr.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: echo-ingress-from-cidr
+spec:
+  description: "Allow echo pods to receive ingress traffic from a specific CIDR"
+  endpointSelector:
+    matchLabels:
+      kind: echo
+  ingress:
+  - fromCIDR:
+    {{range $cidr := .ExternalFromCIDRs}} - {{$cidr}}/32
```
aanm, line 12: I assume this should be `/128` for the IPv6 case.

connectivity/check/deployment.go
christarazi: Synced offline. Moved to the CI infra / workflow logic to install static routes on the infrastructure rather than rely on the CLI to do it.

.github/workflows/kind.yaml
```diff
@@ -67,6 +67,59 @@ jobs:
...
+            done
+          done
+
+          # Join the elements with a comma delimiter, or leave them unmodified
+          # if there's only one element so that it can be passed properly to
+          # the CLI.
+          if [[ ${#EXTERNAL_NODE_IPS[@]} -eq 1 ]]; then
+            EXTERNAL_NODE_IPS_PARAM="${EXTERNAL_NODE_IPS[0]}"
+          else
+            EXTERNAL_NODE_IPS_PARAM=$(IFS=','; echo "${EXTERNAL_NODE_IPS[*]}")
```
asauber, line 126: very cool use of IFS and echo

.github/workflows/kind.yaml
```diff
@@ -59,11 +59,80 @@ jobs:
...
+          # Loop over each pod CIDR from all nodes.
+          for i in "${!EXTERNAL_FROM_CIDRS[@]}"; do
+            EXTERNAL_FROM_CIDR=${EXTERNAL_FROM_CIDRS[i]}
+
+            if [[ $EXTERNAL_FROM_CIDR =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/[0-9]+$ ]]; then
+              IP_FAMILY="v4"
+            elif [[ $EXTERNAL_FROM_CIDR =~ ^[0-9a-fA-F:]+/[0-9]+$ ]]; then
+              IP_FAMILY="v6"
+            else
+              echo "ERROR: External CIDR '${EXTERNAL_FROM_CIDR}' is invalid" >&2
```
christarazi, line 92:
```suggestion
              echo "ERROR: Malformed pod CIDR '${EXTERNAL_FROM_CIDR}'" >&2
```

.github/workflows/kind.yaml
```diff
@@ -135,6 +208,16 @@ jobs:
           cilium sysdump --output-filename cilium-sysdump-out --hubble-flows-count 10000
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
+      # This needs to be in sync with the flag --nodes-without-cilium below
+      # during installation.
```
christarazi, line 212: Actually, no need for this comment because `NODES_WITHOUT_CILIUM` will keep us in sync.

internal/cli/cmd/install.go
```diff
@@ -236,7 +236,7 @@ func addCommonInstallFlags(cmd *cobra.Command, params *install.Parameters) {
 	cmd.Flags().BoolVar(&params.ListVersions, "list-versions", false, "List all the available versions without actually installing")
 	cmd.Flags().BoolVar(&params.Wait, "wait", false, "Wait for helm install to finish")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", defaults.StatusWaitDuration, "Maximum time to wait for status")
-	cmd.Flags().StringSliceVar(&params.NodesWithoutCilium, "nodes-without-cilium", []string{}, "List of node names on which Cilium will not be installed. In Helm installation mode, it's assumed that the no-schedule node labels are present.")
+	cmd.Flags().StringSliceVar(&params.NodesWithoutCilium, "nodes-without-cilium", []string{}, "List of node names on which Cilium will not be installed. In Helm installation mode, it's assumed that the no-schedule node labels are present and that the infastructure has set up routing on these nodes to provide connectivity within the Cilium cluster.")
```
christarazi, line 239: Calling this out as an FYI.

.github/workflows/kind.yaml
```diff
@@ -59,11 +59,80 @@ jobs:
           version: ${{ env.KIND_VERSION }}
           config: ${{ env.KIND_CONFIG }}
 
+      # This needs to be in sync with the flag --nodes-without-cilium below
+      # during installation.
```
christarazi, line 63: No need for this comment as well

asauber (approved):
LGTM, commits made the review easy

.github/workflows/kind.yaml
christarazi: Looks like the kind workflow runs with Ubuntu 22.04 so we should be good here.

brlbil (approved):
I am fond of this much bash magic in the workflow. It makes it hard to follow along with what is going on.
But I do not know a better way to do it myself 😄 so for workflow parts 👍

christarazi:
[Pushed](https://github.com/cilium/cilium-cli/compare/afddc3df0bd31bdc62d441fe57b339dc26828bb7..e0d6f358acbf75f56cbc09b0602e7e670edaef30) to fix up an error msg and code comments, so the push had no functional impact. The CI was already green given MLH marked it as ready to merge.

.github/workflows/kind.yaml
```diff
@@ -59,11 +59,78 @@ jobs:
...
+          done
+
+      # This is needed for the tests that run with nodes without Cilium.
+      - name: Install static routes
+        run: |
+          EXTERNAL_FROM_CIDRS=($(kubectl get nodes -o jsonpath='{range .items[*]}{.spec.podCIDR}{"\n"}{end}'))
+          EXTERNAL_NODE_IPS=() # Nodes IPs are collected to be passed to the Cilium CLI later on.
+
+          # Loop over each pod CIDR from all nodes.
+          for i in "${!EXTERNAL_FROM_CIDRS[@]}"; do
```
joestringer, line 82: I didn't know about getting keys with the `!` syntax, TIL.

.github/workflows/kind.yaml
```diff
@@ -59,11 +59,78 @@ jobs:
...
+            EXTERNAL_FROM_CIDR=${EXTERNAL_FROM_CIDRS[i]}
+
+            if [[ $EXTERNAL_FROM_CIDR =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+/[0-9]+$ ]]; then
+              IP_FAMILY="v4"
+            elif [[ $EXTERNAL_FROM_CIDR =~ ^[0-9a-fA-F:]+/[0-9]+$ ]]; then
```
joestringer, line 87: Not perfect checks for IP address formats but good enough especially in a workflow where I presume we're controlling the inputs.

joestringer (approved):
I think that architecturally the environment configuration stuff belongs in the setup, though I do agree with Birol that embedding bash like this makes the workflows more complicated and difficult to grok. Maybe at some point we'll just reinvent the workflows as a github action with the relevant logic in it (better code reuse for cilium/cilium that way too). Anyway, now that we're not expanding the security footprint of the deployments this LGTM. I stared at the bash for a while and nothing jumped out.

.github/workflows/kind.yaml
christarazi: Yeah, I'm aware they aren't perfect and went for just good enough to be usable. I wanted to avoid complicating the bash code further with super long regexes, but definitely any item for followup if we can also find a way to make it more readable, i.e. using a variable instead and hide the definition of it.

aanm (approved):

internal/cli/cmd/install.go
michi-covalent: appreciating the call-out almost one year later 🙏

PR #1492, opened 2023-04-05
Implement `cilium upgrade` using Helm

asauber:
### Summary

Implements the `cilium upgrade` command using a thin wrapper around Helm rather than direct Kubernetes API calls.

A few of the self-contained features are broken down into separate commits.

### Motivation

This implementation allows for more consistent documentation and behavior among Cilium installation methods, such as those docs which use only Helm directly.

This implementation includes an alternate implementation for the `cilium {clustermesh,hubble} {enable,disable}` family of commands. The "legacy" implementation of those commands came from a time when using Helm was insecure (it installed a daemon into the cluster with `cluster-admin` privileges. Helm is now relatively safe to use, thus we use Helm here as a library to implement all features for Cilium installation, upgrade, and feature enable/disable whcih are currently implemented in the Cilium Helm chart.

### Example Usage

For example, we can use this new upgrade command to enable Cluster Mesh on a set of two Cilium clusters with the following procedure.

Create a kind cluster with the following config

```bash
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: c1
nodes:
  - role: control-plane
  - role: worker
  - role: worker
  - role: worker
networking:
  disableDefaultCNI: true
  podSubnet: "10.10.0.0/16"
  serviceSubnet: "10.50.0.0/16"
```

```
kind create cluster --config c1.kindconfig.yaml
```

Test that the kind cluster came up with `k get pods -A`

Install Cilium using the CLI, making sure to set a Cluster ID and Cluster Name

```bash
export CILIUM_CLI_MODE=helm
cilium install --help
# note the suggestion to use --helm-set flags
cilium install --helm-set cluster.id=1 --helm-set cluster.name=cluster1
```

Create a second cluster using kind with the following config

```bash
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: c2
nodes:
  - role: control-plane
  - role: worker
  - role: worker
  - role: worker
networking:
  disableDefaultCNI: true
  podSubnet: "10.11.0.0/16"
  serviceSubnet: "10.51.0.0/16"
```

```
kind create cluster --config c2.kindconfig.yaml
```

Check that you are authed for Cluster `kind-c2`

Run `k get pods -A` and look for `kube-apiserver-c2-control-plane`

Install Cilium using the CLI, making sure to set a Cluster ID and Cluster Name

```bash
export CILIUM_CLI_MODE=helm
cilium install --help
# note the suggestion to use --helm-set flags
[...]

PR #1495, opened 2023-04-08
flag --chart-directory added to upgrade

### LINKED ISSUE #1493: `cilium upgrade` doesn't support `--chart-directory`
When upgrading to a specific set of images, it may be necessary to pass a new Helm chart to use (for example, see https://github.com/cilium/cilium/issues/24787). That is however not supported today in the `upgrade` command:
```
$ ./cilium upgrade --agent-image quay.io/cilium/cilium-ci:latest --operator-image quay.io/cilium/operator-generic-ci:latest --chart-directory=$CILIUM_SRC/install/kubernetes/cilium
unknown flag: --chart-directory
```

chirag0002:
flag ``` --chart-directory ``` is now supported in the ```upgrade``` command which can be used to pass a new chart to use.

closes #1493

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -213,10 +213,10 @@ cilium upgrade --version %s
 	cmd.Flags().StringVar(&params.Version, "version", defaults.Version, "Cilium version to install")
 	cmd.Flags().BoolVar(&params.Wait, "wait", true, "Wait for status to report success (no errors)")
 	cmd.Flags().DurationVar(&params.WaitDuration, "wait-duration", defaults.StatusWaitDuration, "Maximum time to wait for status")
-	cmd.Flags().StringVar(&params.HelmChartDirectory, "chart-directory", "", "Helm chart directory")
 	cmd.Flags().StringVar(&params.AgentImage, "agent-image", "", "Image path to use for Cilium agent")
 	cmd.Flags().StringVar(&params.OperatorImage, "operator-image", "", "Image path to use for Cilium operator")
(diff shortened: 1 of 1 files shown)


asauber (changes requested):
Hey @chirag0002, appreciate the interest in resolving #1493. At the moment, the CLI has [two modes](https://github.com/cilium/cilium-cli/blob/b54289770a3391604c4e661cb832c17d32f4e282/internal/cli/cmd/cmd.go#L82) for the `Install` command. The first one uses the resulting chart from `HelmChartDirectory` to [render manifests locally](https://github.com/cilium/cilium-cli/blob/b54289770a3391604c4e661cb832c17d32f4e282/internal/helm/helm.go#L303). The second one uses the resulting chart from `HelmChartDirectory` directly with [a Helm install command](https://github.com/cilium/cilium-cli/blob/b54289770a3391604c4e661cb832c17d32f4e282/install/install.go#L1002).

As for the `Upgrade` code path: as it is, your patch would initialize `HelmChartDirectory` for that code path, but it would have no effect, as [the upgrade implementation](https://github.com/cilium/cilium-cli/blob/b54289770a3391604c4e661cb832c17d32f4e282/install/upgrade.go#L151) does not currently use the `helmChart` value as can be found when tracing the `Install` code paths.

You are welcome to implement a Helm-based `Upgrade` mode, and I would be happy to review.

chirag0002:
Thanks for the feedback and suggestions...
 I'll try this out and let you know if there is any progress.

asauber:
Hi @chirag0002, as this feature became relatively higher priority, I included it as part of https://github.com/cilium/cilium-cli/pull/1492. Going to close this PR for now.

Please review #1492 if you have a chance!

PR #1470, opened 2023-03-24
Add CRD feature and add KNP tests

doniacld:
## Motivation

Adding Kubernetes Network Policy tests coverage.
 
## Changes

- Add K8S extension client to retrieve CRD
- Add CRD feature and CNP enabled requirement
- Add Kubernetes Network Policies
- Add Kubernetes Network Policies tests scenarios

## List of scenarios added for Kubernetes Network Policy

| Scenario                                   | CNP                                                                                | KNP                                               | Comment                               |
|--------------------------------------------|------------------------------------------------------------------------------------|---------------------------------------------------|---------------------------------------|
| no-policies                                | -                                                                                  | -                                                 |                                       |
| no-policies-extra                          | -                                                                                  | -                                                 |                                       |
| allow-all-except-world                     | manifests/allow-all-except-world.yaml                                              | ❌                                                 | no entities in KNP                    |       
| client-ingress                             | manifests/client-ingress-from-client2.yaml                                         | manifests/client-ingress-from-client2-knp.yaml    |                                       |
| all-ingress-deny                           | manifests/deny-all-ingress.yaml                                                    | manifests/deny-all-ingress-knp.yaml               |                                       |
| all-egress-deny                            | manifests/deny-all-egress.yaml                                                     | manifests/deny-all-egress-knp.yaml                |                                       |
| all-entities-deny                          | manifests/deny-all-entities.yaml                                                   | ❌                                                 | no entities in KNP                    |
| cluster-entity                             | manifests/allow-cluster-entity.yaml                                                | ❌                                                 | no entities in KNP                    |
| host-entity                                | manifests/allow-host-entity.yaml                                                   | ❌                                                 | no entities in KNP                    |
| echo-ingress                               | manifests/echo-ingress-from-other-client.yaml                                      | manifests/echo-ingress-from-other-client-knp.yaml |                                       |
| client-ingress-icmp                        | manifests/echo-ingress-icmp.yaml                                                   | ❌                                                 | no icmp in KNP                        |
| client-egress                              | manifests/client-egress-to-echo.yaml                                               | manifests/client-egress-to-echo-knp.yaml          |                                       |
| client-egress-expression                   | manifests/client-egress-to-echo-expression.yaml                                    | manifests/client-egress-to-echo-expression-knp.yaml                                                 |  |
| client-egress-to-echo-service-account      | manifests/client-egress-to-echo-service-account.yaml                               | ❌                                                 | no Cilium identity labels in KNP      |
[...]

PR #1505, opened 2023-04-12
Add connectivity tests for auth

meyskens:
This adds tests to validate that the auth handling in policy is working. It uses  the always-fail type to test the fail case. If mTLS-SPIFFE is enabled in the cluster it will also perfom a successful test run with mTLS enabled.

The auth-fail tests will be able to run against `master` of cilium, the mTLS tests can be enabled in installation once https://github.com/cilium/cilium/pull/24765 is merged

### DIFF
--- a/connectivity/check/policy.go
+++ b/connectivity/check/policy.go
@@ -202,6 +202,10 @@ func defaultDenyReason(flow *flowpb.Flow) bool {
 	return flow.GetDropReasonDesc() == flowpb.DropReason_POLICY_DENIED
 }
 
+func authRequiredDropReason(flow *flowpb.Flow) bool {
+	return flow.GetDropReasonDesc() == flowpb.DropReason_AUTH_REQUIRED
+}
+
 var (
 	// ResultNone expects a successful command, don't match any packets.
 	ResultNone = Result{
@@ -246,6 +250,12 @@ var (
 		Drop:           true,
 		ExitCode:       ExitAnyError,
 		DropReasonFunc: defaultDropReason,
+	}
+
+	// ResultDropAuthRequired expects a dropped flow with auth required as reason.
+	ResultDropAuthRequired = Result{
+		Drop:           true,
+		DropReasonFunc: authRequiredDropReason,
 	}
 
 	// ResultAnyReasonEgressDrop expects a dropped flow at Egress and a failed command.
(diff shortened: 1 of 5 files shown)


sayboras (approved):
LGTM :heavy_check_mark:, I don't think there is any dependency to merge this PR in as the feature requirement is in place.

```
$ cilium connectivity test --test auth
...
[=] Test [echo-ingress-auth-always-fail]
........
[=] Test [echo-ingress-auth-mtls-spiffe]
........

[=] Skipping Test [dns-only]

[=] Skipping Test [to-fqdns]

✅ All 2 tests (16 actions) successful, 35 tests skipped, 0 scenarios skipped.
```

mhofstetter (approved):
great - thanks a lot :+1:

connectivity/suite.go
```diff
@@ -824,6 +830,27 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
 			return check.ResultCurlHTTPError, check.ResultNone // if the header is not set the request will get a 401
 		})
 
+	// Test mTLS auth with always-fail
+	ct.NewTest("echo-ingress-auth-always-fail").WithPolicy(echoIngressAuthFailPolicyYAML).
```
tklauser, line 834: #1470 renamed `WithPolicy` to `WithCiliumPolicy`:

```suggestion
	ct.NewTest("echo-ingress-auth-always-fail").WithCiliumPolicy(echoIngressAuthFailPolicyYAML).
```

same fore the other occurrences below.

meyskens: Thanks!

connectivity/check/policy.go
```diff
@@ -248,6 +252,13 @@ var (
 		DropReasonFunc: defaultDropReason,
 	}
 
+	// ResultDropAuthRequired expects a dropped flow with auth required as reason.
+	ResultDropAuthRequired = Result{
+		Drop: true,
+		// IngressDrop:    true,
```
tklauser, line 258: Do we expect to ever use this? I'd either add a TOD comment with an explanation on why it is commented out or drop it in case we don't expect to use it.

tklauser (approved):
LGTM 🚀 One minor non-blocking comment.

connectivity/check/policy.go
meyskens: this was a leftover of my experiments let me fix that

meyskens:
(sorry forgot to push my yubikey so i thought the push worked but it didn't)

PR #1527, opened 2023-04-21
helm mode: add clustermesh enable/disable

asauber:
Two commits for review

#### Add cilium clustermesh {enable,disable} using Helm

Implements a version of `clustermesh enable` and `clustermesh disable` using Helm actions run directly against the cluster

#### Enable Helm mode matrix for multicluster tests

Enables a workflow matrix for "classic" and "helm" mode for multicluster tests using GKE. The recently added parameter for cilium_cli_mode in the `cilium-cli-test-job-chart` is used.

### DIFF
--- a/internal/cli/cmd/clustermesh.go
+++ b/internal/cli/cmd/clustermesh.go
@@ -26,13 +26,23 @@ func newCmdClusterMesh() *cobra.Command {
 	}
 
 	cmd.AddCommand(
-		newCmdClusterMeshEnable(),
-		newCmdClusterMeshDisable(),
 		newCmdClusterMeshConnect(),
 		newCmdClusterMeshDisconnect(),
 		newCmdClusterMeshStatus(),
 		newCmdClusterMeshExternalWorkload(),
 	)
+
+	if os.Getenv("CILIUM_CLI_MODE") == "helm" {
+		cmd.AddCommand(
+			newCmdClusterMeshEnableWithHelm(),
+			newCmdClusterMeshDisableWithHelm(),
+		)
+	} else {
+		cmd.AddCommand(
+			newCmdClusterMeshEnable(),
+			newCmdClusterMeshDisable(),
+		)
+	}
 
 	return cmd
 }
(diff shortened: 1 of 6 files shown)


internal/helm/helm.go
```diff
@@ -361,6 +361,16 @@ func MergeVals(
...
+func ParseVals(helmStrValues []string) (map[string]interface{}, error) {
+	helmValStr := strings.Join(helmStrValues, ",")
+	helmValues := map[string]interface{}{}
+	err := strvals.ParseInto(helmValStr, helmValues)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing helm options %q: %w", helmValStr, err)
+	}
+	return helmValues, nil
+}
+
```
michi-covalent, line 373: this function looks very unit-testable 👀

michi-covalent (approved):
lgtm maybe add some unit tests for `ParseVals`

asauber:
The new workflow was temporarily enabled on this PR for testing purposes. Logs for a successful run of the modified and new workflow can be found here:

(classic) https://github.com/cilium/cilium-cli/actions/runs/4789575696/jobs/8517609808?pr=1527
(helm) https://github.com/cilium/cilium-cli/actions/runs/4789575696/jobs/8517609980?pr=1527

michi-covalent (approved):

internal/cli/cmd/clustermesh.go
```diff
@@ -336,3 +346,49 @@ func newCmdExternalWorkloadStatus() *cobra.Command {
...
+			params.Namespace = namespace
+			ctx := context.Background()
+			if err := clustermesh.EnableWithHelm(ctx, k8sClient, params); err != nil {
+				fatalf("Unable to enable ClusterMesh: %s", err)
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&params.ServiceType, "service-type", "LoadBalancer", "Type of Kubernetes service to expose control plane { ClusterIP | LoadBalancer | NodePort }")
```
YutaroHayakawa, line 369: This changes the default value. Did you have any reason you wanted to change the default?

YutaroHayakawa, line 369: This changes the default value from the old behavior. Did you have any reason you wanted to change the default?

YutaroHayakawa (approved):
Approved with 1 question.

internal/cli/cmd/clustermesh.go
asauber: I believe I chose `LoadBalancer` here, because some version of the clustermesh setup docs (which I can't seem to find at the moment), specified `--service-type NodePort`, as though LoadBalancer was the default. 

The default does appear to be `NodePort` https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/values.yaml#L2385

I'll change this to `NodePort`.

michi-covalent:
@brlbil ping!

internal/cli/cmd/clustermesh.go
YutaroHayakawa: Thanks for your consideration. I would also like to know which of the cilium-cli and Helm defaults we should preserve. I don't know the overall direction of this work, but if we use the old cilium-cli default, the default is an empty string, and cilium-cli does auto-detection depending on the environment (https://github.com/cilium/cilium-cli/blob/main/clustermesh/clustermesh.go#L111-L135). In your new logic, you cut out the auto-detection logic. Is it expected?

brlbil (approved):
CI parts looks good

internal/cli/cmd/clustermesh.go
asauber: Yes, the current direction for much of this autodetection logic is to replace it with documentation, as it has been a source of bugs.

michi-covalent:
it's green, there i click.

PR #1528, opened 2023-04-21
defaults: bump json-mock to 1.3.5

jibi:
which includes the new /client-ip route [1] to echo back the client IP in the response and will be used in a subsequent commit

[1] https://github.com/cilium/json-mock/releases/tag/v1.3.5


squeed (approved):
:ship:

PR #1529, opened 2023-04-21
fixes versioncheck for l7 proxy disabling with wireguard encryption

ShiroDN:
This commit fixes versioncheck so that it includes patch versions of cilium 1.13.x. Versions 1.13.x triggers en error:

level=fatal msg="Wireguard (--enable-wireguard) is not compatible with L7 proxy (--enable-l7-proxy)" subsys=daemon

Fixes: 64924bfa9663 ("enabled l7Proxy & removed err.log message for WireGuard")


brb (approved):
Thanks!

squeed (approved):
nice catch, thanks.

PR #1540, opened 2023-04-26
connectivity: add retry to external targets

brlbil:
Adds --retry and --retry-delay flags to the connectivity test.
They are used in external `pod-to-world` and `po-to-cidr` tests to try to reconnect
on failures thus making flaky tests less likely.

### DIFF
--- a/connectivity/tests/common.go
+++ b/connectivity/tests/common.go
@@ -3,6 +3,12 @@
 
 package tests
 
+import (
+	"strconv"
+
+	"github.com/cilium/cilium-cli/connectivity/check"
+)
+
 type labelsContainer interface {
 	HasLabel(key, value string) bool
 }
@@ -47,4 +53,80 @@ func hasAllLabels(labelsContainer labelsContainer, filters map[string]string) bo
 		}
 	}
 	return true
+}
+
+type retryCondition struct {
+	podLabels map[string]string
+	all       bool
+	destPort  uint32
+	destIP    string
+}
+
+// CurlOptions returns curl retry option or empty slice depending on retry conditions
+func (rc *retryCondition) CurlOptions(peer check.TestPeer, ipFam check.IPFamily, pod check.Pod, params check.Parameters) []string {
+	if params.Retry == 0 {
+		return []string{}
+	}
+	if !rc.all && rc.destIP == "" && rc.destPort == 0 {
+		return []string{}
+	}
+
+	opts := []string{
+		"--retry", strconv.FormatInt(int64(params.Retry), 10),
+		"--retry-all-errors", // add --retry-all-errors to retry on all possible errors
+	}
+
+	if retryDelay := params.RetryDelay.Seconds(); retryDelay > 0.0 {
+		opts = append(opts, "--retry-delay", strconv.FormatFloat(retryDelay, 'f', -1, 64))
+	}
+
+	if rc.all {
+		return opts
+	}
+	if rc.destIP != "" && peer.Address(ipFam) != rc.destIP {
+		return []string{}
+	}
+	if rc.destPort != 0 && peer.Port() != rc.destPort {
+		return []string{}
+	}
+	for n, v := range rc.podLabels {
+		if !pod.HasLabel(n, v) {
+			return []string{}
+		}
+	}
+
+	return opts
+}
+
+type RetryOption func(*retryCondition)
+
+// WithRetryAll sets all condition, returns retry options in every case
+func WithRetryAll() RetryOption {
+	return func(rc *retryCondition) {
+		rc.all = true
+	}
+}
+
+// WithRetryDestIP sets ip address condition
+func WithRetryDestIP(ip string) RetryOption {
+	return func(rc *retryCondition) {
+		rc.destIP = ip
+	}
+}
+
+// WithRetryDestPort sets port condition
+func WithRetryDestPort(port uint32) RetryOption {
+	return func(rc *retryCondition) {
+		rc.destPort = port
+	}
+}
+
+// WithRetryPodLabel sets pod label condition
+func WithRetryPodLabel(name, val string) RetryOption {
+	return func(rc *retryCondition) {
+		if rc.podLabels == nil {
+			rc.podLabels = map[string]string{}
+		}
+		rc.podLabels[name] = val
+	}
 }
--- a/connectivity/tests/to-cidr.go
+++ b/connectivity/tests/to-cidr.go
@@ -11,14 +11,20 @@ import (
 	"github.com/cilium/cilium-cli/connectivity/check"
 )
 
-// PodToCIDR sends an ICMP packet from each client Pod
-// to ExternalIP and ExternalOtherIP.
-func PodToCIDR() check.Scenario {
-	return &podToCIDR{}
+// PodToCIDR sends an HTTPS request from each client Pod
+// to ExternalIP and ExternalOtherIP
+func PodToCIDR(opts ...RetryOption) check.Scenario {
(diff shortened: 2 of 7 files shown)


squeed:
Hmm, it seems like the retries for cases where failure is expected are causing the tests to take too long.

squeed (changes requested):
Is there an easy way to disable retries when the test is expected to fail?

brlbil:
The default retry policy is 3 times and 3s delay. We can reduce this number, but first I will check the test cases for expected fails and see if we approach them differently.

brlbil:
> Is there an easy way to disable retries when the test is expected to fail?

I have made changes that retry options are added for commands where success is expected.
Also, the second commit fixes a couple of comments.

connectivity/tests/common.go
```diff
@@ -48,3 +54,81 @@ func hasAllLabels(labelsContainer labelsContainer, filters map[string]string) bo
...
+	podLabels map[string]string
+	all       bool
+	destPort  uint32
+	destIP    string
+}
+
+// CurlOptions returns curl retry option or empty slice depending on retry conditions
+// if all value set it means always return retry options
+// for pod-to-cidr all or destIP is checked
+// for pod-to-world all, destPort and destPort + pod label is checked
```
tklauser, line 68: Might make sense to move these comments to the respective `With*() RetryOption` funcs below.

connectivity/tests/to-cidr.go
```diff
@@ -11,14 +11,20 @@ import (
 	"github.com/cilium/cilium-cli/connectivity/check"
 )
 
-// PodToCIDR sends an ICMP packet from each client Pod
-// to ExternalIP and ExternalOtherIP.
-func PodToCIDR() check.Scenario {
-	return &podToCIDR{}
+// PodToCIDR sends a HTTPS request from each client Pod
+// to ExternalIP default 1.1.1.1 and ExternalOtherIP default 1.0.0.1.
```
tklauser, line 15: Let's omit the defaults:

```suggestion
// to ExternalIP and ExternalOtherIP.
```

connectivity/tests/common.go
```diff
@@ -48,3 +54,81 @@ func hasAllLabels(labelsContainer labelsContainer, filters map[string]string) bo
...
+	}
+	// destPort and podLabel are checked with pod-to-world case
+	if peer.Port() == rc.destPort {
+		for n, v := range rc.podLabels {
+			if !pod.HasLabel(n, v) {
+				return []string{}
+			}
+		}
+		return opts
+	}
```
tklauser, line 102: It feels a bit strange this this is essentially `all OR destIP matches OR (desPort matches AND label matches)`. Though it currently doesn't have a practical consequence, I'd say it's more straight forward to apply the options if either `all` is true OR all specified options are true, i.e. the `AND` of all specified options. Something like:

```go
func (rc *retryCondition) CurlOptions(peer check.TestPeer, ipFam check.IPFamily, pod check.Pod, params check.Parameters) []string {                                                                                                                
        if params.Retry == 0 { 
                return []string{}
        } 

        opts := []string{
                "--retry", strconv.FormatInt(int64(connectRetry), 10),
                "--retry-all-errors", // add --retry-all-errors to retry on all possible errors
        } 

        if retryDelay := params.RetryDelay.Seconds(); retryDelay > 0.0 { 
                opts = append(opts, "--retry-delay", strconv.FormatFloat(retryDelay, 'f', -1, 64))
        } 
        if rc.all { 
                return opts
        } 
        if rc.destIP != "" && peer.Address(ipFam) != rc.destIP { 
                return []string{}
        } 
        if rc.destPort != 0 && peer.Port() != rc.destPort { 
                return []string{}
        } 
        for n, v := range rc.podLabels { 
                if !pod.HasLabel(n, v) { 
                        return []string{}
                } 
        } 

        return opts
}
```

That way it's independent of the test cases these options are currently used with and allows for easier extensibility in the future.

What do you think?

connectivity/tests/common.go
```diff
@@ -48,3 +54,81 @@ func hasAllLabels(labelsContainer labelsContainer, filters map[string]string) bo
...
+}
+
+func WithRetryPodLabel(name, val string) RetryOption {
+	return func(ef *retryCondition) {
+		if ef.podLabels == nil {
+			ef.podLabels = map[string]string{}
+		}
+		ef.podLabels[name] = val
+	}
+}
```
tklauser, line 134: Nit: change the receiver name from `ef` to `rc`:

```suggestion
func WithRetryAll() RetryOption {
	return func(rc *retryCondition) {
		rc.all = true
	}
}

func WithRetryDestIP(ip string) RetryOption {
	return func(rc *retryCondition) {
		rc.destIP = ip
	}
}

func WithRetryDestPort(port uint32) RetryOption {
	return func(rc *retryCondition) {
		rc.destPort = port
	}
}

func WithRetryPodLabel(name, val string) RetryOption {
	return func(rc *retryCondition) {
		if rc.podLabels == nil {
			rc.podLabels = map[string]string{}
		}
		rc.podLabels[name] = val
	}
}
```

connectivity/tests/world.go
```diff
@@ -10,14 +10,19 @@ import (
 	"github.com/cilium/cilium-cli/connectivity/check"
 )
 
-// PodToWorld sends multiple HTTP(S) requests to one.one.one.one
-// from random client Pods.
-func PodToWorld() check.Scenario {
-	return &podToWorld{}
+// PodToWorld sends multiple HTTP(S) requests to ExternalTarget default one.one.one.one
```
tklauser, line 13: Let's omit the default:

```suggestion
// PodToWorld sends multiple HTTP(S) requests to ExternalTarget
```

tklauser (changes requested):
> Also, the second commit fixes a couple of comments.

Could you please squash the two commits? It looks like the 2nd commit only changes comments that got introduced/changed in the 1st commit.

connectivity/tests/common.go
brlbil: 👍

connectivity/tests/common.go
brlbil: First I used and other name for retryCondition which it short form was ef 😄 
good catch.

connectivity/tests/common.go
brlbil: This implementation is much more elegant 👍 but it is missing one thing if no options are set it still sets the retry options. So I added a check earlier to return in this case
```
	if !rc.all && rc.destIP == "" && rc.destPort == 0 {
		return []string{}
	}
```

squeed (approved):

connectivity/tests/common.go
```diff
@@ -48,3 +54,79 @@ func hasAllLabels(labelsContainer labelsContainer, filters map[string]string) bo
...
+	destPort  uint32
+	destIP    string
+}
+
+// CurlOptions returns curl retry option or empty slice depending on retry conditions
+func (rc *retryCondition) CurlOptions(peer check.TestPeer, ipFam check.IPFamily, pod check.Pod, params check.Parameters) []string {
+	if params.Retry == 0 {
+		return []string{}
+	}
+	if !rc.all && rc.destIP == "" && rc.destPort == 0 {
```
tklauser, line 70: Shouldn't we extend this to also consider empty pod labels?

```suggestion
	if !rc.all && rc.destIP == "" && rc.destPort == 0 && len(rc.podLabels) == 0 {
```

brlbil: Pool labels are only used with other conditions, in this case only with destPort.

tklauser (approved):

Issue #1260, opened 2022-12-03
Connectivity blip leads to failure to remove policy


pchaigno:
[This AKS workflow](https://github.com/cilium/cilium/actions/runs/3609794144/jobs/6083205216) had a lot of tests failing just because of a connectivity blip at the beginning.

The connectivity blip happened while trying to get the policy revision, in preparation to remove the host-cluster policy:
```
Error finalizing 'host-entity': getting policy revisions for Cilium agents: error dialing backend: read unix @->/tunnel-uds/proxysocket: read: connection reset by peer
```
Because we failed to get the policy revisions, [we never deleted that policy](https://github.com/cilium/cilium-cli/blob/50a1b3b4a207c0ac64e3825800d3be92e3af4cd0/connectivity/check/policy.go#L441) and a lot of tests then failed because the egress of pods was severely limited.

I can think of several ways to make this more resilient:
1. Always delete all CNPs in the test namespace instead of only those that are known to be used for the current test case. That would have restored correct state after the next test case.
2. Delete the policies even if we don't have a policy revision. That means we won't be able to wait for them to be "un-enforced"
    - Worst case, the subsequent test also fails.
    - In that case, we could always fallback to a sleep :-/
3. Retry the step to get the policy revision.

What would be preferred? I tend to think we could at least do 1.

PR #1541, opened 2023-04-26
connectivity: Retry getting the policy revision

pchaigno:
This commit adds some retry logic to the policy revision retrieval in case of a timeout. Such timeouts can happen because of simple connectivity blips.

This piece of code has been identified as fairly critical [1]. We attempt to retrieve these policy revisions when applying and deleting policies. If we fail to retrieve the policy revision, we fail the whole policy deletion. That can lead to failures in subsequent tests because of the leftover policy.

1 - https://github.com/cilium/cilium-cli/issues/1260
Fixes: https://github.com/cilium/cilium-cli/issues/1260.

### DIFF
--- a/connectivity/check/policy.go
+++ b/connectivity/check/policy.go
@@ -32,6 +32,13 @@ const (
 
 	ExitCurlHTTPError ExitCode = 22
 	ExitCurlTimeout   ExitCode = 28
+
+	/* How many times we should retry getting the policy revisions before
+	 * giving up. We want to reduce the likelihood that a connectivity blip
+	 * will prevent us from removing policies (dependent on revisions today)
+	 * because that may then cause subsequent tests to fail.
+	 */
+	getPolicyRevisionRetries = 3
 )
 
 func (e ExitCode) String() string {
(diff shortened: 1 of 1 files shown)


connectivity/check/policy.go
```diff
@@ -59,9 +67,17 @@ func (e ExitCode) Check(code uint8) bool {
...
 	revisions := make(map[Pod]int)
 	for _, cp := range ct.ciliumPods {
-		revision, err := getCiliumPolicyRevision(ctx, cp)
-		if err != nil {
-			return revisions, err
+		var revision int
+		var err error
+		for i := 1; i <= getPolicyRevisionRetries; i++ {
+			revision, err = getCiliumPolicyRevision(ctx, cp)
+			if err == spdystream.ErrTimeout {
```
squeed, line 74: Does it even matter why this failed? I would argue we should just retry regardless.

pchaigno: I'd rather be conservative at first given retries are usually a last resort. If we ever see it fail with something else, we can consider extending.

squeed: Sure, but given that this is a read-only request that, if failed, breaks the entire test run, this feels like unnecessary coupling to a single error type. apiservers can blip on, say, dns resolution, and others may not talk SPDY (e.g. ones behind a http/1 L7 proxy).

squeed (approved):

PR #1543, opened 2023-04-26
connectivity: Retry on inconclusive results

pchaigno:
When running the connectivity tests in AKS, we sometimes get interrupted commands that don't have any output [1]. Unfortunately, those commands then exit without any error and are therefore considered successful. We think this is caused by connectivity blips between Kubernetes components.

This commit adds a check for those inconclusive results. If we see a seemingly successful command with no output, we retry it until we get something conclusive. This works because all our test commands (curl, ping, nslookup) dump something to stdout.

1 - https://github.com/cilium/cilium/issues/22162

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -4,6 +4,7 @@
 package check
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"errors"
@@ -26,6 +27,10 @@ import (
 
 	"github.com/cilium/cilium-cli/connectivity/filters"
 	"github.com/cilium/cilium-cli/defaults"
+)
+
+const (
+	testCommandRetries = 3
 )
 
 // Action represents an individual action (e.g. a curl call) in a Scenario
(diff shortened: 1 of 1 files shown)


connectivity/check/action.go
```diff
@@ -236,12 +241,26 @@ func (a *Action) ExecInPod(ctx context.Context, cmd []string) {
...
+			continue
+		}
+		break
+	}
+	// Check for inconclusive results.
+	if err == nil && strings.TrimSpace(output.String()) == "" {
+		a.Failf("inconclusive results: command %q was successful but without output", cmdStr)
+	}
+
 	a.cmdOutput = output.String()
```
tklauser, line 264: Each call to `output.String` will allocate (which might be quite a bit in cases the 
command output is large) and we're calling it two more times below. Let's store the allocated string in `a.cmdOutput` directly after executing the command and reuse it:

```suggestion
		output, err = pod.K8sClient.ExecInPod(ctx,
			pod.Pod.Namespace, pod.Pod.Name, pod.Pod.Labels["name"], cmd)
		a.cmdOutput = output.String()
		// Check for inconclusive results.
		if err == nil && strings.TrimSpace(a.cmdOutput) == "" {
			a.Debugf("retrying command %s due to inconclusive results", cmdStr)
			continue
		}
		break
	}
	// Check for inconclusive results.
	if err == nil && strings.TrimSpace(a.cmdOutput) == "" {
		a.Failf("inconclusive results: command %q was successful but without output", cmdStr)
	}

```

tklauser (changes requested):
> This commit adds a check for those inconclusive results. If we see a seemingly successful command with no output, we retry it until we get something conclusive. This works because all our test commands (curl, ping, nslookup) dump something to stdout.

I'm a bit worried that at one point we start using commands that don't output to stdout and that will then fail due to this logic. But I don't know a better way to work around the issue we're seeing in cilium/cilium#22162. So let's give this a try and worry about this eventuality once we actually have such commands. Might be worth to add the above as a comment to the retry loop in `(*Action).ExecInPod`.

pchaigno:
> I'm a bit worried that at one point we start using commands that don't output to stdout and that will then fail due to this logic. But I don't know a better way to work around the issue we're seeing in [cilium/cilium#22162](https://github.com/cilium/cilium/issues/22162). So let's give this a try and worry about this eventuality once we actually have such commands. Might be worth to add the above as a comment to the retry loop in `(*Action).ExecInPod`.

I have the same concern. If that assumption doesn't hold anymore in the future, we should at least see a clear failure. It's also why I added a separate error message for the "inconclusive results": I'm hoping that makes the failure cause clearer if the assumption doesn't hold anymore.

I've added a comment above the retry loop.

tklauser (approved):
Thanks!

tklauser:
multicluster (helm) workflow is failing due to https://github.com/cilium/cilium-cli/issues/1554: https://github.com/cilium/cilium-cli/actions/runs/4829721713/jobs/8605084154?pr=1543

All other workflows passed. Merging.

PR #1547, opened 2023-04-27
connectivity: test accessing NodePort from outside with L7 policy

jschwinger233:
This test case covers https://github.com/cilium/cilium/issues/21954. 

A new policy `echo-ingress-l7-policy-from-anywhere` is added to allow HTTP GET / on echo pods from outside.

Use `cilium connectivity test --test north-south-loadbalancing --datapath` to run this test.

Signed-off-by: Zhichuan Liang <gray.isovalent.com>

### DIFF
--- a/connectivity/manifests/echo-ingress-l7-http-from-anywhere.yaml
+++ b/connectivity/manifests/echo-ingress-l7-http-from-anywhere.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: "cilium.io/v2"
+kind: CiliumNetworkPolicy
+metadata:
+  name: "echo-ingress-l7-http-from-anywhere"
+spec:
+  description: "Allow all to GET / on echo"
+  endpointSelector:
+    matchLabels:
+      kind: echo
+  ingress:
+  - toPorts:
+    - ports:
+      - port: "8080"
+        protocol: TCP
+      rules:
+        http:
+          - method: "GET"
+            path: "/$"
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -125,6 +125,9 @@ var (
 	//go:embed manifests/echo-ingress-l7-http.yaml
 	echoIngressL7HTTPPolicyYAML string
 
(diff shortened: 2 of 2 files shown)


connectivity/manifests/echo-ingress-l7-http.yaml
```diff
@@ -9,9 +9,11 @@ spec:
     matchLabels:
       kind: echo
   ingress:
-  # Only allow 'other' client to make a GET /public requests.
+  # Only allow 'other' client and world to make a GET /public requests.
   # Allow GET /private' only if a particular HTTP header is set.
   # Disallow L3 traffic for now, flow matcher doesn't yet support L7 drops.
+  - fromEntities:
+    - world
```
pchaigno, line 16: Watch out! That is allowing everything from `world` clients and not just 8080 with the below HTTP path.

pchaigno: I also think it would be best to have a separate policy. Extending this one to allow `world` increases the risk of false negatives. The `world` identity is often our fallback identity when we fail to look up a pod identity. So if that ever happens with the existing test and we've extended the policy like done here, we will miss a failure.

jschwinger233: Of course, thank you Paul!

jschwinger233:
Please note cilium can't pass the new added test case until https://github.com/cilium/cilium/pull/24882 is merged.

brb:
> Please note cilium can't pass the new added test case until https://github.com/cilium/cilium/pull/24882 is merged.

Nice! I assume you ran the tests locally with the fix?

jschwinger233:
> I assume you ran the tests locally with the fix?

Yes!

brb (approved):

brb:
Added `dont-merge/blocked` until https://github.com/cilium/cilium/pull/24882 has landed. Otherwise, the `ci-datapath` will be broken, as it's still using the `main` branch of `cilium-cli`.

pchaigno (approved):

squeed (approved):

PR #1550, opened 2023-04-27
Allow the Helm timeout to be set on uninstall

meyskens:
When running Helm in the CI setup i came against the limit that a cleanup of the tests namespace took longer than the Helm default of 5 minutes. However the timeout option was not exposed in out options thus could not be changed.

This PR exposes the `--timeout` helm option on uninstall when running in Helm mode.

### DIFF
--- a/internal/cli/cmd/install.go
+++ b/internal/cli/cmd/install.go
@@ -325,6 +325,8 @@ func newCmdUninstallWithHelm() *cobra.Command {
 	}
 
 	addCommonUninstallFlags(cmd, &params)
+
+	cmd.Flags().DurationVar(&params.Timeout, "timeout", defaults.UninstallTimeout, "Maximum time to wait resources to be deleted")
 	return cmd
 }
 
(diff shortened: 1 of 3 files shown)


internal/cli/cmd/install.go
```diff
@@ -325,6 +325,8 @@ func newCmdUninstallWithHelm() *cobra.Command {
 	}
 
 	addCommonUninstallFlags(cmd, &params)
+
+	cmd.Flags().DurationVar(&params.Timeout, "timeout", defaults.UninstallTimeout, "Maximum time to wait resources to be deleted")
```
asauber, line 329:
```suggestion
	cmd.Flags().DurationVar(&params.Timeout, "timeout", defaults.UninstallTimeout, "Maximum time to wait for resources to be deleted")
```

internal/cli/cmd/install.go
```diff
@@ -325,6 +325,8 @@ func newCmdUninstallWithHelm() *cobra.Command {
 	}
 
 	addCommonUninstallFlags(cmd, &params)
+
```
asauber, line 328: nit: for consistency with the rest of the file, there shouldn't be a newline here

asauber (approved):

michi-covalent (approved):

PR #1551, opened 2023-04-27
helm mode: add additional deprecated secret logic, and fix `clustermesh connect` for helm mode

asauber:
## Motivation

Previously, we had logic to look for deprecated names like `clustermesh-apiserver-client-certs` when `clustermesh-apiserver-client-cert` was expected. With helm-based installations, we now need to look for `clustermesh-apiserver-client-cert` when `clustermesh-apiserver-remote-cert` is expected.

## Implementation

This implements that concpet recursively, looking for the "even more deprecated" name for secrets if applicable.

## Example Runs

Two example runs that exercise the logic:

Using one level of deprecated names to find the secret
```
✅ ClusterMesh enabled!
Trying to get secret clustermesh-apiserver-remote-cert by deprecated name clustermesh-apiserver-client-cert
⚠️ Deprecated secret name "clustermesh-apiserver-client-cert", should be changed to "clustermesh-apiserver-remote-cert"
⚠️  Service type NodePort detected! Service may fail when nodes are removed from the cluster!
✅ Cluster access information is available:
  - 172.18.0.4:30212
```

Using two levels of deprecated names to find the secret
```
✅ ClusterMesh enabled!
Trying to get secret clustermesh-apiserver-remote-cert by deprecated name clustermesh-apiserver-client-cert
Trying to get secret clustermesh-apiserver-client-cert by deprecated name clustermesh-apiserver-client-certs
⚠️ Deprecated secret name "clustermesh-apiserver-client-certs", should be changed to "clustermesh-apiserver-remote-cert"
⚠️  Service type NodePort detected! Service may fail when nodes are removed from the cluster!
✅ Cluster access information is available:
  - 172.18.0.4:30652
```

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -662,6 +662,8 @@ func (ai *accessInformation) validate() bool {
 
 func getDeprecatedName(secretName string) string {
 	switch secretName {
+	case defaults.ClusterMeshRemoteSecretName:
+		return defaults.ClusterMeshClientSecretName
 	case defaults.ClusterMeshServerSecretName,
 		defaults.ClusterMeshAdminSecretName,
 		defaults.ClusterMeshClientSecretName,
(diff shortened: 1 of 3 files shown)


clustermesh/clustermesh.go
```diff
@@ -672,6 +674,27 @@ func getDeprecatedName(secretName string) string {
...
+
+	deprecatedSecretName := getDeprecatedName(secretName)
+	if deprecatedSecretName == "" {
+		return nil, fmt.Errorf("unable to get secret %q and no deprecated names to try", secretName)
+	}
+
+	k.Log("Trying to get secret %s by deprecated name %s", secretName, deprecatedSecretName)
+
+	secret, err := client.GetSecret(ctx, k.params.Namespace, deprecatedSecretName, metav1.GetOptions{})
+	if err != nil {
```
YutaroHayakawa, line 689: Should we check the error kind here? Currently, we don't distinguish missing secrets and other errors.

asauber: There are many instances in this file of the error type not being checked. I think that cleanup would be best left to another PR.

In this case among those in this file, I think it's least critical to check for errors, because Kubernetes internal APIs are relatively unstable, so if there is an error type that we fail to check for which has a similar meaning to "not found" (e.g. malformed content of some type, which could indicate that we _should_ check for a deprecated name), then we should at least try to request the deprecated name.

michi-covalent (approved):

michi-covalent:
@joamaki ping!

joestringer:
How do we trigger the required test `helm-upgrade-clustermesh`?

asauber:
@joestringer That workflow was merged since this PR was opened. I'll try a rebase.

YutaroHayakawa (approved):

PR #1552, opened 2023-04-27
sysdump: Fix writing to absolute path

pchaigno:
Writing the sysdump to an absolute path doesn't work because we always append the current directory to the output filename:

    $ ./cilium sysdump --output-filename /tmp/cilium-sysdump
    [...]
    ✅ The sysdump has been saved to /home/paul/cilium-cli/tmp/cilium-sysdump.zip

Let's remove the code that does that.

Fixes: 77a1f2389 ("sysdump: initial sysdump implementation")

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -1205,11 +1205,7 @@ func (c *Collector) Run() error {
 
 	// Create the zip file in the current directory.
 	c.log("🗳 Compiling sysdump")
-	p, err := os.Getwd()
-	if err != nil {
-		return fmt.Errorf("failed to get current directory: %w", err)
-	}
-	f := filepath.Join(p, c.replaceTimestamp(c.Options.OutputFileName)+".zip")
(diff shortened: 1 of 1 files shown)


tklauser (changes requested):
Running `./cilium sysdump --output-filename /tmp/cilium-sysdump` built from this PR I get:

```
% ./cilium sysdump --output-filename /tmp/cilium-sysdump
🔍 Collecting sysdump with cilium-cli version: v0.14.0-2-g34e04a38017f, args: [sysdump --output-filename /tmp/cilium-sysdump]
[...]
✅ The sysdump has been saved to /tmp/cilium-sysdump.zip
```

I'd expect the file to be written to `/tmp/cilium-sysdump/cilium-sysdump.zip` though.

tklauser (approved):
Sorry I was confused, never mind my previous comment, `--output-filename` should actually be the filename without extension:

```
      --output-filename string                        The name of the resulting file (without extension)
                                                      '<ts>' can be used as the placeholder for the timestamp (default "cilium-sysdump-<ts>")
```

PR #1559, opened 2023-05-01
connectivity: Get rid of --datapath

brb:
Let's simplify the test suite, and run the "--datapath" tests as part of the regular connectivity tests. The built-in feature detection will skip those if the conditions cannot be met.


connectivity/suite.go
```diff
@@ -893,6 +871,23 @@ func Run(ctx context.Context, ct *check.ConnectivityTest) error {
...
 
+	ct.NewTest("north-south-loadbalancing").
+		WithFeatureRequirements(check.RequireFeatureEnabled(check.FeatureNodeWithoutCilium)).
+		WithScenarios(
+			tests.OutsideToNodePort(),
+		)
+	ct.NewTest("pod-to-pod-encryption").
+		WithFeatureRequirements(check.RequireFeatureEnabled(check.FeatureEncryptionPod)).
+		WithScenarios(
+			tests.PodToPodEncryption(),
```
pchaigno, line 882: IIRC, this test won't work on EKS because we have multiple native interfaces there. Is that fixed now or is the feature detection somehow taking that into account?

pchaigno (commented):
That means, once we update the CLI in cilium/cilium, the pod-to-pod encryption test will run on GKE, EKS, and AKS. Did you check that it already works?

brb:
Preceded by https://github.com/cilium/cilium-cli/pull/1579.

PR #1561, opened 2023-05-01
Replace versioncheck.Version with semver.ParseTolerant

michi-covalent:
ParseCiliumVersion uses versioncheck.Version from Cilium's versioncheck
package ([1]), which ignores certain pre-release identifiers. Silently
ignoring certain pre-release identifiers can be confusing since you
might end up installing a different version of Cilium than what you
specified.

This commit replaces versioncheck.Version with semver.ParseTolerant ([2])
in ParseCiliumVersion function so that cilium-cli interprets the version
provided via the --version flag as is.

[1]: https://pkg.go.dev/github.com/cilium/cilium/pkg/versioncheck#Version
[2]: https://pkg.go.dev/github.com/blang/semver#ParseTolerant

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
-	"github.com/cilium/cilium/pkg/versioncheck"
 
 	"github.com/cilium/cilium-cli/defaults"
 )
@@ -26,8 +25,7 @@ func CheckVersion(version string) error {
 }
 
 func ParseCiliumVersion(version string) (semver.Version, error) {
-	ersion := strings.TrimPrefix(version, "v")
-	return versioncheck.Version(ersion)
+	return semver.ParseTolerant(version)
 }
 
 type ImagePathMode int
--- a/internal/utils/utils_test.go
+++ b/internal/utils/utils_test.go
@@ -76,6 +76,19 @@ func TestParseCiliumVersion(t *testing.T) {
 			version: "v1.9.99",
 			want:    semver.Version{Major: 1, Minor: 9, Patch: 99},
 		},
+		{
+			name:    "valid-pre-release-version",
+			version: "1.13.90-dev.1234-main-5678abcd",
+			want: semver.Version{
+				Major: 1,
+				Minor: 13,
+				Patch: 90,
+				Pre: []semver.PRVersion{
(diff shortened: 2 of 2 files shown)


rolinh (approved):

internal/utils/utils_test.go
```diff
@@ -55,44 +52,6 @@ func TestCheckVersion(t *testing.T) {
 	}
 }
 
-func TestParseCiliumVersion(t *testing.T) {
```
chancez, line 58: Okay, I don't really want to test someone else's code, but I still think it might be useful for us to test our expected values with `semver.ParseTolerant`, just to ensure the values we pass to it actually work as expected.

internal/utils/utils.go
```diff
@@ -25,11 +22,6 @@ func CheckVersion(version string) error {
 	return nil
 }
 
-func ParseCiliumVersion(version string) (semver.Version, error) {
```
chancez, line 28: Should we consider keeping this function and just modifying it to use `semver.ParseTolerant` instead of `versioncheck.Version`? Should we fix version check upstream instead? 

Also, I think we may still want the `strings.TrimPrefix`.

chancez (changes requested):
A few questions about the approach being taken.

internal/utils/utils_test.go
michi-covalent: sure i'll add some tests ✅

internal/utils/utils.go
michi-covalent: ok yeah i can do that.

> Also, I think we may still want the strings.TrimPrefix.

`semver.ParseTolerant` already trims the `v` prefix ✅

internal/utils/utils_test.go
```diff
@@ -76,6 +76,19 @@ func TestParseCiliumVersion(t *testing.T) {
 			version: "v1.9.99",
 			want:    semver.Version{Major: 1, Minor: 9, Patch: 99},
 		},
+		{
+			name:    "valid-pre-release-version",
+			version: "1.13.90-dev.1234-main-5678abcd",
```
chancez, line 81: Perfect, taken straight from our real versions.

chancez (approved):

squeed (approved):

tklauser (approved):

Issue #1567, opened 2023-05-02
Sysdump: improve experience for Gateway API + ingress


meyskens:
## Proposal / RFE

Currently the sysdump generates very useful data used to debug issues with Cilium. It includes many aspects and useful resources to get an idea what is going on in a cluster.

At this point there is no or very limited support for the Ingress & Gateway API features within Cilium. To improve this experience it would be good to have a copy of these resources exported including useful debug info analog to the other parts. The sysdump should be extended to include these resources when available.

Useful resources (some may be present already, this is a suggestion for now):
* IngressClass
* Ingress
* GatewayClass
* Gateway
* HTTPRoute
* TLSRoute
* GRPCRoute
* TCPRoute
* UDPRoute
* Namespaces for reference grants
* Cilium Envoy Config
* Envoy configuration state? 
* Others? 

It could also be useful to think about how these are exported (seperate folder?) for ease of use

meyskens:
/assign

(don't think that will work but i have no powers here)

meyskens:
Related to this is https://github.com/cilium/cilium-cli/pull/1546 which already covers things to get info from Envoy when deployed in DaemonSet

PR #1569, opened 2023-05-03
Add GatewayAPI resources to sysdump

meyskens:
This adds all GatewayAPI objects to the sysdump as well as the IngressClass object. This will be helpful for troubleshooting GatewayAPI related issues.
The objects are dumped into YAML if available prefixed with gatewayapi- for ease of use.

Fixes https://github.com/cilium/cilium-cli/issues/1567

### DIFF
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -798,6 +798,10 @@ func (c *Client) ListEndpoints(ctx context.Context, o metav1.ListOptions) (*core
 	return c.Clientset.CoreV1().Endpoints(corev1.NamespaceAll).List(ctx, o)
 }
 
+func (c *Client) ListIngressClasses(ctx context.Context, o metav1.ListOptions) (*networkingv1.IngressClassList, error) {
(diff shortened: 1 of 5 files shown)


youngnick (approved):

squeed:
Got a lint failure:

```
  Warning: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
  Warning: unused-parameter: parameter 'ctx' seems to be unused, consider removing or renaming it as _ (revive)
```

squeed:
Is there any chance that there is secret data that will need to be scrubbed? I dunno, I always think of Ingress and brethren as having scary annotations.

squeed (approved):
Looks good on fixing lint; see question about potential secrets (probably not an issue).

youngnick:
No, we've specifically designed Gateway API to only use structured data, and to keep secret stuff in Secrets. I hate annotations for Ingress stuff too.

PR #1575, opened 2023-05-04
Add Prometheus metrics validation

doniacld:
## Motivation
Add the possibility to verify metrics from the Cilium agent pod and validate it has the expected behaviour before and after an action. 

## Changes

- [commit 1](https://github.com/cilium/cilium-cli/pull/1575/commits/36277d8e22cc18328b8422eb2ac91939aa89e451) Add a k8s dialer to be able to port forward `/metrics` endpoint outside of the pod
- [commit 2](https://github.com/cilium/cilium-cli/pull/1575/commits/8d5197c451fd056e345337a3d775ee059f2d8a75) Add the logic to call `/metrics` endpoint and parse the metrics to retrieve them into a nice usable open metrics format
- [commit 3](https://github.com/cilium/cilium-cli/pull/1575/commits/86526acefae7242d946281b0719f1fa494cb3fc5) `policy.go` is a pretty long file now, let's move `Result` with its associated methods in its own file.
- [commit 4](https://github.com/cilium/cilium-cli/pull/1575/commits/36757130512333e191d4dd380e00a3335eff5f94) Make the call to collect metrics in the action
- [commit 5](https://github.com/cilium/cilium-cli/pull/1575/commits/7d5a902f35d8cded916db57600f3ffee6d76dae8) Validate the metrics by defining a new type `metricsCompareFunc` which will compare metrics in a generic way.
- [commit 6](https://github.com/cilium/cilium-cli/pull/1575/commits/18bb78f42845fcedddbf9af6a12d6e562c1f44c4) Add a new scenario very basic with a simple metric to verify the feature is working

## Usage

Implement the operation you want to check on your metrics on `Result`:
```go
func (r Result) ExpectMetricsIncrease(source MetricsSource, metrics ...string) Result {
```

Then use it directly in your expectations by passing a MetricSource type (cilium-agent, cilium-operator...) and the name of the metric you want to check:
```go
return check.ResultOK.ExpectMetricsIncrease(ct.CiliumAgentMetrics(), "cilium_forward_count_total")
```

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -78,21 +78,25 @@ type Action struct {
 
 	// Output from action if there is any
 	cmdOutput string
+
+	// metricsPerSource collected at the initialisation of an Action.
+	metricsPerSource promMetricsPerSource
 }
 
 func newAction(t *Test, name string, s Scenario, src *Pod, dst TestPeer, ipFam IPFamily) *Action {
 	return &Action{
-		name:         name,
-		test:         t,
-		scenario:     s,
-		src:          src,
-		dst:          dst,
-		ipFam:        ipFam,
-		CollectFlows: true,
-		flowResults:  map[TestPeer]FlowRequirementResults{},
-		started:      time.Now(),
-		failed:       false,
-		cmdOutput:    "",
+		name:             name,
+		test:             t,
+		scenario:         s,
+		src:              src,
+		dst:              dst,
+		ipFam:            ipFam,
+		CollectFlows:     true,
+		flowResults:      map[TestPeer]FlowRequirementResults{},
+		started:          time.Now(),
+		failed:           false,
+		cmdOutput:        "",
+		metricsPerSource: make(promMetricsPerSource),
 	}
 }
 
@@ -138,6 +142,21 @@ func (a *Action) Run(f func(*Action)) {
 
 	// Emit unbuffered progress indicator.
 	a.test.progress()
+
+	// Retrieve Prometheus metrics only if there are expectations.
+	for _, m := range a.expIngress.Metrics {
+		err := a.collectMetricsPerSource(m)
+		if err != nil {
+			a.Logf("❌ Failed to collect metrics for ingress from source %s: %w", m.Source, err)
+		}
+
+	}
+	for _, m := range a.expEgress.Metrics {
+		err := a.collectMetricsPerSource(m)
+		if err != nil {
+			a.Logf("❌ Failed to collect metrics for egress from source %s: %w", m.Source, err)
+		}
+	}
 
 	// Only perform flow validation if a Hubble Relay connection is available.
 	if a.test.ctx.params.Hubble && a.CollectFlows {
@@ -198,6 +217,18 @@ func (a *Action) Run(f func(*Action)) {
 			a.test.collectSysdump()
 		}
 	}
+}
+
+// collectMetricsPerSource retrieves metrics for the given source.
+func (a *Action) collectMetricsPerSource(m MetricsResult) error {
+	if _, ok := a.metricsPerSource[m.Source.Name]; !ok {
+		metrics, err := a.collectPrometheusMetrics(m.Source)
+		if err != nil {
+			return err
+		}
+		a.metricsPerSource[m.Source.Name] = metrics
+	}
+	return nil
 }
 
 // fail marks the Action as failed.
@@ -960,6 +991,76 @@ func (a *Action) validateFlowsForPeer(ctx context.Context, reqs []filters.FlowSe
 		case <-ctx.Done():
 			a.Fail("Aborting flow matching:", ctx.Err())
 			return res
+		}
+	}
+}
+func (a *Action) GetEgressMetricsRequirements() []MetricsResult {
+	return a.expEgress.Metrics
+}
+func (a *Action) GetIngressMetricsRequirements() []MetricsResult {
+	return a.expIngress.Metrics
+}
+
+const (
+	metricsCollectionTimeout      = time.Second * 30
+	metricCollectionIntervalRetry = time.Second * 5
+)
+
+// ValidateMetrics confronts the expected metrics against the last ones retrieves.
+func (a *Action) ValidateMetrics(ctx context.Context, pod Pod, results []MetricsResult) {
+	node := pod.NodeName()
+
+	if len(results) == 0 {
+		// Early exit.
+		a.Debugf("no metrics requirements to validate for this node %s", node)
+		return
+	}
+
+	for _, r := range results {
+		a.validateMetric(ctx, node, r)
+	}
+}
+
+func (a *Action) validateMetric(ctx context.Context, node string, result MetricsResult) {
+	if result.IsEmpty() {
+		a.Debugf("there is no source configured to retrieve metrics for node %s", node)
+		return
+	}
+
+	// ctx is used here as the duration of the retries to collect metrics.
+	ctx, cancel := context.WithTimeout(ctx, metricsCollectionTimeout)
+	defer cancel()
+
+	// ticker here is used as the interval duration before the next retry.
+	ticker := time.NewTicker(metricCollectionIntervalRetry)
+	defer ticker.Stop()
+
+	for {
+		// Collect the new metrics.
+		newMetrics, err := a.collectPrometheusMetricsForNode(result.Source, node)
+		if err != nil {
+			a.Failf("failed to collect new metrics on node %s: %w", node, err)
+			return
+		}
+
+		// Check the metrics comparing the first retrieved metrics and the new ones.
+		err = result.Assert(a.metricsPerSource[result.Source.Name][node], newMetrics)
+		if err != nil {
+			a.Debugf("failed to check metrics on node %s: %s\n", node, err)
+		} else {
+			// Metrics check succeed, let's exit.
+			a.Debugf("checked metrics properly on node %s: %s\n", node)
+			return
+
+		}
+
+		select {
+		case <-ctx.Done():
+			// Context timeout is reached, let's exit.
+			a.Failf("failed to collect metrics on node %s, context timeout: %w", node, ctx.Err().Error())
+			return
+		case <-ticker.C:
+			// Ticker is delivered, let's retry.
 		}
 	}
 }
--- a/connectivity/check/metrics.go
+++ b/connectivity/check/metrics.go
@@ -47,6 +47,21 @@ func (a *Action) collectPrometheusMetrics(source MetricsSource) (promMetricsPerN
 	return m, nil
 }
 
+// collectPrometheusMetricsForNode retrieves all the metrics for a source on a particular node.
+func (a *Action) collectPrometheusMetricsForNode(source MetricsSource, node string) (promMetricsFamily, error) {
+	for _, pod := range source.Pods {
+		if pod.NodeName() == node {
+			metrics, err := a.collectMetricsForPod(pod, source.Port)
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve prometheus metrics for pod %s on node %s: %w", pod.Name(), pod.NodeName(), err)
+			}
+			return metrics, nil
+		}
+	}
+
+	return promMetricsFamily{}, nil
+}
+
 // collectMetricsForPod retrieves the metrics for one pod.
 func (a *Action) collectMetricsForPod(pod Pod, port string) (promMetricsFamily, error) {
 	// The context is in charge if closing the port-forward when it is cancelled.
--- a/connectivity/check/metrics.go
+++ b/connectivity/check/metrics.go
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package check
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+
+	"github.com/cilium/cilium-cli/k8s"
+)
+
+// metricsURLFormat is the path format to retrieve the metrics on the
+const metricsURLFormat = "http://localhost:%d/metrics"
+
+// promMetricsPerSource holds all the metrics per source name per node.
+// sources e.g.: cilium agent, cilium operator, hubble
+type promMetricsPerSource map[string]promMetricsPerNode
+
+// promMetricsPerNode stores for each node Prometheus Metrics.
+type promMetricsPerNode map[string]promMetricsFamily
+
+// promMetricsFamily holds Prometheus metrics per metric name.
+type promMetricsFamily map[string]*dto.MetricFamily
+
+// collectPrometheusMetrics retrieves the Prometheus metrics by
+// port-forwarding the Prometheus port of each source pod and calling /metrics endpoint.
+func (a *Action) collectPrometheusMetrics(source MetricsSource) (promMetricsPerNode, error) {
+	m := make(promMetricsPerNode)
+
+	// Retrieve metrics for all Cilium pods.
+	for _, pod := range source.Pods {
+		metrics, err := a.collectMetricsForPod(pod, source.Port)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve prometheus metrics for pod %s: %w", pod.Name(), err)
+		}
+
+		// Store metrics per node for the ease of use when validating.
+		m[pod.NodeName()] = metrics
+	}
+
+	return m, nil
+}
+
+// collectMetricsForPod retrieves the metrics for one pod.
+func (a *Action) collectMetricsForPod(pod Pod, port string) (promMetricsFamily, error) {
+	// The context is in charge if closing the port-forward when it is cancelled.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	p := k8s.PortForwardParameters{
+		Namespace:  pod.Namespace(),
+		Pod:        pod.NameWithoutNamespace(),
+		Ports:      []string{fmt.Sprintf(":%s", port)},
+		Addresses:  nil, // default is localhost
+		OutWriters: k8s.OutWriters{Out: &debugWriter{ct: a.test.ctx}, ErrOut: &warnWriter{ct: a.test.ctx}},
+	}
+
+	// Call the k8s dialer to port forward,
+	// a random port will be generated to avoid conflict.
+	res, err := pod.K8sClient.PortForward(ctx, p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to port forward: %w", err)
+	}
+
+	// Call the metrics path on the retrieved local port.
+	url := fmt.Sprintf(metricsURLFormat, res.ForwardedPorts[0].Local)
+	resp, err := http.Get(url) //nolint:gosec
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve metrics: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Convert the text output into handy format metrics.
+	metrics, err := parseMetrics(resp.Body)
+	if err != nil {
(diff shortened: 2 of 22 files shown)


connectivity/check/policy.go
```diff
@@ -221,6 +222,31 @@ func authRequiredDropReason(flow *flowpb.Flow) bool {
...
 
+// expectMetricsToIncrease compares metrics retrieved before any action were run and after;
+// may return an error if metrics did not increase.
+func expectMetricsToIncrease(metrics []string) metricsCompareFunc {
+	return func(before, after map[string]*dto.MetricFamily) error {
+		for _, metricName := range metrics {
+			bValue, ok := before[metricName]
+			if !ok {
+				// TODO DONIA: should I return an error here, should I store the error and range over all the metrics?
+				return fmt.Errorf("failed to retrieve metric %s in first retrieved metrics", metricName)
```
gandro, line 233: Yeah, I think it's better to use `errors.Join` here to report on all metrics. It will make debugging test runs easier if you have more information on what worked and what didn't.

connectivity/check/policy.go
```diff
@@ -333,6 +359,12 @@ var (
 		Drop:     true,
 		ExitCode: ExitCurlHTTPError,
 	}
+
+	// ResultOKWithMetricForwardCountTotal expects OK and
+	// a metric of the number of total of forward count increases.
+	ResultOKWithMetricForwardCountTotal = Result{
```
gandro, line 365: I think we should make this a method on `Result`, that also takes in the metric names as an argument. So in expectations, you could use something like the following in `suite.go`:

```go
WithExpectations(func(a *check.Action) (egress, ingress check.Result) {
	return check.ResultOK.ExpectMetricsIncrease("cilium_forward_count_total", "cilium_foo_bar_total"), check.ResultOk
})
```

This allows you to easily test for drops too:
```go
WithExpectations(func(a *check.Action) (egress, ingress check.Result) {
	return check.ResultDrop.ExpectMetricsIncrease("cilium_dropped_count_total"), check.ResultDrop
})
```

connectivity/check/policy.go
```diff
@@ -399,6 +437,9 @@ func (r Result) String() string {
 		ret += "-"
 		ret += r.HTTP.Status
 	}
+	if r.MetricsCompareFunc != nil {
+		ret += fmt.Sprintf("-metricsCompareFunc")
```
gandro, line 441: nit:

```suggestion
		ret += "-MetricsCompareFunc"
```

k8s/dialer.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, err
+	}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, req.URL())
+	stopChan, readyChan := make(chan struct{}, 1), make(chan struct{}, 1)
+	if len(p.Addresses) == 0 {
+		p.Addresses = []string{"localhost"}
+	}
+
+	pw, err := portforward.NewOnAddresses(dialer, p.Addresses, p.Ports, stopChan, readyChan, os.Stdout, os.Stderr)
```
gandro, line 54: I think the output of Stdout and Stderr are only useful when debugging. To avoid this pouting the output, we should make those two `io.Writers` a parameter in `PortForwardParameters`.

And then in `collectPrometheusMetrics`, implement the `io.Writer` interface on top of the logging infrastructure in `connectivity/check/logging.go`. For example:

```
type debugWriter struct {
   ct *ConnectivityTest
}

func (d *debugWriter) Write(b []byte) (int, error) {
   d.ct.Debug(string(b))
   return len(b), nil
}

type warnWriter struct {
   ct *ConnectivityTest
}

func (w *warnWriter) Write(b []byte) (int, error) {
   w.ct.Warn(string(b))
   return len(b), nil
}
```

And then pass in a debugWriter instead of stdout, and a warnWriter instead of stderr.

connectivity/check/check.go
```diff
@@ -72,6 +72,9 @@ type Parameters struct {
 
 	CollectSysdumpOnFailure bool
 	SysdumpOptions          sysdump.Options
+
+	// PrometheusServeAddr is the exposed port on cilium agent to retrieve metrics.
+	PrometheusServeAddr string // TODO DONIA: Not sure where to store it
```
gandro, line 77: I think should define an interface here, that tells us what pods to collect from (e.g. agent, operator, hubble etc) on which port.

It could be something like:

```suggestion
	CollectMetrics []MetricsSource
```

Where the interface is:

```go
type MetricsSource interface {
   func Pods() ([]*corev1.Pod, error)
   func Port() string
}
```

And then implement this interface e.g. for the agent metrics. The expectations also should specify where they want to collect metrics from and what metrics they actually care about.

```go
WithExpectations(func(a *check.Action) (egress, ingress check.Result) {
	return check.ResultOK.ExpectMetricsIncrease(a.CiliumAgentMetrics(), "cilium_forward_count_total"), check.ResultOk
})
```

Where as `CiliumAgentMetrics` defines a `MetricsSource`. This is just a rough sketch and needs to be fleshed out (i.e. I don't expect my proposal to work as is), but generally speaking the idea here should be that expectations declare what metrics they are interested in, and that in turn tells the framework not only that it needs to collect metrics, but also where it needs to collect metrics from.

Happy to discuss this further offline.

gandro (changes requested):
Awesome work, thanks a lot! I've left some feedback regarding the open TODOs.

connectivity/check/metrics_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParsePromMetrics(t *testing.T) {
+	input := `
+# HELP isovalent_external_dns_proxy_policy_l7_total Number of total proxy requests handled
+# TYPE isovalent_external_dns_proxy_policy_l7_total counter
+isovalent_external_dns_proxy_policy_l7_total{rule="forwarded"} 132
+isovalent_external_dns_proxy_policy_l7_total{rule="received"} 132
+`
```
tklauser, line 18: Could we please use metrics here that are not proprietary, i.e. that are available in an  Cilium OSS version?

doniacld: Oups! Yes of course!

connectivity/check/policy.go
doniacld: Fixed, thanks!

connectivity/check/metrics_test.go
doniacld: Fixed.

connectivity/check/policy.go
doniacld: Fixed.

k8s/dialer.go
doniacld: Fixed but I still have a log from the k8s go client port forward library:
https://github.com/kubernetes/client-go/blob/v0.27.1/tools/portforward/portforward.go#L312

The line of code:
```
runtime.HandleError(fmt.Errorf("error accepting connection on port %d: %v", port.Local, err))
```

The log (and there some lines...):
```
E0508 15:06:18.542005   72309 portforward.go:417] error closing listener: close tcp4 127.0.0.1:64508: use of closed network connection
```

I would be happy to discuss how to handle it.

connectivity/check/policy.go
doniacld: I did some refactoring to isolate more Result and its associated methods, which might not be the best idea. It is in a new commit (the last one) for the moment but I will rework the history if we go with that.

k8s/dialer.go
doniacld: Fixed.

connectivity/check/policy.go
doniacld: Fixed.

k8s/dialer.go
```diff
@@ -0,0 +1,100 @@
...
+	ErrOut io.Writer
+}
+
+// PortForwardResult are the ports that have been forwarded.
+type PortForwardResult struct {
+	ForwardedPorts []ForwardedPort
+}
+
+// PortForward executes in a goroutine a port forward command.
+// To kill the thread, you should use the context by cancelling it.
```
gandro, line 46: nit:

`To stop the port-forwarding, use the context by cancelling it`

connectivity/check/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+type promMetricsPerSource map[string]promMetricsPerNode
+
+// promMetricsPerNode stores for each node Prometheus Metrics.
+type promMetricsPerNode map[string]promMetricsFamily
+
+// promMetricsFamily holds Prometheus metrics per metric name.
+type promMetricsFamily map[string]*dto.MetricFamily
+
+// collectPrometheusMetrics retrieves the Prometheus metrics by
+// port-forwarding the Prometheus port on a Cilium node and calling /metrics endpoint.
```
gandro, line 30:
```suggestion
// port-forwarding the Prometheus port of each source pod and calling /metrics endpoint.
```

connectivity/check/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, fmt.Errorf("failed to parse metrics: %w", err)
+	}
+
+	return metrics, nil
+}
+
+// parseMetrics transforms the response from the call to prometheus metric endpoint
+// into a dto model MetricFamily.
+func parseMetrics(content []byte) (promMetricsFamily, error) {
+	reader := bytes.NewReader(content)
```
gandro, line 94: Given that `resp.Body` is already a reader, we should probably pass in `resp.Body` directly here, instead of turning it into `content []byte` and then back into a reader.

k8s/dialer.go
```diff
@@ -0,0 +1,100 @@
+package k8s
```
gandro, line 1: Missing copyright header:

```
// SPDX-License-Identifier: Apache-2.0
// Copyright Authors of Cilium
```

connectivity/check/metrics.go
```diff
@@ -0,0 +1,102 @@
+package check
```
gandro, line 1: Missing copyright header:

```
// SPDX-License-Identifier: Apache-2.0
// Copyright Authors of Cilium
```

connectivity/check/metrics_test.go
```diff
@@ -0,0 +1,48 @@
+package check
```
gandro, line 1: Missing copyright header

connectivity/check/metricssource.go
```diff
@@ -0,0 +1,56 @@
+package check
```
gandro, line 1: Missing copyright header

connectivity/check/metricssource_test.go
```diff
@@ -0,0 +1,81 @@
+package check
```
gandro, line 1: Missing copyright header

connectivity/check/metricssource.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// Retrieve the container port value for Prometheus.
+	for _, p := range ciliumPods {
+		// parse all the containers
+		for _, c := range p.Pod.Spec.Containers {
+			if c.Name == components.CiliumAgentName {
+				// parse all the container ports
+				for _, port := range c.Ports {
+					if port.Name == prometheusContainerPortName {
+						source.Port = fmt.Sprintf("%d", port.ContainerPort)
```
gandro, line 43: consider exiting the `range ciliumPods` loop early here once we found a port - alternatively, it might be cleaner to actually `(pod, port)` pairs in the `MetricsSouce`.

connectivity/check/metricssource.go
```diff
@@ -0,0 +1,56 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	// Prometheus port was not find, let's return an empty MetricsSource.
+	if source.Port == "" {
+		return MetricsSource{}
```
gandro, line 52: Consider defining a `EmptyMetricsSource` constant for this, so callers can compare to the constant to determine if the metrics source exists or not.

connectivity/check/action.go
```diff
@@ -139,6 +143,24 @@ func (a *Action) Run(f func(*Action)) {
 	// Emit unbuffered progress indicator.
 	a.test.progress()
 
+	// Retrieve Prometheus metrics only if there are expectations.
+	if a.expIngress.Metrics != nil {
```
gandro, line 147: No need for a nil check if we only iterate in the loop body

connectivity/check/policy.go
```diff
@@ -24,43 +24,12 @@ import (
 	"github.com/cilium/cilium-cli/k8s"
 )
 
-type ExitCode int16
```
gandro, line 27: Consider moving this refactor in a separate commit to make it easier on reviewers

connectivity/check/result.go
```diff
@@ -0,0 +1,127 @@
...
+	if r.HTTP.URL != "" {
+		ret += "-"
+		ret += r.HTTP.URL
+	}
+	if r.HTTP.Status != "" {
+		ret += "-"
+		ret += r.HTTP.Status
+	}
+	if r.Metrics != nil {
+		ret += "-metricsCompareFunc"
```
gandro, line 122: I think we should print the metric source names here.

connectivity/check/action.go
```diff
@@ -942,3 +977,68 @@ func (a *Action) validateFlowsForPeer(ctx context.Context, reqs []filters.FlowSe
...
+	ctx, cancel := context.WithTimeout(ctx, metricsCollectionTimeout)
+	defer cancel()
+
+	// ticker here is used as the interval duration between before the next retry.
+	ticker := time.NewTicker(metricCollectionIntervalRetry)
+	defer ticker.Stop()
+
+	for {
+		// Collect the new metrics.
+		newMetrics, err := a.collectPrometheusMetrics(result.Source)
```
gandro, line 1018: As an optimization, we could consider adding a `collectPrometheusMetricsForNode` function, that only port-forwards to a particular node

connectivity/tests/pod.go
```diff
@@ -59,6 +59,9 @@ func (s *podToPod) Run(ctx context.Context, t *check.Test) {
 
 					a.ValidateFlows(ctx, client, a.GetEgressRequirements(check.FlowParameters{}))
 					a.ValidateFlows(ctx, echo, a.GetIngressRequirements(check.FlowParameters{}))
+
+					a.ValidateMetrics(ctx, echo, a.GetIngressMetricsRequirements())
+					a.ValidateMetrics(ctx, echo, a.GetEgressMetricsRequirements())
```
gandro, line 64: We probably want to add this to all tests where it makes sense (check for calls to `a.ValidateFlows` on where to add)

connectivity/check/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	// Call the k8s dialer to port forward,
+	// a random port will be generated to avoid conflict.
+	res, err := pod.K8sClient.PortForward(ctx, p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to port forward: %w", err)
+	}
+
+	// Call the metrics path on the retrieved local port.
+	url := fmt.Sprintf(metricsURLFormat, res.ForwardedPorts[0].Local)
+	resp, err := http.Get(url)
```
gandro, line 71: Here is a linting error on this:
```
G107: Potential HTTP request made with variable url (gosec)
```

I think our usage is safe, so consider adding a ` //nolint:gosec` annotation here

gandro (changes requested):
Awesome! High-level this looks good to me. I did leave some minor feedback that should be addressed before we can merge. Thanks for working on this!

connectivity/check/metrics.go
doniacld: Fixed.

connectivity/check/metrics.go
doniacld: Good catch! 👁️ 👓

connectivity/check/metricssource.go
doniacld: I added the early exit, but I will be happy to discuss the interest of pairing pod, port :)

connectivity/check/metricssource.go
doniacld: Cannot define a constant on a non-primitive type, should I create a function?

connectivity/tests/pod.go
doniacld: I am pretty unsure of what I've done here, so please be careful in the review :)

connectivity/check/action.go
doniacld: Fixed.

connectivity/check/metricssource.go
doniacld: Done! Using a function on MetricsSource.

connectivity/check/metricssource.go
doniacld: For the record, we decided with @gandro to not pair a pod to a port for the moment but we could have a usecase in the future (e.g. cluster mesh use-case: different configuration of ports for a component through different clusters)

connectivity/check/metrics.go
```diff
@@ -53,7 +53,7 @@ func (a *Action) collectPrometheusMetrics(source MetricsSource) (promMetricsPerN
 }
 
 // collectMetricsForPod retrieves the metrics for one pod.
-func (a *Action) collectMetricsForPod(pod Pod, port string) (map[string]*dto.MetricFamily, error) {
```
gandro: nit: This change here seems to be part of the wrong commit

connectivity/check/result.go
```diff
@@ -105,8 +124,24 @@ func (r Result) String() string {
...
+	}
 	if r.ExitCode >= 0 && r.ExitCode <= 255 {
 		ret += fmt.Sprintf("-exit(%d)", r.ExitCode)
 	}
 	return ret
 }
+
+func displayMetrics(metrics []MetricsResult) string {
+	var ret string
+	ret += "-Sources("
```
gandro, line 139: nit: Let's call this `MetricSources(` or `Metrics(`. It's not clear from context otherwise that this refers to metrics.

gandro (commented):
Awesome, overall looks good to me! A couple minor nits remaining

connectivity/check/metrics.go
doniacld: Oupsy! Fixed!

connectivity/check/action.go
```diff
@@ -1027,7 +1027,7 @@ func (a *Action) validateMetric(ctx context.Context, node string, result Metrics
 	ctx, cancel := context.WithTimeout(ctx, metricsCollectionTimeout)
 	defer cancel()
 
-	// ticker here is used as the interval duration between before the next retry.
```
gandro: nit: I think typo fix should be in an earlier commit

gandro (approved):
Looks good to me! Two minor (non-blocking) nits remaining

connectivity/check/metrics.go
```diff
@@ -115,3 +115,23 @@ func parseMetrics(reader io.Reader) (promMetricsFamily, error) {
...
+func metricsIncrease(mf1, mf2 dto.MetricFamily) error {
+	metrics1 := mf1.GetMetric()
+	metrics2 := mf2.GetMetric()
+
+	if len(metrics1) != len(metrics2) {
+		return fmt.Errorf("metric %s has different length metrics 1: %d and metrics 2: %d", mf1.GetName(), len(metrics1), len(metrics2))
+	}
+
+	for i := range metrics1 {
+		value1 := metrics1[i].GetCounter().GetValue() // Here we assume that metrics are of Counter type.
```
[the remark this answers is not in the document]
doniacld: Is it fine to add these checks?

```go
if metrics1[i].GetCounter() == nil {
	return fmt.Errorf("metric %s is not a Counter: %v", mf1.GetName(), metrics1[i])
}
if metrics2[i].GetCounter() == nil {
	return fmt.Errorf("metric %s is not a Counter: %v", mf1.GetName(), metrics2[i])
}
```

brlbil (approved):
The CI part is 👍 Awesome work, thanks a lot!

youngnick (approved):

tklauser (approved):
Nice work!

PR #1576, opened 2023-05-04
Handle OCI repositories

michi-covalent:
Tweak newChartFromRemoteWithCache() to support OCI repos:

- chartRef must be the full OCI repository path.
- RepoURL must be empty.
- RegistryClient must be initialized.


asauber (approved):
LGTM

Issue #1566, opened 2023-05-02
Helm mode install doesn't support AKS


meyskens:
## Bug report

When installing Cilium on AKS with `CILIUM_CLI_MODE=helm` it will not work as `--azure-resource-group` is not implemented in this mode. 

**General Information**

- Cilium CLI version: v0.14.0

**How to reproduce the issue**
1. `CILIUM_CLI_MODE=helm cilium install --azure-resource-group "something"`

michi-covalent:
💯 thanks for filing the issue @meyskens i'll take care of this.

PR #1578, opened 2023-05-04
Add helm mode test coverage for AKS BYOCNI

michi-covalent:
Fixes: #1566

ended up adding an extra commit to make waitForRelay() more robust: https://github.com/cilium/cilium-cli/pull/1578/commits/e2e7202b3437c4b164f4e0f67d3468b40be40066 cc @cilium/sig-hubble 

### DIFF
--- a/.github/workflows/aks-byocni.yaml
+++ b/.github/workflows/aks-byocni.yaml
@@ -36,7 +36,6 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  name: ${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
   location: westeurope
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
   cilium_version: v1.13.2
@@ -47,6 +46,9 @@ jobs:
     if: ${{ github.repository == 'cilium/cilium-cli' }}
     runs-on: ubuntu-22.04
     timeout-minutes: 45
+    strategy:
+      matrix:
+        mode: ["classic", "helm"]
     steps:
       - name: Checkout
         uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
@@ -89,19 +91,20 @@ jobs:
 
           echo "sha=${SHA}" >> $GITHUB_OUTPUT
           echo "owner=${OWNER}" >> $GITHUB_OUTPUT
+          echo "name=${{ github.event.repository.name }}-${{ github.run_id }}-${{ matrix.mode }}" >> $GITHUB_OUTPUT
 
       - name: Create AKS cluster
         run: |
           # Create group
           az group create \
-            --name ${{ env.name }} \
+            --name ${{ steps.vars.outputs.name }} \
             --location ${{ env.location }} \
             --tags usage=${{ github.repository_owner }}-${{ github.event.repository.name }} owner=${{ steps.vars.outputs.owner }}
 
           # Create AKS cluster
           az aks create \
-            --resource-group ${{ env.name }} \
-            --name ${{ env.name }} \
+            --resource-group ${{ steps.vars.outputs.name }} \
+            --name ${{ steps.vars.outputs.name }} \
             --location ${{ env.location }} \
             --network-plugin none \
             --node-count 2 \
@@ -111,8 +114,8 @@ jobs:
       - name: Get cluster credentials
         run: |
           az aks get-credentials \
-            --resource-group ${{ env.name }} \
-            --name ${{ env.name }}
+            --resource-group ${{ steps.vars.outputs.name }} \
+            --name ${{ steps.vars.outputs.name }}
 
       - name: Create kubeconfig and load it in configmap
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/aks-byocni.yaml
```diff
@@ -14,7 +14,7 @@ on:
   #   will disappear from the PR checks: please provide a direct link to the
   #   successful workflow run (can be found from Actions tab) in a comment.
   #
-  # pull_request: {}
+  pull_request: {}
```
viktor-kurchenko, line 17: Should it be uncommented?

connectivity/check/action.go
```diff
@@ -721,18 +721,31 @@ func (a *Action) GetIngressRequirements(p FlowParameters) []filters.FlowSetRequi
...
+				return nil
+			}
+		} else {
+			a.Debugf("hubble relay server status failed: %v", err)
 		}
 		select {
 		case <-ctx.Done():
 			return fmt.Errorf("hubble server status failure: %w", ctx.Err())
 		case <-time.After(time.Second):
+			a.Debugf("retrying hubble relay server status request")
```
viktor-kurchenko, line 748: NIT: `a.Debugf` -> `a.Debug`.

connectivity/check/action.go
```diff
@@ -721,18 +721,31 @@ func (a *Action) GetIngressRequirements(p FlowParameters) []filters.FlowSetRequi
...
 	for {
-		res, err := client.ServerStatus(ctx, &observer.ServerStatusRequest{})
-		if err == nil && (res.NumUnavailableNodes == nil || res.NumUnavailableNodes.Value == 0) {
-			// This means all the nodes are available.
-			return nil
+		// Sometimes ServerStatus request just gets stuck until the context gets cancelled. Call
+		// ServerStatus with a shorter timeout than ctx so that we get to try it multiple times.
+		timeoutContext, cancel := context.WithTimeout(ctx, 3*time.Second)
+		res, err := client.ServerStatus(timeoutContext, &observer.ServerStatusRequest{})
+		cancel()
```
viktor-kurchenko, line 731: NIT:
This method looks a bit tricky, so I just wonder what if we move the status check into a new private method:
```
func (a *Action) isServerStatusOK(ctx context.Context, client observer.ObserverClient) bool {
	// Sometimes ServerStatus request just gets stuck until the context gets cancelled. Call
	// ServerStatus with a shorter timeout than ctx so that we get to try it multiple times.
	timeoutContext, cancel := context.WithTimeout(ctx, 3*time.Second)
	defer cancel()
	res, err := client.ServerStatus(timeoutContext, &observer.ServerStatusRequest{})
	if err != nil {
		a.Debugf("hubble relay server status failed: %v", err)
		return false
	}
	a.Debugf("hubble relay server status %+v", res)
	// This means all the nodes are available, and Hubble Relay started receiving flows.
	// Ideally we should check res.NumConnectedNodes matches the expected number of Cilium
	// nodes instead of relying on res.NumUnavailableNodes, but I don't know if that information
	// is available to us.
	return (res.NumUnavailableNodes == nil || res.NumUnavailableNodes.Value == 0) && res.NumFlows > 0
}
```

and then the current method will look like:
```
func (a *Action) waitForRelay(ctx context.Context, client observer.ObserverClient) error {
	for {
		if a.isServerStatusOK(ctx, client) {
			return nil
		}
		select {
		case <-ctx.Done():
			return fmt.Errorf("hubble server status failure: %w", ctx.Err())
		case <-time.After(time.Second):
			a.Debugf("retrying hubble relay server status request")
		}
	}
}
```

Just a suggestion, the current implementation looks good.

viktor-kurchenko (approved):

gandro (approved):
Relay changes look good to me conceptually. I assume we're seeing the linked issues partially because port-forward is unreliable on AKS?

connectivity/check/action.go
```diff
@@ -721,18 +721,31 @@ func (a *Action) GetIngressRequirements(p FlowParameters) []filters.FlowSetRequi
 }
 
 // waitForRelay polls the server status from Relay until either it's connected to all the Hubble
-// instances (success) or the context is cancelled (failure).
+// instances, and it started receiving flows (success) or the context is cancelled (failure).
```
kaworu, line 724: FWIW The flow count returned by `ServerStatus` is [an aggregation of all the Hubble server answers](https://github.com/cilium/cilium/blob/a1c949c488b23bd46922c9974ccd4eec6475b814/pkg/hubble/relay/observer/server.go#L298-L310), technically no flow is going through Hubble Relay until a `GetFlows` request is made.

kaworu (approved):
Relay check changes LGTM, I like @viktor-kurchenko suggestion about splitting into two functions.

connectivity/check/action.go
michi-covalent: hmm wonder how i should phrase this 🤔

michi-covalent: changed it to:

```
// waitForRelay polls the server status from Relay until it indicates that there are no unavailable
// nodes and num_flows is greater than zero (success), or the context is cancelled (failure).
```

connectivity/check/action.go
michi-covalent: updated 👍

.github/workflows/aks-byocni.yaml
michi-covalent: https://github.com/cilium/cilium-cli/pull/1578/commits/fecd9d467c05c2c34953357a3aa510f9af751a7b will be removed before merging.

asauber (approved):

michi-covalent:
removed the test commit. time to 🚀

PR #1579, opened 2023-05-05
connectivity: Automatically install IP routes on nodes w/o Cilium for from CIDR tests AND get rid of --datapath

brb:
This PR introduces a few changes:

1. Moves IP routes (pod CIDR => node on nodes w/o Cilium) installation to the CLI.
2. Removes `--datapath`, and introduces `--include-unsafe-tests`. The formal removal required a few changes to the flow validation and skipping some tests to make it pass on some GHA workflows.

Please review it per commit.

Successful run - https://github.com/cilium/cilium/pull/25510.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -308,8 +308,12 @@ func (ct *ConnectivityTest) SetupAndValidate(ctx context.Context) error {
 		}
 	}
 	if ct.features.MatchRequirements(RequireFeatureEnabled(FeatureNodeWithoutCilium)) {
-		if err := ct.validateExternalFromCIDRsWithNodesWithoutCilium(); err != nil {
-			return fmt.Errorf("invalid configuration for nodes without Cilium: %w", err)
+		if err := ct.detectPodCIDRs(ctx); err != nil {
+			return fmt.Errorf("unable to detect pod CIDRs: %w", err)
+		}
+
+		if err := ct.detectNodesWithoutCiliumIPs(); err != nil {
+			return fmt.Errorf("unable to detect nodes w/o Cilium IPs: %w", err)
 		}
 	}
 	return nil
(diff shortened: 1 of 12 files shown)


nathanjsweet (approved):

connectivity/check/deployment.go
```diff
@@ -266,7 +266,7 @@ func newDaemonSet(p daemonSetParameters) *appsv1.DaemonSet {
 							ReadinessProbe:  p.ReadinessProbe,
 							SecurityContext: &corev1.SecurityContext{
 								Capabilities: &corev1.Capabilities{
-									Add: []corev1.Capability{"NET_RAW"},
+									Add: []corev1.Capability{"NET_ADMIN", "NET_RAW"},
```
joestringer, line 269: Is this proposing to grant `NET_ADMIN` rights to connectivity test deployments in users' production environments? That seems like a sort of security footprint we are explicitly trying to avoid in a "safe" tool like this. :thinking:

joestringer, line 269: Is this proposing to grant `NET_ADMIN` rights to connectivity test deployments in users' production environments? That seems like a sort of security footprint we are explicitly trying to avoid in a "safe" tool like this. :thinking:

brb: Good concern, and I overlooked that this granting applies to all deployments (not only to those running in host netns). Anyway, `NET_ADMIN` is required only for host netns pods which runs on special testing nodes which do not run Cilium (`--nodes-without-cilium`) for running tests from outside a cluster (e.g., to test N/S loadbalancing). I'm going to change the granting in this PR.

connectivity/check/context.go
```diff
@@ -528,17 +533,62 @@ func (ct *ConnectivityTest) enableHubbleClient(ctx context.Context) error {
...
-		ct.params.ExternalFromCIDRMasks = append(ct.params.ExternalFromCIDRMasks, addr.BitLen())
 	}
+
+	return nil
+}
+
+func (ct *ConnectivityTest) detectNodesWithoutCiliumIPs() error {
+	for _, n := range ct.nodesWithoutCilium {
+		pod := ct.hostNetNSPodsByNode[n]
+		for _, ip := range pod.Pod.Status.PodIPs {
```
viktor-kurchenko, line 562: What if `ct.hostNetNSPodsByNode` doesn't contain `n` key? Probably there will be panic in runtime (coz `Pod` is pointer), isn't it?

brb: Yep, it will panic. However, it's not expected that such an entry does not exist, so the panic would happen only if there is a bug in the code.

brb:
Successful CI run - https://github.com/cilium/cilium/pull/25638. The Go linter failure is a red-herring.

connectivity/check/deployment.go
joestringer: LGTM thanks.

christarazi (approved):
LGTM, thanks!

Last commit could be squashed with the first?

viktor-kurchenko (approved):

brb:
> Last commit could be squashed with the first?

I've tried squashing it, but hit many merge conflicts. Anyway, I think having two separate commits gives a good understanding to a reader about our preference for granting permissions in a fine-grained manner.

PR #1590, opened 2023-05-08
Add junit output to connectivity test

brlbil:
Junit format varies from implementation to implementation. The below references were used as references.

https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
https://llg.cubic.org/docs/junit/
`k8s e2e junit output`

`--junit-file` option is added to the `connectivity test`
test output is converted to junit xml format, and written to the destination file


michi-covalent (commented):
we should probably update our github workflows to write junit output right? are you planning to do it as a followup pull request?

brlbil:
> we should probably update our github workflows to write junit output right? are you planning to do it as a followup pull request?

I actually thought about adding a test but since `cilium-cli` tests run in the cluster I thought the output would not be very useful. Now I think we should add one even regardless of output is useful to test the functionality.

I think adding one of the tests is enough, what do you think @michi-covalent?

michi-covalent:
> I think adding one of the tests is enough

sounds good to me

brlbil:
I missed the kind workflow, even better we can easily upload the connectivity test output with the kind workflow.
I will do this first thing tomorrow.

connectivity/check/context.go
```diff
@@ -366,6 +372,75 @@ func (ct *ConnectivityTest) skip(t *Test) {
...
+	f, err := os.Create(ct.Params().JunitFile)
+	if err != nil {
+		return err
+	}
+
+	if err := suites.WriteReport(f); err != nil {
+		return err
+	}
+
+	return f.Close()
```
viktor-kurchenko, line 441: File won't be closed if `WriteReport` returns an error.
What do you think about closing the file in the `defer` block:
```
	f, err := os.Create(ct.Params().JunitFile)
	if err != nil {
		return err
	}
	defer func() {
		if err := f.Close(); err != nil {
			//log error here
		}
	}()

	return suites.WriteReport(f)
```

connectivity/check/test.go
```diff
@@ -245,6 +248,8 @@ func (t *Test) Run(ctx context.Context) error {
 		s.Run(ctx, t)
 	}
 
+	t.completionTime = time.Now()
```
viktor-kurchenko, line 251: The `completionTime` field won't be populated in case of an error in the loop.
How about moving this line into the `defer` block above?

internal/junit/junit_test.go
```diff
@@ -0,0 +1,58 @@
...
+				{Name: "test3", Classname: "test", Status: "failed", Time: 7,
+					Failure: &Failure{Message: "test3 failed", Type: "failure", Value: "failure > expected"}},
+			},
+		},
+	},
+}
+
+func TestWriteReport(t *testing.T) {
+	buf := &bytes.Buffer{}
+	if err := suites.WriteReport(buf); err != nil {
```
viktor-kurchenko, line 51: Why don't you want to use `asserts`?
For example:
```
func TestWriteReport(t *testing.T) {
	buf := &bytes.Buffer{}
	assert.NoError(t, suites.WriteReport(buf))
	assert.Equal(t, suitesText, buf.String())
}
```

brlbil: Great point, habits 😄 I usually do not use test helper libraries, and I did not know it was used in this codebase.

connectivity/check/test.go
brlbil: Good catch, we need to set it before returning an error, there is not in setting up the test as well

viktor-kurchenko (approved):
LGTM.

michi-covalent (approved):
xml ftw

michi-covalent:
the button is too green not to click

PR #1595, opened 2023-05-08
Remove az-binary check

michi-covalent:
I assert that az-binary check is a bit too excessive since it's not always required to have az binary installed when you are installing Cilium on AKS.

Instead of forcing users to specify `--disable-check=az-binary` flag, let's just fail if az binary is needed but not found. This behavior is consistent with how we handle gcloud binary.

### DIFF
--- a/.github/in-cluster-test-scripts/aks-byocni-install.sh
+++ b/.github/in-cluster-test-scripts/aks-byocni-install.sh
@@ -6,7 +6,6 @@ set -e
 # Install Cilium
 cilium install \
   --version "${CILIUM_VERSION}" \
-  --disable-check=az-binary \
   --datapath-mode=aks-byocni \
   --wait=false \
   --helm-set loadBalancer.l7.backend=envoy \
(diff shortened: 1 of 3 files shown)


asauber (approved):
What does the output look like when the binary is missing?

michi-covalent:
> What does the output look like when the binary is missing?

```
❌ Unable to execute "az account show --output json --only-show-errors":
```

asauber:
@michi-covalent does the output below that indicate that the binary is missing? I think some messaging along those lines is useful.

michi-covalent:
yes it does, sorry i didn't paste the whole output:

```
❌ Unable to execute "az account show --output json --only-show-errors":
↩️ Rolling back installation...

Error: Unable to install Cilium: unable to execute "az account show --output json --only-show-errors": exec: "az": executable file not found in $PATH
```

tommyp1ckles (approved):

tommyp1ckles:
makes sense to me

sayboras (approved):

nebril (approved):

PR #1606, opened 2023-05-08
Add a function to check if the CLI is in Helm mode

michi-covalent:
There are multiple places where we check if the CLI is in Helm mode. Add a function utils.IsInHelmMode() so that this logic is in a single place.

### DIFF
--- a/internal/utils/utils_test.go
+++ b/internal/utils/utils_test.go
@@ -5,10 +5,12 @@ package utils
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"testing"
 
 	"github.com/blang/semver/v4"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestCheckVersion(t *testing.T) {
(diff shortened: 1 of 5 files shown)


nathanjsweet (approved):

internal/utils/utils_test.go
```diff
@@ -203,3 +205,14 @@ func TestBuildImagePath(t *testing.T) {
...
+
+func TestIsInHelmMode(t *testing.T) {
+	orig := os.Getenv(CLIModeVariableName)
+	assert.NoError(t, os.Setenv(CLIModeVariableName, "helm"))
+	assert.True(t, IsInHelmMode())
+	assert.NoError(t, os.Setenv(CLIModeVariableName, "classic"))
+	assert.False(t, IsInHelmMode())
+	assert.NoError(t, os.Setenv(CLIModeVariableName, "random"))
+	assert.False(t, IsInHelmMode())
+	assert.NoError(t, os.Setenv(CLIModeVariableName, orig))
```
tklauser, line 217: Let's move the setting of the original value to a `defer` func such that is reset also in case of an error or panic:

```suggestion
	orig := os.Getenv(CLIModeVariableName)
	defer func() {
		assert.NoError(t, os.Setenv(CLIModeVariableName, orig))
	}()
	assert.NoError(t, os.Setenv(CLIModeVariableName, "helm"))
	assert.True(t, IsInHelmMode())
	assert.NoError(t, os.Setenv(CLIModeVariableName, "classic"))
	assert.False(t, IsInHelmMode())
	assert.NoError(t, os.Setenv(CLIModeVariableName, "random"))
	assert.False(t, IsInHelmMode())
```

michi-covalent: https://github.com/cilium/cilium-cli/pull/1612 😎

PR #1612, opened 2023-05-10
unit test: Restore the original env variable using defer

michi-covalent:
Fixes: 840b570e4515 ("Add a function to check if the CLI is in Helm mode")


tklauser (approved):
Thanks :sunglasses:

Issue #1620, opened 2023-05-12
Implement `cilium clustermesh connect` using new Helm functions


asauber:
The `cilium clustermesh connect` subcommand is the final subcommand in the Cilium CLI which templates and creates Kubernetes resources outside the context of Helm.

It has a significant amount of autoconfiguration logic, such as determining the public IP addresses of the `clustermesh-apiserver` for each cluster, which we want to maintain.

However, there are aspects of this autoconfiguration which need significant improvement, for example, it currently hardcodes hostnames of the form `<clustername>.mesh.cilium.io` into the etcd configuration for each cluster, even when these names would not be resolvable. This very recently led to an [open proposal related to this limitation](https://github.com/cilium/cilium-cli/issues/1539).

As described in that Issue, it is now possible to use the Helm chart to configure a variety of values that were previously not exposed by this subcommand.

The goal of this task is to make use of the values and configuration logic that now exist in the form of Helm template functions in the upstream chart in order to implement this subcommand.

One of the goals of this implementation is that it should support automatic configuration of a shared Certificate Authority among all clusters using both a `cert-manager` Issuer-based implementation and and implementation which does not use `cert-manager` (e.g. Helm-generated or Certgen-based).

giorio94:
> It currently hardcodes hostnames of the form `<clustername>.mesh.cilium.io` into the etcd configuration for each cluster, even when these names would not be resolvable. This very recently led to a user to open https://github.com/cilium/cilium-cli/issues/1539.

One additional point here is that host aliases are configured for the `<clustername>.mesh.cilium.io` names (since they would otherwise not resolve to valid address) in the cilium agent daemonset. This works fine when the clustermesh-apiserver is exposed through a NodePort/LoadBalancer IP address, while it incurs in possible shortcomings when the LoadBalancer is associated with a DNS name. AFAIK, the current `cilium clustermesh connect` implementation tries to resolve that DNS name and then configures the resulting IP address(es) as host alias(es). Yet, the DNS name might not be resolvable from the client machine (as mentioned in the issue), as well as this would lead to issues if the underlying IP addresses change.

One additional complexity of directly using the DNS name, though, is that the server TLS certificate needs to match it. The `connect` command will likely need to also configure proper [`server.entraDnsNames` entries](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/values.yaml#L2601). This cannot be performed while enabling clustermesh, since at that point the hostname might not yet be known.

Issue #1539, opened 2023-04-26
Proposal / RFE: Change to use the value of `external-dns.alpha.kubernetes.io/hostname` instead of `${ClusterName}.mesh.cilium.io` when `external-dns.alpha.kubernetes.io/hostname` is set on the `clustermesh-apiserver` LoadBalancer Service


kahirokunn:
## Proposal / RFE

In the current implementation, it is necessary to perform a DNS lookup for all target clusters in the environment running `cilium clustermesh connect`. This can be difficult depending on network configurations.

As a security best practice, the `clustermesh-apiserver` NLB should be created in a private subnet. However, the default hostname's A record for an NLB in a private subnet is registered in the `Route53 Private Hosted Zone`. As a result, lookup from a local PC will fail.

The primary solution is to create a custom A record for the `clustermesh-apiserver` NLB using tools like `External DNS`, and register it in a Public Hosted Zone.

![image](https://user-images.githubusercontent.com/22343391/234726650-b304ba37-7e3f-48ee-960f-0b43f99a5476.png)

However, the primary reason for performing the DNS lookup is to set the IP for `hostAliases` of `${ClusterName}.mesh.cilium.io`. The initial motivation for wanting to perform a DNS lookup during the `cilium clustermesh connect` process is when the DNS hostname of the `clustermesh-apiserver` LoadBalancer Service is unknown, right?

## Implementation Pattern 1

As a solution, how about adding the following flags?

* `--use-status-hostname`
  * Referencing `status.loadBalancer.ingress[].hostname`
* `--use-destination-status-hostname`
  * Referencing `status.loadBalancer.ingress[].hostname`
* `--use-externaldns-hostname`
  * Referencing `metadata.annotations."external-dns.alpha.kubernetes.io/hostname"`
* `--use-destination-externaldns-hostname`
  * Referencing `metadata.annotations."external-dns.alpha.kubernetes.io/hostname"`

In this case, there will be no need for `hostAliases` and they will not be set.
There is also a rollOutCiliumPods option in the helm values, so we think that those who want to restart the system should activate it.

## Implementation Pattern 2

1. If `external-dns.alpha.kubernetes.io/hostname` is set on `clustermesh-apiserver`, use that hostname as the target `clustermesh-apiserver` for the connection. Do not set hostAliases.
https://github.com/cilium/cilium-cli/blob/9c3ea4c5e55ee7b6d466c57ade34091e0314b13a/clustermesh/clustermesh.go#L648

[...]

PR #1628, opened 2023-05-13
helm mode: add implementation for clustermesh connect

asauber:
Add an implementation for `clustermesh connect` using Helm.

Like the classic mode implementation, we autodetect the `clustermesh-apiserver` IPs and port numbers in order to write the `endpoints:` configuration. However, rather than generate the certificates in-process, we rely on the `cronJob` (certgen) mode of the Helm chart to generate (and re-generate) all of the ClusterMesh related certs.

As a first pass, we support the same set of flags as the classic mode.

A second pass PR will include all `--helm-*` flags, which will allow the use of `extraDnsNames` for the purpose of Service resolvability, and the use of other PKI modes.

Fixes: #1620

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -6,8 +6,10 @@ package clustermesh
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -23,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
@@ -30,6 +33,7 @@ import (
 	"github.com/cilium/cilium/api/v1/models"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	"github.com/cilium/cilium/pkg/versioncheck"
+	"helm.sh/helm/v3/pkg/release"
 
 	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/internal/certs"
@@ -953,36 +957,56 @@ func (k *K8sClusterMesh) patchConfig(ctx context.Context, client k8sClusterMeshI
 	return nil
 }
 
-func (k *K8sClusterMesh) Connect(ctx context.Context) error {
-	remoteCluster, err := k8s.NewClient(k.params.DestinationContext, "")
+// getClientsForConnect returns a k8s.Client for the local and remote cluster, respectively
+func (k *K8sClusterMesh) getClientsForConnect() (*k8s.Client, *k8s.Client, error) {
+	remoteClient, err := k8s.NewClient(k.params.DestinationContext, "")
 	if err != nil {
-		return fmt.Errorf("unable to create Kubernetes client to access remote cluster %q: %w", k.params.DestinationContext, err)
+		return nil, nil, fmt.Errorf(
+			"unable to create Kubernetes client to access remote cluster %q: %w",
+			k.params.DestinationContext, err)
 	}
+	return k.client.(*k8s.Client), remoteClient, nil
+}
 
-	aiRemote, err := k.extractAccessInformation(ctx, remoteCluster, k.params.DestinationEndpoints, true, false)
+// connectAccessInit initializes a Kubernetes client for the local and remote cluster
+// and performs some validation that the two clusters can be connected via clustermesh
+func (k *K8sClusterMesh) getAccessInfoForConnect(
+	ctx context.Context, localClient, remoteClient *k8s.Client,
+) (*accessInformation, *accessInformation, error) {
+	aiRemote, err := k.extractAccessInformation(ctx, remoteClient, k.params.DestinationEndpoints, true, false)
 	if err != nil {
-		k.Log("❌ Unable to retrieve access information of remote cluster %q: %s", remoteCluster.ClusterName(), err)
-		return err
+		k.Log("❌ Unable to retrieve access information of remote cluster %q: %s", remoteClient.ClusterName(), err)
+		return nil, nil, err
 	}
 
-	if !aiRemote.validate() {
-		return fmt.Errorf("remote cluster has non-unique name (%s) and/or ID (%s)", aiRemote.ClusterName, aiRemote.ClusterID)
-	}
-
-	aiLocal, err := k.extractAccessInformation(ctx, k.client, k.params.SourceEndpoints, true, false)
+	aiLocal, err := k.extractAccessInformation(ctx, localClient, k.params.SourceEndpoints, true, false)
 	if err != nil {
 		k.Log("❌ Unable to retrieve access information of local cluster %q: %s", k.client.ClusterName(), err)
-		return err
+		return nil, nil, err
+	}
+
+	return aiLocal, aiRemote, nil
+}
+
+// connectAccessInit initializes a Kubernetes client for the local and remote cluster
+// and performs some validation that the two clusters can be connected via clustermesh
+func (k *K8sClusterMesh) validateInfoForConnect(aiLocal, aiRemote *accessInformation) error {
+	if !aiRemote.validate() {
+		return fmt.Errorf("remote cluster has non-unique name (%s) and/or ID (%s)",
+			aiRemote.ClusterName, aiRemote.ClusterID)
 	}
 
 	if !aiLocal.validate() {
-		return fmt.Errorf("local cluster has the default name (cluster name: %s) and/or ID 0 (cluster ID: %s)",
+		return fmt.Errorf(
+			"local cluster has the default name (cluster name: %s) and/or ID 0 (cluster ID: %s)",
 			aiLocal.ClusterName, aiLocal.ClusterID)
 	}
 
 	cid, err := strconv.Atoi(aiRemote.ClusterID)
 	if err != nil {
-		return fmt.Errorf("remote cluster has non-numeric cluster ID %s. Only numeric values 1-255 are allowed", aiRemote.ClusterID)
+		return fmt.Errorf(
+			"remote cluster has non-numeric cluster ID %s. Only numeric values 1-255 are allowed",
+			aiRemote.ClusterID)
 	}
 	if cid < 1 || cid > 255 {
 		return fmt.Errorf("remote cluster has cluster ID %d out of acceptable range (1-255)", cid)
@@ -996,17 +1020,36 @@ func (k *K8sClusterMesh) Connect(ctx context.Context) error {
 		return fmt.Errorf("remote and local cluster have the same, non-unique ID: %s", aiLocal.ClusterID)
 	}
 
-	k.Log("✨ Connecting cluster %s -> %s...", k.client.ClusterName(), remoteCluster.ClusterName())
+	return nil
+}
+
+func (k *K8sClusterMesh) Connect(ctx context.Context) error {
+	localClient, remoteClient, err := k.getClientsForConnect()
+	if err != nil {
+		return err
+	}
+
+	aiLocal, aiRemote, err := k.getAccessInfoForConnect(ctx, localClient, remoteClient)
+	if err != nil {
+		return err
+	}
+
+	err = k.validateInfoForConnect(aiLocal, aiRemote)
+	if err != nil {
+		return err
+	}
+
+	k.Log("✨ Connecting cluster %s -> %s...", k.client.ClusterName(), remoteClient.ClusterName())
 	if err := k.patchConfig(ctx, k.client, aiRemote); err != nil {
 		return err
 	}
 
-	k.Log("✨ Connecting cluster %s -> %s...", remoteCluster.ClusterName(), k.client.ClusterName())
-	if err := k.patchConfig(ctx, remoteCluster, aiLocal); err != nil {
+	k.Log("✨ Connecting cluster %s -> %s...", remoteClient.ClusterName(), k.client.ClusterName())
+	if err := k.patchConfig(ctx, remoteClient, aiLocal); err != nil {
 		return err
 	}
 
-	k.Log("✅ Connected cluster %s and %s!", k.client.ClusterName(), remoteCluster.ClusterName())
+	k.Log("✅ Connected cluster %s and %s!", localClient.ClusterName(), remoteClient.ClusterName())
 
 	return nil
 }
@@ -1757,7 +1800,7 @@ func (k *K8sClusterMesh) ExternalWorkloadStatus(ctx context.Context, names []str
 		}
 		cews = cewList.Items
 		if len(cews) == 0 {
-			k.Log("⚠️  No external workloads found.")
+			k.Log("⚠️ No external workloads found.")
 			return nil
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


asauber:
Converting to draft to fix `cronJob` config.

internal/cli/cmd/clustermesh.go
```diff
@@ -393,3 +394,29 @@ func newCmdClusterMeshDisableWithHelm() *cobra.Command {
...
+			if err := cm.ConnectWithHelm(context.Background()); err != nil {
+				fatalf("Unable to connect cluster: %s", err)
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&params.DestinationContext, "destination-context", "", "Kubernetes configuration context of destination cluster")
+	cmd.Flags().StringSliceVar(&params.DestinationEndpoints, "destination-endpoint", []string{}, "IP of ClusterMesh service of destination cluster")
+	cmd.Flags().StringSliceVar(&params.SourceEndpoints, "source-endpoint", []string{}, "IP of ClusterMesh service of source cluster")
```
michi-covalent, line 419: let's do the common flags thingy we've been doing with other commands so that it's easier to see which flags are common.

clustermesh/clustermesh.go
```diff
@@ -1810,3 +1810,131 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+	_, err = helm.Upgrade(ctx, remoteCluster.RESTClientGetter, upgradeParams)
+	if err != nil {
+		return err
+	}
+
+	k.Log("✅ Connected cluster %s and %s!", k.client.ClusterName(), remoteCluster.ClusterName())
+	return nil
+}
+
+func genClusterMeshConfig(aiLocal, aiRemote *accessInformation) map[string]interface{} {
```
michi-covalent, line 1910: this function looks very unit-testable 👀

clustermesh/clustermesh.go
```diff
@@ -1810,3 +1810,131 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
 	_, err = helm.Upgrade(ctx, k8sClient.RESTClientGetter, upgradeParams)
 	return err
 }
+
+// ConnectWithHelm enables clustermesh using a Helm Upgrade action Certificates are generated via
+// the Helm chart's cronJob (certgen) mode As with classic mode, only autodetected IP-based
+// clustermesh-apiserver Service endpoints are currently supported.
+func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
+	v, err := k.client.GetRunningCiliumVersion(ctx, k.params.Namespace)
```
michi-covalent, line 1818: let's use version from helm chart instead of calling this function. it is a very shady function: https://github.com/cilium/cilium-cli/blob/main/k8s/client.go#L928-L944

clustermesh/clustermesh.go
```diff
@@ -1901,3 +1886,35 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+					},
+				},
+			},
+			"config": map[string]interface{}{
+				"enabled": true,
+				"clusters": []map[string]interface{}{
+					{
+						"name": aiLocal.ClusterName,
+						"ips":  []string{aiLocal.ServiceIPs[0]},
+						"port": aiLocal.ServicePort,
```
giorio94, line 1929: Given that the TLS config is not specified, this does work only when the CA is shared. I would suggest to add a check at the beginning to prevent connecting two clusters when this condition does not hold, as it otherwise leads to a broken but difficult to troubleshoot setup.

clustermesh/clustermesh.go
```diff
@@ -1811,7 +1811,27 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+	v, err := k.client.GetRunningCiliumVersion(ctx, k.params.Namespace)
+	if err != nil {
+		return err
+	}
+	ciliumVer, err := utils.ParseCiliumVersion(v)
+	if err != nil {
+		k.Log("⚠️ Failed to parse Cilium version %s: %w. "+
+			"1.14+ is assumed. Continuing with Helm-mode", v, err)
+	} else {
+		if ciliumVer.LT(versioncheck.MustVersion("1.14.0")) {
```
giorio94, line 1827: Snapshot versions seem not to be included by this check (I'm not sure if that is intentional).

giorio94:
I've tested this PR trying to connect more than two clusters, and it seems to overwrite the previous configuration rather than adding the new cluster (as the legacy implementation did).

clustermesh/clustermesh.go
```diff
@@ -1854,45 +1854,30 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
 		Name:        defaults.HelmReleaseName,
-		Values:      helmValues,
+		Values:      helmValuesLocal,
 		ResetValues: false,
 		ReuseValues: true,
 	}
 
-	// TODO (ajs): After classic mode removal, use a k8s.Client for k.client
+	// TODO (ajs): After classic mode removal, use a k8s.Client for this k.client
 	_, err = helm.Upgrade(ctx, k.client.(*k8s.Client).RESTClientGetter, upgradeParams)
```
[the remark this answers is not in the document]
asauber: The delay is likely from the cronJob execution. I agree that more friendly output would be nice here.

clustermesh/clustermesh.go
asauber: The tricky thing here is treating snapshots of < 1.14 and 1.14 appropriately. We have a function that attempts to exec into a pod to grab the version, which I was using here previously, but it seemed like overkill.

I’ll think about this more.

asauber:
Great suggestions all, working on it

clustermesh/clustermesh.go
michi-covalent: try `>=1.14.0-dev`.

it should work with `1.14.0-snapshot.*` versions as well as the main branch chart https://github.com/cilium/cilium/blob/main/VERSION

clustermesh/clustermesh.go
```diff
@@ -1810,3 +1810,131 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
 	_, err = helm.Upgrade(ctx, k8sClient.RESTClientGetter, upgradeParams)
 	return err
 }
+
+// ConnectWithHelm enables clustermesh using a Helm Upgrade action Certificates are generated via
```
squeed, line 1814: nit: It seems like this docbloc had more punctuation or newlines that disappeared.

clustermesh/clustermesh.go
```diff
@@ -1810,3 +1810,131 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+
+	// TODO (ajs): Support hostname-based endpoints via --helm-values override
+	//   Using the Helm value `extraDnsNames` should work as-is.
+	//   This can be useful for LoadBalancer reachability.
+
+	// Enable clustermesh using a Helm Upgrade command
+	upgradeParams := helm.UpgradeParameters{
+		Namespace:   k.params.Namespace,
+		Name:        defaults.HelmReleaseName,
+		Values:      helmValuesLocal,
```
squeed, line 1889: Unfortunately, the helm value coalescer is not smart enough to merge arrays; they're always fully replaced.

This means that this will **disconnect** any existing connections, since the `clustermesh.config.clusters` array will be fully overwritten.

You'll have to retrieve the in-use Helm values and mutate them manually, rather than relying on the [automagic helm value coalescing](https://github.com/cilium/cilium-cli/blob/main/vendor/helm.sh/helm/v3/pkg/chartutil/coalesce.go#L36).

clustermesh/clustermesh.go
asauber: unit tests for `updateClustermeshConfig` and `GetPath` will be added in a follow-up PR

asauber:
In order for `Multicluster / installation-and-connectivity (helm) (pull_request_target)` to pass, this PR requires a rebase after this is merged: https://github.com/cilium/cilium-cli/pull/1685

asauber:
@giorio94 connecting 3+ clusters is now supported

asauber:
As part of the CI rebase, I will rebase this into atomic commits.

internal/utils/maps.go
```diff
@@ -0,0 +1,42 @@
...
+		out[k] = v
+	}
+	return out
+}
+
+// GetPath gets a value by a path in a nested map[string]interface{}.
+// An error is returned if:
+//   - At any point along the path, the map is not a map[string]interface{}
+//   - At any point along the path, the path segment does not appear in the map
+func GetPath(m map[string]interface{}, path []string) (interface{}, error) {
```
squeed, line 29: Check out https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured, it might be useful

squeed (approved):
Can you squash to one or two commits? Then, looks good!

internal/utils/maps.go
asauber: Implemented :+1:

clustermesh/clustermesh.go
```diff
@@ -1885,3 +1941,199 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+		return err
+	}
+
+	k.Log("✅ Connected cluster %s and %s!", localClient.ClusterName(), remoteClient.ClusterName())
+	return nil
+}
+
+func updateClustermeshConfig(
+	values map[string]interface{},
+	aiLocal, aiRemote *accessInformation,
```
giorio94, line 2082: Given that `updateClustermeshConfig` needs to be called twice (once for the local cluster and once for the remote cluster), you could drop one of the `accessInformation` parameters, and pass only the appropriate one.

clustermesh/clustermesh.go
```diff
@@ -1885,3 +1941,199 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+	// merge in other required values
+	baseValues := map[string]interface{}{
+		"clustermesh": map[string]interface{}{
+			"config": map[string]interface{}{
+				"enabled":  true,
+				"clusters": outputClusters,
+			},
+		},
+	}
+	newValues := helm.MergeMaps(values, baseValues)
```
giorio94, line 2136: Given that `ReuseValues` is set, it shouldn't be necessary to manually merge back the full original values here.

giorio94 (changes requested):
Overall, the PR looks OK to me, I've left a few minor comments inline. Additionally, there seems to be something wrong with the merging logic, since the `cilium clustermesh connect` command fails with `Error: Unable to connect cluster: existing clustermesh.config.cluters array is invalid` when issued a second time.

clustermesh/clustermesh.go
```diff
@@ -1885,3 +1941,199 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+	err = k.validateInfoForConnect(aiLocal, aiRemote)
+	if err != nil {
+		return err
+	}
+
+	// Validate that CA certificates match between the two clusters
+	err = k.validateCAMatch(aiLocal, aiRemote)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
asauber: In the default case (using `cronJob` mode), we would have the `caCert` for the `tls` config but not the `cert` or `key`. Does this config now support passing only the `caCert` for each cluster?

clustermesh/clustermesh.go
asauber: `updateClustermeshConfig` is being called once against the Helm values of the local cluster, and once against the Helm values of the remote cluster, each time it needs to ensure that both the information from `aiLocal` and `aiRemote` is contained in those Helm values.

clustermesh/clustermesh.go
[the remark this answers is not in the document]
giorio94: I think that the `cert` and `key` info are already retrieved through `extractAccessInformation` (unless there is a secret name mismatch, which is 100% possible). That key/cert pair in the secret will be overwritten the subsequent time the certgen job is run (most likely when helm is executed to connect the two clusters), but the old ones will still work. I agree that this is not the best solution, but it might help to smooth the user experience when the clusters are configured with a different CA (which I think is fairly common especially in test setups). WDYT?

clustermesh/clustermesh.go
giorio94: In the config, you need only to specify the parameters about the remote clusters, while the ones for the local cluster are not necessary (and they are indeed ignored if present).

asauber: I see, that's surprising to me, and appears to be under-documented.

clustermesh/clustermesh.go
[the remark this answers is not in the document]
asauber: One of the reasons for defaulting to `crontJob` mode is so that certificate renewals will happen automatically. Setting `tls:` at connect time would diminish that benefit.

This does raise an interesting concern which is that some users may wish to set `address:` at connect time, and the current implementation would overwrite that silently.

I need to think about this more.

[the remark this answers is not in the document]
giorio94: > One of the reasons for defaulting to crontJob mode is so that certificate renewals will happen automatically. Setting tls: at connect time would diminish that benefit.

I 100% agree. If we decide to go with this, there should be a warning that it is a compatibility mode with many limitations (including at lease hubble issues and lack of certs renewal). 

> This does raise an interesting concern which is that some users may wish to set address: at connect time, and the current implementation would overwrite that silently.

I'm not sure I got what you mean here.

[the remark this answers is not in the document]
asauber: > I'm not sure I got what you mean here.

We know that some users are interested in this connect implementation using hostnames (possibly not resolving `ips:` at all, see https://github.com/cilium/cilium-cli/issues/1539). The current implementation of the config merging in this PR, when it constructs the config entry for the remote cluster, does not include any `address:` even if there was an attempt to provide it with `--helm-set` or `--helm-file`.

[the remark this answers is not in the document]
giorio94: Gotcha. Yeah, I didn't try that possibility since I saw in the comments that `address` was not yet supported.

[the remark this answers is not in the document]
asauber: Implemented your initial suggestion of warning the user on CA mismatch and configuring `tls:` https://github.com/cilium/cilium-cli/pull/1628/commits/82eedfceb01fdf0a8a14dc221ca50a17bfd41ff9

clustermesh/clustermesh.go
```diff
@@ -1885,3 +1939,224 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+}
+
+// validateCAMatch determines if the certificate authority certificate being
+// used by aiLocal and aiRemote uses a matching keypair. The bool return value
+// is one of: nil (unknown), true (keypairs match), false (keys do not match)
+func (k *K8sClusterMesh) validateCAMatch(aiLocal, aiRemote *accessInformation) (*bool, error) {
+	caLocal := aiLocal.CA
+	block, _ := pem.Decode(caLocal)
+	if block == nil {
+		panic("failed to parse certificate PEM for local CA cert")
```
giorio94, line 1959: Nit. It is likely better to return an error rather than panic here. Same below.

clustermesh/clustermesh.go
```diff
@@ -1885,3 +1939,224 @@ func DisableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Paramete
...
+	if err != nil {
+		return nil, err
+	}
+	return release, nil
+}
+
+// validateCAMatch determines if the certificate authority certificate being
+// used by aiLocal and aiRemote uses a matching keypair. The bool return value
+// is one of: nil (unknown), true (keypairs match), false (keys do not match)
+func (k *K8sClusterMesh) validateCAMatch(aiLocal, aiRemote *accessInformation) (*bool, error) {
```
giorio94, line 1955: Nit. I would suggest to return a plain bool rather than a pointer (for the sake of simplicity). It currently seems to be nil only when an error is not nil.

giorio94 (approved):
Looks awesome to me! Just a couple of very minor nits inline. I guess that the last missing piece is now the `clustermesh disconnect` support.

clustermesh/clustermesh.go
asauber: Ah, of course, I meant to replace these

clustermesh/clustermesh.go
asauber: I'll make this change, but it feels wrong to me to not return "maybe" somehow. I agree that this is a simple enough function that it's unlikely to cause any issues.

internal/helm/helm.go
```diff
@@ -96,7 +96,8 @@ func filterManifests(manifest string) map[string]string {
 	return manifestsToRender
 }
 
-func mergeMaps(a, b map[string]interface{}) map[string]interface{} {
+// Merge maps recursively merges the values of b into a copy of a, preferring the values from b
+func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
```
tklauser, line 100: Does this still need to be exported? I don't see any call size outside of `internal/helm`.

asauber: It doesn't, reverting it now

Issue #1625, opened 2023-05-12
Migrate all remaining cilium-cli workflows to run using Helm mode


asauber:
As part of the Cilium CLI Helm mode implementation, a variety of the CI workflows were updated to run both a Helm-based implementation and "classic" implementation.

https://github.com/cilium/cilium-cli/blob/main/.github/workflows/gke.yaml#L41
https://github.com/cilium/cilium-cli/blob/main/.github/workflows/multicluster.yaml#L42

One of these was the first CI usage of the `cilium upgrade` subcommand.

https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml#L226

The goal of this task is to migrate all of the remaining workflows to use the Helm mode, in preparation for Helm mode becoming the default mode in the next release of `cilium-cli`

michi-covalent:
opened draft PRs:

- eks: https://github.com/cilium/cilium-cli/pull/1638
- eks tunnel mode: https://github.com/cilium/cilium-cli/pull/1639
- external workloads: https://github.com/cilium/cilium-cli/pull/1640

@asauber i didn't touch https://github.com/cilium/cilium-cli/blob/main/.github/workflows/multicluster.yaml. i'm assuming you'll take care of it after https://github.com/cilium/cilium-cli/pull/1628 gets merged ✨

asauber:
@michi-covalent multicluster was completed a while back https://github.com/cilium/cilium-cli/commit/41d8616942dde523026a38dd396b018d6d000cf5 :grin:

michi-covalent:
oh wow things are moving at an amazing pace

PR #1640, opened 2023-05-16
helm mode: Add external workloads test coverage

michi-covalent:
Ref: #1625

### DIFF
--- a/clustermesh/certs.go
+++ b/clustermesh/certs.go
@@ -137,6 +137,7 @@ func (k *K8sClusterMesh) createClusterMeshClientCertificate(ctx context.Context)
 }
 
 func (k *K8sClusterMesh) createClusterMeshExternalWorkloadCertificate(ctx context.Context) error {
+	certName := getExternalWorkloadCertName()
 	certReq := &csr.CertificateRequest{
 		Names:      []csr.Name{{C: "US", ST: "San Francisco", L: "CA"}},
 		KeyRequest: csr.NewKeyRequest(),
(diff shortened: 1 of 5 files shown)


michi-covalent:
it's failing in helm mode because `clustermesh-apiserver-external-workload-cert` doesn't exist:

https://github.com/cilium/cilium-cli/blob/64bd65cbf4cc27253a8aec2525aaafa3c551c44b/clustermesh/certs.go#L139

need to look into whether this is configurable via helm 👀

michi-covalent:
✅ https://github.com/cilium/cilium-cli/actions/runs/5009483637/jobs/8978707657?pr=1640

removing the temporary commit and marking it ready for review

christarazi (approved):

clustermesh/certs.go
```diff
@@ -137,6 +137,7 @@ func (k *K8sClusterMesh) createClusterMeshClientCertificate(ctx context.Context)
 }
 
 func (k *K8sClusterMesh) createClusterMeshExternalWorkloadCertificate(ctx context.Context) error {
+	certName := getExternalWorkloadCertName()
```
giorio94, line 140: Nit. The changes in this file should not be required, since these functions are used only in the classic mode.

clustermesh/clustermesh.go
```diff
@@ -1776,6 +1783,7 @@ func EnableWithHelm(ctx context.Context, k8sClient *k8s.Client, params Parameter
 	helmStrValues := []string{
 		"clustermesh.useAPIServer=true",
 		fmt.Sprintf("clustermesh.apiserver.service.type=%s", params.ServiceType),
+		"externalWorkloads.enabled=true",
```
giorio94, line 1786: I would suggest not to enable this unconditionally (maybe add a flag?), given that in most cases external workloads support is not needed, and it would defeat the optimization introduced in https://github.com/cilium/cilium/pull/25259.

michi-covalent: let's do a flag 🎏

clustermesh/certs.go
michi-covalent: yeah it's not required, but i'd like to use `getExternalWorkloadCertName()` function throughout the code base consistently. when people see `defaults.ClusterMeshExternalWorkloadSecretName` being used like this, people might get the wrong impression that this is the correct secret name regardless of which mode the CLI is in.

clustermesh/clustermesh.go
michi-covalent: added a flag ✅

michi-covalent:
✅ successful run: https://github.com/cilium/cilium-cli/actions/runs/5018586743?pr=1640 removing the temporary commit.

internal/cli/cmd/clustermesh.go
```diff
@@ -368,6 +368,7 @@ func newCmdClusterMeshEnableWithHelm() *cobra.Command {
 	}
 
 	cmd.Flags().StringVar(&params.ServiceType, "service-type", "NodePort", "Type of Kubernetes service to expose control plane { LoadBalancer | NodePort | ClusterIP }")
+	cmd.Flags().BoolVar(&params.EnableExternalWorkloads, "enable-external-workloads", false, "Set externalWorkloads.enabled Helm value to true")
```
giorio94, line 371:
```suggestion
	cmd.Flags().BoolVar(&params.EnableExternalWorkloads, "enable-external-workloads", false, "Enable support for external workloads, such as VMs")
```

Nit.

giorio94 (approved):
/lgtm

internal/cli/cmd/clustermesh.go
michi-covalent: 👍 i'll make the change before merging ✨

asauber (approved):
Thanks for the updates

PR #1651, opened 2023-05-23
check: make (Service)Address() return an IP address

jibi:
change the Address method to return an IP address when available (i.e. when ClusterIP is not empty, which happens in case of headless services) instead of just returning the service name.

towards: https://github.com/cilium/cilium/issues/24151

### DIFF
--- a/connectivity/check/peer.go
+++ b/connectivity/check/peer.go
@@ -10,6 +10,7 @@ import (
 
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	"github.com/cilium/cilium-cli/k8s"
 )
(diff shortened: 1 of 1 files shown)


jrajahalme (approved):
LGTM

jibi:
Always returning an IP (beside the `clusterIP == ""` case) causes the `PodToService()` and `PodToHostPort()` scenarios to fail, specifically the flow validation:

```
2023-05-23T13:17:40.1564008Z   ℹ️  SYN and(ip(src=10.244.0.252,dst=10.245.249.71),tcp(dstPort=8080),tcpflags(syn)) not found
2023-05-23T13:17:40.1569614Z   ❌ Aborting flow matching: context deadline exceeded
2023-05-23T13:17:40.1570235Z   ❌ Flow validation failed for peer test-namespace/client-64c655df64-mb6pd: 2 failures (first: -1, last: -1, matched: 0)
```

as it's not clear yet what's causing that, and I'm not able to reproduce this locally, for the meantime I switched the first condition to:

```go
	if s.Service.Spec.ClusterIP == "" || family == IPFamilyAny {
		return s.Service.Name
	}
```

so that we keep the current behavior of returning the service name when the IP family is set to any

PR #1673, opened 2023-05-30
mutual-auth: Update the terminology

sayboras:
The main change is to update vendor, also some updates are done on
related CNP:

- Adjust the CR based on latest spec (e.g. authentication.mode instead
  of auth.type)
- Remove hard-code namespace so that the tests can run in user provided
  namespace via CLI
- Rename the policy name to match file name (like what we have with
  other policies)
- Remove the podToEndpoints scenarios as this scenario is mainly for
  testing L7 related feature, for mutual auth, L3/L4 is good enough

Relates: https://github.com/cilium/cilium/issues/24867


rolinh (approved):

michi-covalent (approved):

sayboras:
As https://github.com/cilium/cilium/pull/25904 is merged, remove preview only label after update the commit hash from cilium/cilium main branch.

PR #1683, opened 2023-05-31
connectivity: Add upgrade tests

brb:
Please see each individual commit msg when reviewing.

Testing in https://github.com/cilium/cilium/pull/25790 (red builds can be ignored, as they are due to Cilium issues which the upgrade tests have caught).

Related https://github.com/cilium/cilium/issues/25037.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/blang/semver/v4"
@@ -381,6 +382,26 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 
 	if err := ct.writeJunit(); err != nil {
 		ct.Failf("writing to junit file %s failed: %s", ct.Params().JunitFile, err)
+	}
+
+	if ct.Params().FlushCT {
+		var wg sync.WaitGroup
+
+		wg.Add(len(ct.CiliumPods()))
+		for _, ciliumPod := range ct.CiliumPods() {
+			cmd := strings.Split("cilium bpf ct flush global", " ")
+			go func(ctx context.Context, pod Pod) {
+				defer wg.Done()
+
+				ct.Debugf("Flushing CT entries in %s/%s", pod.Pod.Namespace, pod.Pod.Name)
+				_, err := pod.K8sClient.ExecInPod(ctx, pod.Pod.Namespace, pod.Pod.Name, defaults.AgentContainerName, cmd)
+				if err != nil {
+					ct.Fatal("failed to flush ct entries: %w", err)
+				}
+			}(ctx, ciliumPod)
+		}
+
+		wg.Wait()
 	}
 
 	// Report the test results.
--- a/connectivity/tests/upgrade.go
+++ b/connectivity/tests/upgrade.go
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package tests
+
+import (
+	"context"
+	gojson "encoding/json"
+	"os"
+	"strconv"
+
+	"github.com/cilium/cilium-cli/connectivity/check"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NoInterruptedConnections checks whether there are no interruptions in
+// long-lived E/W LB connections. The test case is used to validate Cilium
+// upgrades.
+//
+// The test case consists of three steps:
+//
+// 1. Deploying pods and a service which establish the long-lived connections
+// (done by "--upgrade-test-setup"). The client pods ("migrate-svc-client")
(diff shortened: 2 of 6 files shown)


connectivity/tests/upgrade.go
```diff
@@ -0,0 +1,101 @@
...
+			t.Fatalf("Failed to create %q file for writing upgrade test temp results: %s",
+				ct.Params().UpgradeTestResultPath, err)
+		}
+		defer file.Close()
+
+		for pod, count := range restartCount {
+			if _, err := file.WriteString(fmt.Sprintf("%s=%s,", pod, count)); err != nil {
+				t.Fatalf("Failed to write upgrade test temp result into file: %s", err)
+			}
+		}
```
aanm, line 75: Let's simply do a json.Marshall / json.Unmarshall :smile:

connectivity/tests/upgrade.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Fatalf("Failed to read upgrade test result files: %s", err)
+	}
+	prevRestartCount := make(map[string]string)
+	for _, pair := range strings.Split(string(b), ",") {
+		if pair == "" {
+			continue
+		}
+		res := strings.Split(pair, "=")
+		prevRestartCount[res[0]] = res[1]
+	}
```
aanm, line 91: Let's simply do a json.Marshall / json.Unmarshall :smile:

aanm (changes requested):
Can we also have instructions on which order we should run the commands?

brb:
> Can we also have instructions on which order we should run the commands?

It's in the last commit msg :sunglasses:

aanm (approved):

connectivity/check/context.go
```diff
@@ -383,6 +383,16 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 		ct.Failf("writing to junit file %s failed: %s", ct.Params().JunitFile, err)
 	}
 
+	if ct.Params().FlushCT {
+		for _, ciliumPod := range ct.CiliumPods() {
```
asauber, line 387: This can be done concurrently, similar to https://github.com/cilium/cilium-cli/blob/main/bgp/peers.go#L67-L78

connectivity/tests/upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+	prevRestartCount := make(map[string]string)
+	if err := gojson.Unmarshal(b, &prevRestartCount); err != nil {
+		t.Fatalf("Failed to unmarshal JSON test result file: %s", err)
+	}
+
+	for pod, count := range restartCount {
+		if prevCount, found := prevRestartCount[pod]; !found {
+			t.Fatalf("Could not found Pod %s restart count", pod)
+		} else if prevCount != count {
```
asauber, line 98: What is it about Pod restarts here that implies that a connection was reset? Is there a one-shot connection, which if it fails, fails a `livenessProve` for the Pod?

A link to the manifest or documentation about this would be helpful here.

asauber (approved):
Looks good for the most part. Commented with some suggested improvements.

connectivity/tests/upgrade.go
brb: Yep, a Pod restart means that a long-lived connection was interrupted. I put some docs in the beginning of this file. Please let me know if more details are needed.

For now I reused the same migrate-svc deployments from the ginkgo test suite. In the future, we should at least rename them, document and extend their use-cases (e.g., to test N/S traffic interruptions).

connectivity/check/context.go
brb: Aye, good suggestion. Currently, our CI K8s clusters consist of ~4 nodes, so the parallelism was never a concern. However, we have plans to run those tests on way larger clusters.

internal/cli/cmd/connectivity.go
```diff
@@ -172,6 +172,11 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 
 	initSysdumpFlags(cmd, &params.SysdumpOptions, "sysdump-", hooks)
 
+	cmd.Flags().BoolVar(&params.IncludeUpgradeTest, "include-upgrade-test", false, "Include upgrade test")
+	cmd.Flags().BoolVar(&params.UpgradeTestSetup, "upgrade-test-setup", false, "Set up upgrade test dependencies")
+	cmd.Flags().StringVar(&params.UpgradeTestResultPath, "upgrade-test-result-path", "/tmp/cilium-upgrade-test-restart-counts", "Upgrade test temporary result file (used internally)")
+	cmd.Flags().BoolVar(&params.FlushCT, "flush-ct", false, "Flush conntrack of Cilium on each node")
```
pchaigno, line 178: Should we hide this flag given connectivity tests are supposed to be usable by folks in their production cluster without side-effects? The blast radius of a misuse here is quite high.

joestringer: Oh I see yes, this known issue: https://github.com/cilium/cilium/issues/17459 :pray:

PR #1687, opened 2023-06-01
connectivity: Ability to add custom annotations to test deployment

gandro:
connectivity: Ability to add custom annotations to test deployment

This commit adds the ability to add custom namespace and pod annotations
to the `cilium connectivity test` deployment. The two flags are
`--deployment-pod-annotations` and `--namespace-annotations`.

For the pod annotations, we accept a JSON map which contains the
deployment name as the key, and the annotations as a string-to-string
map. For the namespace annotation, we simply expect a string-to-string
map.

We could have used Viper's `StringToString` map for the namespace
annotations, but not for the pod annotations, since there we would need
a "`StringToStringToString`" map. Therefore, to remain consistent
between the two flags, both flags exclusively JSON syntax.

If the deployment name in the annotations map ends with a `*` symbol,
then we perform a longest prefix match.

The flags are currently hidden, since we are not fully commited to this
command-line syntax yet and it might still change in the future.

Example:

```console
$ cilium connectivity test \
    --namespace-annotations='{"foo":"bar"}' \
    --deployment-pod-annotations='{"client":{"baz":"qux"},"echo-*":{"quux":"corge"}}'
```


tklauser (approved):
Nice, thanks! Maybe provide a working example command using these flags in the commit message and/or PR description?

gandro:
Some of the CI failures look legitimate: `connectivity test failed: unable to create namespace cilium-test: namespaces "cilium-test" already exists`. Not sure how that could happen. Looking into it!

squeed:
Looks good. One question: is there any need for setting annotations on all pods? Something like `--deployment-pod-annotations='{"*":{"foo":"bar"}}'`

squeed (approved):

gandro:
> Looks good. One question: is there any need for setting annotations on all pods? Something like `--deployment-pod-annotations='{"*":{"foo":"bar"}}'`

Oh that's a nice suggestion. Yeah I think supporting a wildcard match would be nice.

PR #1706, opened 2023-06-07
while applying policies, wait for all expected policy revisions.

tommyp1ckles:
Previously, we always waited for the r+1 revision, which in the case of multiple policies being applied may result in a race condition where some, but not all, of the policies are in place prior to proceeding with tests.


tommyp1ckles:
@cilium/sig-policy @nathanjsweet Can you confirm that every new/deleted policy will result in a separate revision (i.e. will not ever be batched up into a single revision)?

nathanjsweet (approved):

PR #1708, opened 2023-06-08
install: Add node affinity for Operator

sayboras:
This to make sure that Cilium Operator will not be scheduled to the node having label io.cilium/no-schedule:true


michi-covalent (approved):
reasonable

PR #1734, opened 2023-06-19
clustermesh enable: add enable-kvstoremesh flag

giorio94:
This PR extends the `cilium clustermesh enable` command with a new flag which can be used to enable kvstoremesh (cilium/cilium#26083). The new flag is supported only when using the new Helm mode, and requires cilium >= 1.14 (it has no effect on earlier versions). Related: https://github.com/cilium/cilium/pull/26348

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -479,6 +479,10 @@ type Parameters struct {
 	// EnableExternalWorkloads indicates whether externalWorkloads.enabled Helm value
 	// should be set to true. For Helm mode only.
 	EnableExternalWorkloads bool
+
+	// EnableKVStoreMesh indicates whether kvstoremesh should be enabled.
+	// For Helm mode only.
+	EnableKVStoreMesh bool
 }
 
 func (p Parameters) validateParams() error {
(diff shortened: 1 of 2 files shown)


asauber (approved):
Approving this as a clean implementation that matches the current style.

It brings up two questions for me:

1. Maybe we should reconsider allowing `--helm-set` for subcommands like `clustermesh enable`, because we would get features like this for free.

2. The chained type-assertion on the Helm values instead of using types is starting to look a bit messy, we should revisit that.

giorio94:
>     1. Maybe we should reconsider allowing `--helm-set` for subcommands like `clustermesh enable`, because we would get features like this for free.

I personally agree. While for certain features it might still better to have a dedicated flag, we also need `--helm-set` to be able to configure image versions or other parameters.

>     2. The chained type-assertion on the Helm values instead of using types is starting to look a bit messy, we should revisit that.

That's true. Maybe we could add an helper to set a given value at a certain "path" in the values map.

michi-covalent (approved):
😇

YutaroHayakawa (approved):

PR #1736, opened 2023-06-20
connectivity: Add checks for agent error logs and drops due to missed tail calls

brb:
See commit msgs.

### DIFF
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -217,6 +217,8 @@ func Run(ctx context.Context, ct *check.ConnectivityTest, addExtraTests func(*ch
 			// include --include-upgrade-tests"
 			return ct.Run(ctx)
 		}
+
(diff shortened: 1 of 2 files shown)


michi-covalent (approved):
😇

connectivity/suite.go
```diff
@@ -217,6 +217,8 @@ func Run(ctx context.Context, ct *check.ConnectivityTest, addExtraTests func(*ch
 			// include --include-upgrade-tests"
 			return ct.Run(ctx)
 		}
+
+		ct.NewTest("no-missed-tail-calls").WithScenarios(tests.NoMissedTailCalls())
```
pchaigno, line 221: Probably out of scope, but I wonder if we could generalize that to more than just the upgrade and more than just missed tail calls drops. There are a lot of drops that we never expect to see. In a sense, they are the same as level=error log messages.

brb: Yep, good idea. Maybe let's start an issue to collect all thoughts?

PR #1739, opened 2023-06-20
Add Helm Clustermesh disconnect command

derailed:
- Surface new command to match cluster disconnect when CLI helm mode in enabled

Signed-off-by: Fernand Galiana <fernand.galiana@isovalent.com>

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -968,6 +968,26 @@ func (k *K8sClusterMesh) getClientsForConnect() (*k8s.Client, *k8s.Client, error
 	return k.client.(*k8s.Client), remoteClient, nil
 }
 
+func (k *K8sClusterMesh) shallowExtractAccessInfo(ctx context.Context, c *k8s.Client) (*accessInformation, error) {
+	cm, err := c.GetConfigMap(ctx, k.params.Namespace, defaults.ConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve ConfigMap %q: %w", defaults.ConfigMapName, err)
+	}
+	id, ok := cm.Data[configNameClusterID]
+	if !ok {
+		return nil, fmt.Errorf("unable to locate key: %q in ConfigMap %q", configNameClusterID, defaults.ConfigMapName)
+	}
+	name, ok := cm.Data[configNameClusterName]
+	if !ok {
+		return nil, fmt.Errorf("unable to locate key: %q in ConfigMap %q", configNameClusterName, defaults.ConfigMapName)
+	}
+
+	return &accessInformation{
+		ClusterID:   id,
+		ClusterName: name,
+	}, nil
+}
+
 // connectAccessInit initializes a Kubernetes client for the local and remote cluster
 // and performs some validation that the two clusters can be connected via clustermesh
 func (k *K8sClusterMesh) getAccessInfoForConnect(
@@ -2082,6 +2102,95 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
 	return nil
 }
 
+func (k *K8sClusterMesh) DisconnectWithHelm(ctx context.Context) error {
+	localRelease, err := getRelease(k.client.(*k8s.Client), k.params.Namespace)
+	if err != nil {
+		k.Log("❌ Unable to find Helm release for the target cluster")
+		return err
+	}
+	version := localRelease.Chart.AppVersion()
+	semv, err := utils.ParseCiliumVersion(version)
+	if err != nil {
+		return fmt.Errorf("failed to parse Cilium version: %w", err)
+	}
+	k.Log("✅ Detected Helm release with Cilium version %s", semv)
+
+	const minCiliumHelmRev = "1.14.0"
+	cv, err := semver.Parse(minCiliumHelmRev)
+	if err != nil {
+		return fmt.Errorf("failed to parse Cilium version: %w", err)
+	}
+	if cv.Compare(semv) < 0 {
+		// Helm-based clustermesh enable is only supported on Cilium
+		// v1.14+ due to a lack of support in earlier versions for
+		// autoconfigured certificates (tls.{crt,key}) for cluster
+		// members when running in certgen (cronJob) PKI mode
+		k.Log("⚠️ Cilium Version is less than 1.14.0. Continuing in classic mode.")
+		return k.Disconnect(ctx)
+	}
+
+	localClient, remoteClient, err := k.getClientsForConnect()
+	if err != nil {
+		return err
+	}
+	aiLocal, err := k.shallowExtractAccessInfo(ctx, localClient)
+	if err != nil {
+		return err
+	}
+	aiRemote, err := k.shallowExtractAccessInfo(ctx, remoteClient)
+	if err != nil {
+		return err
+	}
+
+	if err = k.validateInfoForConnect(aiLocal, aiRemote); err != nil {
+		return err
+	}
+
+	// Modify the clustermesh config to remove the intended cluster if any
+	localHelmValues, err := removeFromClustermeshConfig(localRelease.Config, aiRemote.ClusterName)
+	if err != nil {
+		return err
+	}
+
+	// Get existing helm values for the remote cluster
+	remoteRelease, err := getRelease(remoteClient, k.params.Namespace)
+	if err != nil {
+		k.Log("❌ Unable to find Helm release for the remote cluster")
+		return err
+	}
+	// Modify the clustermesh config to remove the intended cluster if any
+	remoteHelmValues, err := removeFromClustermeshConfig(remoteRelease.Config, aiLocal.ClusterName)
+	if err != nil {
+		return err
+	}
+
+	upgradeParams := helm.UpgradeParameters{
+		Namespace:   k.params.Namespace,
+		Name:        defaults.HelmReleaseName,
+		Values:      localHelmValues,
+		ResetValues: false,
+		ReuseValues: true,
+	}
+
+	// Disconnect clustermesh using a Helm Upgrade command against our target cluster
+	k.Log("ℹ️ Configuring Cilium in cluster '%s' to disconnect from cluster '%s'",
+		localClient.ClusterName(), remoteClient.ClusterName())
+	if _, err = helm.Upgrade(ctx, localClient.RESTClientGetter, upgradeParams); err != nil {
+		return err
+	}
+
+	// Disconnect clustermesh using a Helm Upgrade command against the remote cluster
+	k.Log("ℹ️ Configuring Cilium in cluster '%s' to disconnect from cluster '%s'",
+		remoteClient.ClusterName(), localClient.ClusterName())
+	upgradeParams.Values = remoteHelmValues
+	if _, err = helm.Upgrade(ctx, remoteClient.RESTClientGetter, upgradeParams); err != nil {
+		return err
+	}
+	k.Log("✅ Disconnected clusters %s and %s!", localClient.ClusterName(), remoteClient.ClusterName())
+
+	return nil
+}
+
 func updateClustermeshConfig(
 	values map[string]interface{}, aiRemote *accessInformation, configTLS bool,
 ) (map[string]interface{}, error) {
@@ -2090,7 +2199,7 @@ func updateClustermeshConfig(
 	if err != nil {
 		return nil, fmt.Errorf("existing clustermesh.config is invalid")
 	}
-	if !found {
+	if !found || c == nil {
 		c = []interface{}{}
 	}
 
@@ -2150,6 +2259,45 @@ func updateClustermeshConfig(
 	newValues := map[string]interface{}{
 		"clustermesh": map[string]interface{}{
 			"config": map[string]interface{}{
+				"enabled":  true,
+				"clusters": outputClusters,
+			},
+		},
+	}
+
+	return newValues, nil
+}
+
+func removeFromClustermeshConfig(values map[string]any, clusterName string) (map[string]any, error) {
+	// get current clusters config slice, if it exists
+	c, found, err := unstructured.NestedFieldCopy(values, "clustermesh", "config", "clusters")
+	if err != nil {
+		return nil, fmt.Errorf("existing clustermesh.config is invalid")
+	}
+	if !found || c == nil {
+		c = []any{}
+	}
+
+	cs, ok := c.([]any)
+	if !ok {
+		return nil, fmt.Errorf("existing clustermesh.config.clusters array is invalid")
+	}
+	outputClusters := make([]map[string]any, 0, len(cs))
+	for _, m := range cs {
+		cluster, ok := m.(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("existing clustermesh.config.clusters map is invalid")
+		}
+		name, ok := cluster["name"].(string)
+		if ok && name == clusterName {
+			continue
+		}
+		outputClusters = append(outputClusters, cluster)
+	}
+
+	newValues := map[string]any{
+		"clustermesh": map[string]any{
+			"config": map[string]any{
 				"enabled":  true,
 				"clusters": outputClusters,
 			},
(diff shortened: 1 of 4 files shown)


clustermesh/clustermesh.go
```diff
@@ -2082,6 +2082,97 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse Cilium version: %w", err)
+	}
+	if cv.Compare(semv) < 0 {
+		// Helm-based clustermesh enable is only supported on Cilium
+		// v1.14+ due to a lack of support in earlier versions for
+		// autoconfigured certificates (tls.{crt,key}) for cluster
+		// members when running in certgen (cronJob) PKI mode
+		k.Log("⚠️ Cilium Version is less than 1.14.0. Continuing in classic mode.")
+		return k.Connect(ctx)
```
asauber, line 2109: This should be `k.Disconnect`

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2082,97 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+	}
+
+	// Get existing helm values for the remote cluster
+	remoteRelease, err := getRelease(remoteClient, k.params.Namespace)
+	if err != nil {
+		k.Log("❌ Unable to find Helm release for the remote cluster")
+		return err
+	}
+	remoteHelmValues := remoteRelease.Config
+	// Expand those values to include the clustermesh configuration
```
asauber, line 2141: This comment doesn't match what's being executed here.

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2082,97 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
 	return nil
 }
 
+func (k *K8sClusterMesh) DisconnectWithHelm(ctx context.Context) error {
```
asauber, line 2085: Refactoring the duplication here with `ConnectWithHelm` might be in order, as it would have prevented the `Connect` bug below.

asauber (changes requested):

clustermesh/clustermesh.go
derailed: Not sure I get that? We are still updating the clustermesh config section...

clustermesh/clustermesh.go
derailed: This code being so sequential is a bit hard to refact. Did a quick pass. Let's see if we're happier.

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2073,107 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
 	return nil
 }
 
+// Helm-based clustermesh enable is only supported on Cilium
+// v1.14+ due to a lack of support in earlier versions for
+// autoconfigured certificates (tls.{crt,key}) for cluster
+// members when running in certgen (cronJob) PKI mode
+func (k *K8sClusterMesh) inClassicMode(r *release.Release) (bool, error) {
```
asauber, line 2080: nit: This would be better named `needsClassicMode` rather than `inClassicMode`, since we are definitely in Helm mode if we call this function.

asauber (changes requested):
The current build failure indicates that the new semver logic is treating the Helm version under test (a 1.13.x version) not as classic mode.

Note that runs of the test "Kind / helm-upgrade-clustermesh (pull_request)" show the log message "⚠️ Cilium Version is less than 1.14.0. Continuing in classic mode."

clustermesh/clustermesh.go
derailed: Thanks Andrew! I thought `ClassicMode` meant: "run the standard Clustermesh connect/disconnect as if helm upgrade mode is disabled".  If that's the case and the cilium rev is < 1.14.0 then we are `in classic mode'.
Am I wrong about that?

derailed: Also do you have an idea on what's up with the tests failures? 
When I've first checked in this am, all tests were green and now 2 suites are failing consistently relating to a hosed tls secret key? Do you have a sense on why they started failing?

asauber:
Yes, the failure is caused by the modifications here to the version
checking logic. semver.Compare was specifically avoided previously because
it did not handle prefixes and suffixes present in the version strings that
are used by Cilium for pre-release versions.

I would advise debugging the code using the values present in the CI
context, or reverting to the previous implementation.

asauber:
Correct, this is a semantic nitpick from me. This boolean predicate
function is returning whether or not we “needClassicMode”, the value of the
environment variable which causes us to reach this function means we must
already be in Helm mode. In other words, a Helm version is not
“isClassicMode”, but it might “needClassicMode”. It’s really a nitpick.

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2073,107 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to parse Cilium version: %w", err)
+	}
+	k.Log("✅ Detected Helm release with Cilium version %s", semv)
+
+	const ciliumHelmMinRev = "1.14.0"
+	cv, err := semver.Parse(ciliumHelmMinRev)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse Cilium version: %w", err)
+	}
```
giorio94, line 2092: Could use `semver.MustParse()` given that the input is a constant.

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2073,107 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+	if cv.Compare(semv) < 0 {
+		k.Log("⚠️ Cilium Version is less than 1.14.0. Continuing in classic mode.")
+		return true, nil
+	}
```
giorio94, line 2096: This will not work for pre-releases. I guess that it should be either restored the previous approach, or the version changed to something like 1.13.99 (don't know if there's a better way though). The comparison seems also to be inverted (it is checking if `1.14.0 < semver`). I'd suggest to use `LT/GT` rather than `Compare` for increased clarity.

clustermesh/clustermesh.go
```diff
@@ -2158,3 +2250,59 @@ func updateClustermeshConfig(
...
 	return newValues, nil
 }
+
+func removeFromClustermeshConfig(values map[string]interface{}, aiRemote *accessInformation) (map[string]any, error) {
+	// get current clusters config slice, if it exists
+	c, found, err := unstructured.NestedFieldCopy(values, "clustermesh", "config", "clusters")
+	if err != nil {
+		return nil, fmt.Errorf("existing clustermesh.config is invalid")
+	}
+	if found && c == nil {
```
giorio94, line 2260:
```suggestion
	if !found || c == nil {
```

I guess (dropping the `if !found` below).

clustermesh/clustermesh.go
```diff
@@ -2158,3 +2250,59 @@ func updateClustermeshConfig(
...
+		}
+	}
+	if !victim {
+		return nil, fmt.Errorf("active cluster is not currently connected to cluster %q", aiRemote.ClusterName)
+	}
+
+	outputClusters := make([]map[string]any, 0, len(clusters))
+	for _, v := range clusters {
+		outputClusters = append(outputClusters, v)
+	}
```
giorio94, line 2297: This logic could be likely simplified to avoid creating multiple intermediate maps. Given that no subsequent modification needs to be performed, a single pass which copies all existing elements (but the victim) to a new slice should be enough.

internal/cli/cmd/clustermesh.go
```diff
@@ -418,6 +418,28 @@ func newCmdClusterMeshConnectWithHelm() *cobra.Command {
...
+		Run: func(cmd *cobra.Command, args []string) {
+			params.Namespace = namespace
+			cm := clustermesh.NewK8sClusterMesh(k8sClient, params)
+			if err := cm.DisconnectWithHelm(context.Background()); err != nil {
+				fatalf("Unable to disconnect clusters: %s", err)
+			}
+		},
+	}
+
+	addCommonConnectFlags(cmd, &params)
```
giorio94, line 438: Should be changed to the only flag which is registered in the disconnect case.

giorio94 (changes requested):
The current semver check does not seem to be correct, which is the cause for the tests failure. I'd also reverse the order of the commits, to reduce the churn.

clustermesh/clustermesh.go
derailed: I think this is correct. Field copy could return found=true with a nil copy.

clustermesh/clustermesh.go
derailed: Good point! Thank you Marco!

internal/cli/cmd/clustermesh.go
derailed: Not sure I follow... But I think these are used by both connect and disconnect via getAccessForConnect

clustermesh/clustermesh.go
derailed: Good point! I'll take another pass. No keen on the whole string compare deal.

clustermesh/clustermesh.go
giorio94: Yeah, I mean that regardless of whether it is found but the value is nil, or it is not found, I guess that you could return an error here.

internal/cli/cmd/clustermesh.go
giorio94: Oh, that's right. The weird thing though is that they don't make much sense for disconnect as they are actually never used (for instance the remote endpoint).

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2073,108 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	if err = k.validateInfoForConnect(aiLocal, aiRemote); err != nil {
+		return err
+	}
+
+	// Get existing helm values for the local cluster
+	localHelmValues := localRelease.Config
+	// Expand those values to include the clustermesh configuration
```
asauber, line 2130: Another nitpick: what I was pointing out with this comment (and the similar one below), was that in this case we already expect to have a clustermesh config, so we are not "expanding" it but rather: `// Modify the clustermesh config to remove the intended cluster`

asauber (approved):
Looking good, thanks for the additional tests and updates.

clustermesh/clustermesh.go
derailed: Ok - I'll update. Tx for the review Andrew!

internal/cli/cmd/clustermesh.go
derailed: OK - Good point indeed! Tx for the clarification! 
Took another swing at it. Let's see if we're happier...

clustermesh/clustermesh.go
```diff
@@ -2082,6 +2102,95 @@ func (k *K8sClusterMesh) ConnectWithHelm(ctx context.Context) error {
...
+
+	upgradeParams := helm.UpgradeParameters{
+		Namespace:   k.params.Namespace,
+		Name:        defaults.HelmReleaseName,
+		Values:      localHelmValues,
+		ResetValues: false,
+		ReuseValues: true,
+	}
+
+	// Enable clustermesh using a Helm Upgrade command against our target cluster
```
giorio94, line 2178: Nit. this comment should be updated (same for the one below).

clustermesh/clustermesh.go
```diff
@@ -2158,3 +2267,64 @@ func updateClustermeshConfig(
...
+		if name, ok := c["name"].(string); ok {
+			if name == aiRemote.ClusterName {
+				victim = true
+				continue
+			}
+			clusters[name] = c
+		}
+	}
+	if !victim {
+		return nil, fmt.Errorf("active cluster is not currently connected to cluster %q", clusterName)
```
giorio94, line 2305: Should we return an error in case the target cluster is not present, or just ignore it? Let's say that we executed `cilium clustermesh disconnect` and something went wrong while applying the helm changes to the second cluster. At that point we would have a broken setup in which it is impossible to execute disconnect again. 

Another related aspect is whether connect and disconnect changes should be atomic, in the sense that if a failure occurs while applying them to the second cluster we should trigger a helm rollback on the first one. I honestly don't have a strong opinion here. WDYT? (/cc @michi-covalent @asauber)

clustermesh/clustermesh_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestRemoveFromClustermeshConfig(t *testing.T) {
+	uu := map[string]struct {
+		vv      map[string]any
+		cluster string
+		err     error
+		e       map[string]any
+	}{
+		"missing": {
+			cluster: "test1",
+			err:     errors.New(`active cluster is not currently connected to cluster "test1"`),
```
giorio94, line 96: Why is this error different from the empty case below?

clustermesh/clustermesh.go
```diff
@@ -2285,7 +2288,7 @@ func removeFromClustermeshConfig(values map[string]interface{}, clusterName stri
 	if !ok {
 		return nil, fmt.Errorf("existing clustermesh.config.clusters array is invalid")
 	}
-	oldClusters := make([]map[string]any, 0, len(cs))
+	outputClusters, victim := make([]map[string]any, 0, len(cs)), false
```
giorio94: It seems to me that these changes (and the ones below) should be part of the other commit.

giorio94 (commented):
The overall changes look good to me. I've left only a couple of minor cleanup comments and one question inline. I'd finally suggest to move the commit about 'needsClassicMode` first to reduce the churn and ensure that they all result in a working version.

clustermesh/clustermesh.go
derailed: Interesting Q Marco! I do think the connect/disconnect operations should be transactional and not leave the clusters in a dangling state given the symmetrical nature of these calls. 

That said this error does not trigger state change on the clusters as the actual helm update takes place in subsequent stage. But it does leave users with the inability to perform an operation if the validation fails.

giorio94: > That said this error does not trigger state change on the clusters as the actual helm update takes place in subsequent stage. But it does leave users with the inability to perform an operation if the validation fails.

Yeah, it would cause a failure only if the previous attempt failed in the middle of the upgrade. That's something that could be addressed in a followup PR though (probably just moving to a transactional approach).

clustermesh/clustermesh_test.go
giorio94: Actually both in the `missing` and the `empty` cases the config is empty.

derailed: No sure what you are getting at? Not quite, one use case is nil the other empty. This is testing the returns of the copy.

clustermesh/clustermesh.go
michi-covalent: yeah that's probably out of scope for this PR. let me open a github issue to follow up on this 📝

clustermesh/clustermesh_test.go
giorio94: Never mind, it was just a nitpick. My point is that since in both cases the `clusters` map is empty (whether it is nil or not is only an implementation detail IMHO) I would expect the same error.

clustermesh/clustermesh.go
michi-covalent: filed https://github.com/cilium/cilium-cli/issues/1748 ✅ 😇

clustermesh/clustermesh_test.go
derailed: Got it! Tx Marco. No quite an implementation detail but a potential bug ie copy can return found=true and the copy=nil, hence the differentiation in the test and the fix in the original impl. Does this make sense?

giorio94: Yeah, the fix makes perfectly sense to me. I would just expect in both cases to immediately return with the same error, but it's really just a nitpick (related: https://github.com/cilium/cilium-cli/pull/1739#discussion_r1236503178)

clustermesh/clustermesh.go
derailed: Decided to axe the validation and simply just gen the config. This should alleviate the issue you'd pointed out where one cluster is left dangling.

giorio94 (approved):
Looks good to me, thanks!

PR #1741, opened 2023-06-21
.github/workflows: don't error out if pkill finds no processes

tklauser:
The pkill invocation in the kind workflow is meant to kill background tasks started earlier in the workflow. It seems like some of them exit of their own accord, with tests still passing.

Ignore the pkill error code 1 which indicates that no processes were matched / signalled:

       1      No processes matched or none of them could be signalled.

Ref. https://github.com/cilium/cilium/issues/26075


michi-covalent (approved):
😇

PR #1742, opened 2023-06-21
Minor fixes for error messages on failure to disable Hubble

qmonnet:
Here are a few minor fixes for some error messages from the CLI, in particular on uninstall when trying to disable Hubble.

- internal/cli/cmd: Add line breaks after error messages
- internal/cli/cmd: Fix error displayed on failure to disable Hubble


tklauser (approved):
Thanks!

PR #1744, opened 2023-06-21
install: Normalize --helm-* flags

michi-covalent:
Normalize the following --helm-* flags so that you can specify these flags with or without the "helm-" prefix:

* --helm-set: Normalize to --set.
* --helm-set-file: Normalize to --set-file.
* --helm-set-string: Normalize to --set-string.
* --helm-values: Normalize to --values, and add -f shorthand flag.

Fixes: #1701


asauber (approved):
clean 😎

PR #1758, opened 2023-06-23
connectivity tests: wait functions improvements and refactoring

giorio94:
Please refer to the individual commit messages for additional details.

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -5,13 +5,11 @@ package check
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"strconv"
-	"strings"
 	"sync"
 	"time"
 
+	"golang.org/x/exp/maps"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
@@ -1083,13 +1081,14 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	ct.Debug("Validating Deployments...")
 
 	srcDeployments, dstDeployments := ct.deploymentList()
-	if len(srcDeployments) > 0 {
-		if err := ct.waitForDeployments(ctx, ct.clients.src, srcDeployments); err != nil {
+	for _, name := range srcDeployments {
+		if err := WaitForDeployment(ctx, ct, ct.clients.src, ct.Params().TestNamespace, name); err != nil {
 			return err
 		}
 	}
-	if len(dstDeployments) > 0 {
-		if err := ct.waitForDeployments(ctx, ct.clients.dst, dstDeployments); err != nil {
+
+	for _, name := range dstDeployments {
+		if err := WaitForDeployment(ctx, ct, ct.clients.dst, ct.Params().TestNamespace, name); err != nil {
 			return err
 		}
 	}
@@ -1107,9 +1106,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 
 			// Individual endpoints will not be created for pods using node's network stack
 			if !ct.params.PerfHostNet {
-				ctx, cancel := context.WithTimeout(ctx, ct.params.ciliumEndpointTimeout())
-				defer cancel()
-				if err := ct.waitForCiliumEndpoint(ctx, ct.clients.src, ct.params.TestNamespace, perfPod.Name); err != nil {
+				if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, perfPod.Name); err != nil {
 					return err
 				}
 			}
@@ -1136,9 +1133,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	}
 
 	for _, pod := range clientPods.Items {
-		ctx, cancel := context.WithTimeout(ctx, ct.params.ciliumEndpointTimeout())
-		defer cancel()
-		if err := ct.waitForCiliumEndpoint(ctx, ct.clients.src, ct.params.TestNamespace, pod.Name); err != nil {
+		if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, pod.Name); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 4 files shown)


giorio94:
Converting back to draft to check failures.

giorio94:
All checks are green (it appears that the previous failures were unrelated from these changes).
Marking again ready for review.

connectivity/check/deployment.go
```diff
@@ -1206,10 +1197,8 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 	}
 
-	svcDNSCtx, svcDNSCancel := context.WithTimeout(ctx, ct.params.ipCacheTimeout())
-	defer svcDNSCancel()
 	for _, cp := range ct.clientPods {
-		err := ct.waitForServiceDNS(svcDNSCtx, cp)
+		err := WaitForCoreDNS(ctx, ct, cp)
```
derailed, line 1201: nit: one liner err check.

connectivity/check/wait.go
```diff
@@ -0,0 +1,250 @@
...
+const (
+	LongTimeout  = 5 * time.Minute
+	ShortTimeout = 30 * time.Second
+
+	PollInterval = 1 * time.Second
+)
+
+// WaitForDeployment waits until the specified deployment becomes ready.
+func WaitForDeployment(ctx context.Context, log Logger, client *k8s.Client, namespace string, name string) error {
+	log.Logf("⌛ [%s] Waiting for deployment %s/%s to become ready...", client.ClusterName(), namespace, name)
```
derailed, line 31: This is a common pattern across all subsequent waitFor calls. It seems it might be a  prime candidates for abstracting out these calls. 
Also do these need to be exported?

connectivity/check/wait.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// WaitForCiliumEndpoint waits until the specified cilium endpoint gets created.
+func WaitForCiliumEndpoint(ctx context.Context, log Logger, client *k8s.Client, namespace, name string) error {
+	log.Logf("⌛ [%s] Waiting for CiliumEndpoint for pod %s/%s to appear...", client.ClusterName(), namespace, name)
+
+	ctx, cancel := context.WithTimeout(ctx, ShortTimeout)
+	defer cancel()
+	for {
+		_, err := client.GetCiliumEndpoint(ctx, namespace, name, metav1.GetOptions{})
```
derailed, line 59: nit: one liner check

connectivity/check/wait.go
```diff
@@ -0,0 +1,250 @@
...
+// WaitForIPCache waits until all the specified pods are present in the IPCache of the given agent.
+func WaitForIPCache(ctx context.Context, log Logger, agent Pod, pods []Pod) error {
+	log.Logf("⌛ [%s] Waiting for Cilium pod %s to have all the pod IPs in eBPF IPCache...",
+		agent.K8sClient.ClusterName(), agent.Name())
+
+	ctx, cancel := context.WithTimeout(ctx, ShortTimeout)
+	defer cancel()
+
+	for {
+		err := validateIPCache(ctx, agent, pods)
```
derailed, line 280: nit: one liner check

connectivity/check/wait.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+// WaitForDeployment waits until the specified deployment becomes ready.
+func WaitForDeployment(ctx context.Context, log Logger, client *k8s.Client, namespace string, name string) error {
+	log.Logf("⌛ [%s] Waiting for deployment %s/%s to become ready...", client.ClusterName(), namespace, name)
+
+	ctx, cancel := context.WithTimeout(ctx, LongTimeout)
+	defer cancel()
+	for {
+		err := client.CheckDeploymentStatus(ctx, namespace, name)
```
derailed, line 36: nit: one liner check

derailed (changes requested):
@giorio94 Nice work Marco! A few picks and perhaps another potential refactor.

connectivity/check/deployment.go
giorio94: I'd personally keep as it is as now to reduce the code churn, given that it would be functionally equivalent. I don't have a strong opinion though, let me know in case you prefer that I change it.

connectivity/check/wait.go
giorio94: I don't think that this can be refactored to the be a one liner check, because we return if the error is nil, and we access it otherwise. The same for the ones below.

connectivity/check/deployment.go
derailed: Just a nit. Figured we should be consistent.

connectivity/check/wait.go
derailed: Thanks Marco. I see. Figured this is common logic and best to abstract with a closure. 
Personally I am not keen about making logging decisions deep in the stack as it should be the responsibility of the call site to determine whether to handle the error, log it, etc... If potentially multi errors can be surfaced then we should return these accordingly. If this is outside the scope of this review, I am good with that but figured I'll flag.

asauber:
Is part of this PR an attempt to quiet down this error message that we have been seeing in the multicluster workflow?

```
2023-06-26T06:50:33.155740215Z E0626 06:50:33.155593      26 memcache.go:287] couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```

Fine if not, we can address that elsewhere.

giorio94:
> Is part of this PR an attempt to quiet down this error message that we have been seeing in the multicluster workflow?
> 
> ```
> 2023-06-26T06:50:33.155740215Z E0626 06:50:33.155593      26 memcache.go:287] couldn't get resource list for metrics.k8s.io/v1beta1: the server is currently unable to handle the request
> ```
> 
> Fine if not, we can address that elsewhere.

Nope, we'll need another PR for that.

connectivity/check/deployment.go
giorio94: Makes sense. Let me update it.

connectivity/check/wait.go
giorio94: I appreciate the clarification and I see your point. I'd personally not change it further in this PR, to avoid modifying the pre-existing behavior. Especially considering the logs, given that the CLI used to output a debug log for every retry in case of errors.

michi-covalent:
all the feedback comments have been addressed either in this pull request or via some offline discussions. time to 🚢

PR #1768, opened 2023-06-27
connectivity: Add op=Exists toleration for test-conn-disrupt

brb:
It was observed on a Kind cluster, that during Cilium upgrade (v1.13.4 -> main) kubelet set the taint node.kubernetes.io/network-unavailable which evicted test-conn-disrupt pods. This resulted in a false negative in the upgrade tests.

Fix the issue by tolerating any taint.


asauber (approved):
Makes sense for these pods :+1:

michi-covalent:
is the button green? yes.

PR #1771, opened 2023-06-27
helm mode: don't suggest removed Azure flag

asauber:
In Helm mode, the flag `--azure-resource-group` was removed. However, we currently suggest it as a possible fix for various invocations against an AKS cluster. This change suggests only the Helm value in Helm mode.


michi-covalent (approved):
😇

tommyp1ckles (approved):

PR #1774, opened 2023-06-28
connectivity: Remove test-conn-disrupt tolerations

brb:
Unfortunately, the op=Exists toleration (127d9b8c2d) made the upgrade pods to be scheduled on nodes which do not run Cilium.

The upgrade issue got resolved by setting "cni.uninstall=false", so let's remove the toleration.


asauber (approved):
thanks for keeping an eye on this

PR #1776, opened 2023-06-29
sysdump: do not fail if cilium is not installed

kkourt:
If Cilium is not installed, the namespace of Cilium and Cilium operator cannot be detected. This means that, unless those two parameters are set by the user, the sysdump command will return an error. This means that sysdump cannot be used in Tetragon installations where Cilium is not installed.

This patch changes the code to split the tasks cilium and tetragon tasks, and disable the cilium tasks instead of erring out when the namesapce cannot be autodetected.


kkourt:
Multicluster failure seems unrelated since it does not involve sysdump.

tklauser (approved):
Thanks

PR #1780, opened 2023-06-30
go.mod, vendor: update cilium to v1.14.0-rc.0

tklauser:
Renovate does not seem to recognize that 1.14.0-rc.0 is newer than 1.14.0-snapshot.4. Update it manually.


sayboras:
Seems to be already done in https://github.com/cilium/cilium-cli/actions/runs/5420254427?pr=1763

PR #1796, opened 2023-07-04
Helm mode: Handle "cluster.name" Helm value

michi-covalent:
- Refactor preinstall to call HelmOpts.MergeValues() once at the top of
  the function, and pass the merged values to autodetectAndValidate().
- Set install.Parameters.ClusterName to cluster.name Helm value if it's
  not empty.
- Skip the cluster name validation altogether in Helm mode. I boldly
  assert that any input validation must be performed inside the Helm
  chart, and not in cilium-cli. This ensures that the validation logic
  is consistent between "helm install" and "cilium install".

Fixes: #490

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/cilium/cilium-cli/internal/utils"
 	"github.com/cilium/cilium-cli/k8s"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


install/autodetect.go
```diff
@@ -128,9 +129,20 @@ func (k *K8sInstaller) autodetect(ctx context.Context) {
...
 }
 
-func (k *K8sInstaller) autodetectAndValidate(ctx context.Context) error {
-	k.autodetect(ctx)
+func getClusterName(helmValues map[string]interface{}) string {
+	cluster, ok := helmValues["cluster"].(map[string]interface{})
+	if !ok {
+		return ""
+	}
+	chainingMode, ok := cluster["name"].(string)
```
asauber, line 137: `chainingMode` here should be `name`

asauber (approved):

install/autodetect.go
michi-covalent: this is what happens when you copy and paste functions

nebril (approved):

Issue #1371, opened 2023-02-04
Warning user if will run cilium sysdump with more than N nodes


aanm:
Getting a sysdump on a cluster with more than N nodes is not always required since usually problems are diagnosed between 1-3 nodes. Thus, we should add a "Sysdumps can be retrieved on particular nodes using the flag --nodes <node1,node2,node3>. By default, this command will grab information from all the N nodes on your cluster which might take a long period of time, do you want to continue? [y/N]" prompt. This will prevent users from hitting issues when getting sysdumps on large clusters.

tommyp1ckles:
Does this address this: https://github.com/cilium/cilium-cli/blob/master/sysdump/sysdump.go#L195?

(aside from there being no prompt)

PR #1799, opened 2023-07-05
Expand warning about big sysdump

nebril:
fixes: https://github.com/cilium/cilium-cli/issues/1371


derailed (changes requested):
@nebril LGTM - small pick... I think it would be good to show what we consider large ie 20 nodes + the current settings for the logs configs ie limit and since to make users aware of these settings. 
Interestingly 19 nodes running for ~1 year with current defaults could dump ~19Gib worth of agent logs with no warnings...

nebril:
Forgot to re-push :facepalm:

christarazi (approved):

derailed (approved):
LGTM

PR #1801, opened 2023-07-06
connectivity: don't wait for service backends if the IP family is disabled

giorio94:
0dccab2ea3f1 ("connectivity: wait for service propagation in agents") introduced an additional check to wait until all test services have been synchronized by cilium agents before starting the actual tests. Yet, this check will never succeed if the cluster is configured in dual stack mode, but only one IP family is enabled by Cilium. Indeed, the backends for the disabled IP family will never be populated. Hence, let's skip this check if the ClusterIP belongs to a disabled IP family.

Fixes: 0dccab2ea3f1 ("connectivity: wait for service propagation in agents")
Reported-by: Donia Chaiehloudj


doniacld (approved):
Tested on the cluster I observed it and it is working. 🚀 Thanks!

michi-covalent (approved):
😇

PR #1812, opened 2023-07-07
Fix pod summary for large numbers of pods

nebril:
This change fixes a bug which caused clusters with more than 500 pods to fail pod summary step.

The code assumed that if result.Infos had more than one element, the k8s request returned more than 1 resource type which is not correct for a table.

The result specifies pods as a resource type, so the assumption was wrong. Multiple Infos objects resulted from client dividing requests into chunks, as specified by RequestChunksOf function on request builder.

The change merges all Info objects into one table and returns it.

fixes: https://github.com/cilium/cilium-cli/issues/905


joamaki (approved):

pchaigno:
Nice catch! :tada:

PR #1816, opened 2023-07-10
hooks: Add new hook SetupAndValidate()

doniacld:
- In order to extend the possibility to setup and initialise for extended functionalities, a new hook is needed `SetupAndValidate`. 
- Create a method on service (like for pod) NameWithoutNamespace() to remove the namespace prefix. It will enable to make call to the domain easily.
- Realised that an error log was not properly formatted, just fixing it (not related to the PR).

### DIFF
--- a/cli/cmd.go
+++ b/cli/cmd.go
@@ -4,6 +4,8 @@
 package cli
 
 import (
+	"context"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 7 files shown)


tklauser (commented):
LGTM.

It looks like there are some unrelated formatting changes in the 2nd commit which are reverted in the 3rd commit. Might be worth to run `gofmt` at each commit to avoid these.

tklauser (approved):

tklauser:
> It looks like there are some unrelated formatting changes in the 2nd commit which are reverted in the 3rd commit. Might be worth to run gofmt at each commit to avoid these.

FYI, I've force pushed to this branch to omit the formatting changes so we can merge this for v0.15.2.

doniacld:
@tklauser Thanks a lot for fixing my commits ❤️ and merging!

PR #1829, opened 2023-07-12
Update README.md

saintdle:
corrected ```helm list``` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ To install the default version of Cilium:
 
 To see the Helm release that got deployed:
 
-    helm list -n kube-system cilium
+    helm list -n kube-system --filter "cilium"
 
 To see non-default Helm values that `cilium-cli` used for this Cilium installation:
 


asauber:
In the context of this example, the Helm release name will always be "cilium". I don't see this change as needed.

saintdle:
> In the context of this example, the Helm release name will always be "cilium". I don't see this change as needed.
> 
> I appreciate the attention to detail here. I'm going to close this PR in favor of simplicity in the front-matter examples.

Sure makes sense, however the provided command in the docs is incorrect. If the release name is always cilium, and you want to run the ```list``` command to only show that release, you need to use the ```--filter``` argument. You cannot simply declare the name of the release.

https://helm.sh/docs/helm/helm_list/

asauber (approved):

michi-covalent:
this is a readme file only change, not need to wait for the full tests. merging ✅

Issue #1773, opened 2023-06-28
cilium clustermesh status --wait does not wait for all clusters to be connected


marseel:
## Bug report

CI flake: https://github.com/cilium/cilium/issues/26513

Currently, `cilium clustermesh status --wait` does not wait on all clusters to be connected. It only checks connectivity status for clusters that cilium-agent already knows: https://github.com/cilium/cilium-cli/blob/9c3ea4c5e55ee7b6d466c57ade34091e0314b13a/status/k8s.go#L121
This means, that in the following scenario:
- Connect two clusters (clustermesh configmap is updated)
- Kubelet does not update configmap instantly - cilium agent does not know about the new remote cluster yet
- We check for cilium status for agents that are unaware of remote clusters, which passes the check

clustermesh is not fully initialized, but `clustermesh status --wait` finishes.

We should base it on the number of clusters that are expected in configmap and wait for all statuses to appear in `cilium status`

PR #1834, opened 2023-07-13
Improve the clustermesh status command

giorio94:
This PR includes a set of improvements for the `cilium clustermesh status` command, to simplify troubleshooting and reduce CI flakiness:

* Always check the readiness of the clustermesh-apiserver deployment.
* Report a node as not ready also when the configuration has not yet been detected (as the secret modification has not propagated yet).
* Report explicit errors about why a node is not ready, based on the additional information introduced in https://github.com/cilium/cilium/pull/26788 (e.g., etcd connection, cluster configuration retrieval, data synchronization).
* Remove duplicate checks and fix inaccuracies in output messages.

Please refer to the individual commit messages for additional information.

Fixes: https://github.com/cilium/cilium-cli/issues/1773

~~I'm currently marking this PR as ready for review as it makes sense that the review occurs in parallel with that of https://github.com/cilium/cilium/pull/26788, so that possible modifications can be synchronized. The commit with the go.mod changes will be dropped before merging.~~

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -1158,18 +1158,28 @@ func (k *K8sClusterMesh) statusAccessInformation(ctx context.Context, log bool, 
 	}
 }
 
-func (k *K8sClusterMesh) waitForDeployment(ctx context.Context) error {
-	k.Log("⌛ [%s] Waiting for deployment %s to become ready...", k.client.ClusterName(), defaults.ClusterMeshDeploymentName)
-
-	for k.client.CheckDeploymentStatus(ctx, k.params.Namespace, defaults.ClusterMeshDeploymentName) != nil {
-		select {
-		case <-time.After(time.Second):
-		case <-ctx.Done():
-			return fmt.Errorf("waiting for deployment %s to become ready has been interrupted: %w", defaults.ClusterMeshDeploymentName, ctx.Err())
+func (k *K8sClusterMesh) statusDeployment(ctx context.Context) (err error) {
+	w := utils.NewWaitObserver(ctx, utils.WaitParameters{Log: func(err error, wait string) {
+		k.Log("⌛ Waiting (%s) for deployment %s to become ready: %s", wait, defaults.ClusterMeshDeploymentName, err)
+	}})
+	defer func() {
+		w.Cancel()
+		if err != nil {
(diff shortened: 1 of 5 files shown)


rolinh (dismissed):
It looks like you forgot to remove the commit aptly named "TO DROP: use forked version of cilium"

giorio94:
> It looks like you forgot to remove the commit aptly named "TO DROP: use forked version of cilium"

Yeah, the commit is still there because otherwise the CLI doesn't compile. This is the reason for the `dont-merge/blocked label` being set, as it depends on https://github.com/cilium/cilium/pull/26788. At the same time, I wanted to mark this one ready for review so that they can be reviewed in parallel, and possible changes can be propagated if necessary.

asauber (commented):
We're now checking the health of the Deployment, but it may be the case that the Service is separately not able to receive traffic (e.g. an LB Service has not been assigned an IP). Is the idea here that if nodes have synced then the Service must be reachable?

giorio94:
> We're now checking the health of the Deployment, but it may be the case that the Service is separately not able to receive traffic (e.g. an LB Service has not been assigned an IP). Is the idea here that if nodes have synced then the Service must be reachable?

There's actually a service readiness check, but it is hidden inside the `statusAccessInformation` function (which in turn calls `extractAccessInformation`). The inner function returns an error (thus triggering the retry mechanism) if no service IP is found. This is also the reason for the removal of the explicit service readiness check in https://github.com/cilium/cilium-cli/pull/1834/commits/218fc9a24ec9cf0d3630e909255888ea69a063d5, as it was already tested earlier.

asauber (approved):

clustermesh/clustermesh.go
```diff
@@ -803,12 +804,14 @@ func (k *K8sClusterMesh) extractAccessInformation(ctx context.Context, client k8
...
 		ExternalWorkloadKey:  externalWorkloadKey,
 		ExternalWorkloadCert: externalWorkloadCert,
+		ServiceType:          string(svc.Spec.Type),
 		ServiceIPs:           []string{},
 		Tunnel:               cm.Data[configNameTunnel],
 	}
 
 	switch {
 	case len(endpoints) > 0:
+		svc.Spec.Type = "Manual"
```
YutaroHayakawa, line 817: What's this "Manual"?

YutaroHayakawa (approved):
LGTM with one question

PR #1857, opened 2023-07-21
connectivity: Special inconclusive result treatment for ping command

mhofstetter:
When running the connectivity tests in AKS, we sometimes get interrupted commands that don't have any output [1]. Therefore, successful command terminations with an empty output are treated as inconclusive that result in a retry [2].

Unfortunately, the ping command might get stuck by only outputting its header line to the output.

`PING 10.0.0.1 (10.0.0.1) 56(84) bytes of data.`

This commit adds an additional check for the ping commands. If we see an output only containing the header, we treat it as inconclusive result.

1 - cilium/cilium#22162
2 - cilium/cilium-cli#1543

### DIFF
--- a/connectivity/check/action.go
+++ b/connectivity/check/action.go
@@ -33,6 +33,11 @@ const (
 	testCommandRetries = 3
 )
 
+var (
+	// PING 10.0.0.1 (10.0.0.1) 56(84) bytes of data.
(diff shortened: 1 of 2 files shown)


michi-covalent (approved):
🏓

PR #1864, opened 2023-07-25
metrics: Set first retrieved metrics to 0 if absent

doniacld:
When a cluster is freshly deployed and no traffic was observed, certain metrics does not appear yet.  This commit intents to set to 0 the counters of the MetricFamily if the metric was not present.

Testing on a fresh cluster:

No metrics at the beginning, set to 0:
```
2023/07/25 10:45:42 [WARNING] metric xxx was not found before the action was run
2023/07/25 10:45:42 [WARNING] metric xxx counters was set to 0
```

Finally checked the metrics properly:
```
🐛 Forwarding from [::1]:60476 -> 9967

🐛 Handling connection for 60476

  🐛 checked metrics properly on node kind-worker

  🐛 Finalizing Test external-cilium-dns-proxy
  🐛 Pod kube-system/cilium-dzmnb's current policy revision: 2
  🐛 Pod kube-system/cilium-wmvfn's current policy revision: 2
  ℹ️  📜 Deleting CiliumNetworkPolicy 'dns-via-proxy-for-client' from namespace 'cilium-test'..
  🐛 Pod kind-kind/kube-system/cilium-dzmnb revision > 2
  🐛 Pod kind-kind/kube-system/cilium-wmvfn revision > 2
  🐛 📜 Successfully deleted 1 CiliumNetworkPolicies

✅ All 1 tests (4 actions) successful, 54 tests skipped, 0 scenarios skipped.
```

### DIFF
--- a/connectivity/check/metrics.go
+++ b/connectivity/check/metrics.go
@@ -116,6 +116,20 @@ func metricsIncrease(mf1, mf2 *dto.MetricFamily) error {
 	metrics1 := mf1.GetMetric()
 	metrics2 := mf2.GetMetric()
 
+	// Absent initial metric means it's zero, check that after metric is non-zero.
+	if mf1 == nil && mf2 != nil {
+		for _, m2 := range metrics2 {
+			if m2.GetCounter() == nil {
+				return fmt.Errorf("metric %s is not a Counter: %v", mf2.GetName(), mf2)
+			}
+
+			if val := m2.GetCounter().GetValue(); val == 0.0 {
+				return fmt.Errorf("metric %s did not increase as expected, value 1 was assumed zero: %f and value 2: %f", mf2.GetName(), 0.0, val)
+			}
+		}
+		return nil
+	}
+
 	if len(metrics1) != len(metrics2) {
 		return fmt.Errorf("metric %s has different length metrics 1: %d and metrics 2: %d", mf1.GetName(), len(metrics1), len(metrics2))
 	}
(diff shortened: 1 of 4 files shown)


connectivity/check/metrics.go
```diff
@@ -113,9 +113,24 @@ func parseMetrics(reader io.Reader) (promMetricsFamily, error) {
...
+			}
+
+			if val := m2.GetCounter().GetValue(); val == 0.0 {
+				return fmt.Errorf("metric %s did not increase as expected, value 1: %f and value 2: %f", mf2.GetName(), 0.0, val)
+			}
+		}
+		return nil
+	}
+
+	metrics1 := mf1.GetMetric()
```
michi-covalent, line 132: `mf1` can be nil right?

connectivity/check/result.go
```diff
@@ -167,19 +167,20 @@ func assertMetricsIncrease(metrics ...string) assertMetricsFunc {
 	return func(before, after map[string]*dto.MetricFamily) error {
 		var err error
 		for _, metricName := range metrics {
-			bValue, ok := before[metricName]
+			afterValue, ok := after[metricName]
```
michi-covalent, line 170: it makes i bit difficult to review because variable names changed, and also the order of checking `afterValue` and `beforeValue` flipped. could you keep the same order?

doniacld: Yes getting back to the original order! Thanks :)

connectivity/check/metrics.go
doniacld: Yes but if `mf2==nil`, we should have catch where the method is called meaning if mf1 is nil we should have failed before. 
I know that's a bit tricky...

gandro (changes requested):
Looks good to me overall.

One minor nit (https://github.com/cilium/cilium-cli/pull/1864/files#r1276205954) and one issue with the log package

connectivity/check/metrics.go
doniacld: Thanks! Moved at the top `mf1.GetMetric()`

PR #1879, opened 2023-07-31
dockerfile: include CA certificates

giorio94:
Busybox does not include CA certificates by default, which causes https requests to fail during certificate verification. For instance, the Cilium CLI fails to retrieve the hem chart with the following error:

```
looks like "https://helm.cilium.io" is not a valid chart repository or cannot be reached: Get "https://helm.cilium.io/index.yaml": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

Hence, let's propagate the CA certificates from the builder container.

Note: this error did not show up previously because the helm charts were vendored inside the CLI binary, hence avoiding the need to contact an external website. The v1.14 one, instead, is not vendored (yet?), hence causing all workflows to fail due to the impossibility of reaching `helm.cilium.io`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,13 @@
 
 FROM docker.io/library/golang:1.20.4-alpine3.16@sha256:6469405d7297f82d56195c90a3270b0806ef4bd897aa0628477d9959ab97a577 as builder
 WORKDIR /go/src/github.com/cilium/cilium-cli
-RUN apk add --no-cache git make
+RUN apk add --no-cache git make ca-certificates
 COPY . .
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
fascinating

giorio94:
Successful run for the external workloads workflow: https://github.com/cilium/cilium-cli/actions/runs/5715701135/job/15485617703?pr=1880 (it needed to be tested separately since this PR is from a fork, and the workflow is run as `pull_request_target`).

brlbil (approved):
LGTM

PR #1882, opened 2023-08-01
Update the error log exceptions

michi-covalent:
Add "removing identity not added to the identity manager!" to the error log exception until https://github.com/cilium/cilium/issues/16419 is fixed.


asauber (commented):
As the issue mentions, this error would be better as a warning.

asauber (approved):

PR #1887, opened 2023-08-02
Fix connectivity disruption tests for multi cluster scenario

giorio94:
Please refer to the individual commit messages for additional details about the specific changes.

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -632,13 +632,16 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 			ct.Failf("%s deployment is not ready: %s", testConnDisruptServerDeploymentName, err)
 		}
 
-		_, err = ct.clients.src.GetService(ctx, ct.params.TestNamespace, testConnDisruptServiceName, metav1.GetOptions{})
-		if err != nil {
-			ct.Logf("✨ [%s] Deploying %s service...", ct.clients.dst.ClusterName(), testConnDisruptServiceName)
-			svc := newService(testConnDisruptServiceName, map[string]string{"app": "test-conn-disrupt-server"}, nil, "http", 8000)
-			_, err = ct.clients.src.CreateService(ctx, ct.params.TestNamespace, svc, metav1.CreateOptions{})
+		for _, client := range ct.Clients() {
+			_, err = client.GetService(ctx, ct.params.TestNamespace, testConnDisruptServiceName, metav1.GetOptions{})
 			if err != nil {
-				return err
+				ct.Logf("✨ [%s] Deploying %s service...", client.ClusterName(), testConnDisruptServiceName)
+				svc := newService(testConnDisruptServiceName, map[string]string{"app": "test-conn-disrupt-server"}, nil, "http", 8000)
+				svc.ObjectMeta.Annotations = map[string]string{"service.cilium.io/global": "true"}
+				_, err = client.CreateService(ctx, ct.params.TestNamespace, svc, metav1.CreateOptions{})
+				if err != nil {
+					return fmt.Errorf("unable to create service %s: %w", testConnDisruptServiceName, err)
+				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


asauber:
This test doesn't appear to be exercised by CI.

There is [a clustermesh test which makes use of the `cilium upgrade` command](https://github.com/cilium/cilium-cli/blob/5fcc9421eb1f48ec274991d1cddfaad03dc21832/.github/workflows/kind.yaml#L301) and also has [a step which runs the connectivity tests](https://github.com/cilium/cilium-cli/blob/5fcc9421eb1f48ec274991d1cddfaad03dc21832/.github/workflows/kind.yaml#L325). It looks like this workflow could be updated with [the instructions for running this test](https://github.com/cilium/cilium-cli/blob/main/connectivity/tests/upgrade.go#L23-L32). How do you feel about adding that to this PR?

connectivity/check/deployment.go
```diff
@@ -389,24 +389,28 @@ func newIngress() *networkingv1.Ingress {
...
+			ct.Logf("✨ [%s] Creating namespace %s for connectivity check...", client.ClusterName(), ct.params.TestNamespace)
+			namespace := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        ct.params.TestNamespace,
+					Annotations: ct.params.NamespaceAnnotations,
+				},
+			}
+			_, err = client.CreateNamespace(ctx, namespace, metav1.CreateOptions{})
+			if err != nil {
+				return fmt.Errorf("unable to create namespace %s: %s", ct.params.TestNamespace, err)
```
asauber, line 412: This line and the error handling for the Service creation (line 643) should be made similar. A message could be added below and `%w` could be used for the err format specifier.

connectivity/check/deployment.go
```diff
@@ -1043,8 +1027,8 @@ func (ct *ConnectivityTest) deploymentList() (srcList []string, dstList []string
 	}
 
 	if ct.params.IncludeConnDisruptTest {
-		srcList = append(srcList, testConnDisruptClientDeploymentName)
-		dstList = append(dstList, testConnDisruptServerDeploymentName)
+		srcList = append(srcList, testConnDisruptServerDeploymentName)
```
asauber, line 1030: Could you add a comment here explaining why `src` is the Server and `dst` is the client? (are there even better names for them given the test scenario?). I see the commit message, but it's not clear to me why this is needed for the multicluster case.

giorio94: Commend added. I agree that `src` and `dst` are not extremely clear as names, but I would not change them as consistent with the rest of the occurrences. For more context, the clients are initialized [here](https://github.com/cilium/cilium-cli/blob/main/connectivity/check/context.go#L641-L705) (`src` and `dst` are identical when not running in multicluster mode), the server deployment [here](https://github.com/cilium/cilium-cli/blob/main/connectivity/check/deployment.go#L619-L622) (using the `src` client), and the client one [here](https://github.com/cilium/cilium-cli/blob/main/connectivity/check/deployment.go#L672-L675) (using the `dst` client).

giorio94:
> This test doesn't appear to be exercised by CI.

> There is [a clustermesh test which makes use of the cilium upgrade command](https://github.com/cilium/cilium-cli/blob/5fcc9421eb1f48ec274991d1cddfaad03dc21832/.github/workflows/kind.yaml#L301) and also has [a step which runs the connectivity tests](https://github.com/cilium/cilium-cli/blob/5fcc9421eb1f48ec274991d1cddfaad03dc21832/.github/workflows/kind.yaml#L325). It looks like this workflow could be updated with [the instructions for running this test](https://github.com/cilium/cilium-cli/blob/main/connectivity/tests/upgrade.go#L23-L32). How do you feel about adding that to this PR?

Yeah, makes sense. I've introduced that in the last commit. As mentioned in the description, the goal is only to ensure that the command works (both in the single and in the multi-cluster case), so no restarts/upgrades are performed in between. The actual clustermesh upgrade/downgrade tests are being introduced in https://github.com/cilium/cilium/pull/27232, but it still depends on other fixes to prevent connectivity disruptions.

aanm (approved):

tklauser (approved):

asauber (approved):

PR #1889, opened 2023-08-02
implement helm get values with sysdump

liyihuang:
include the helm get values in the sysdump collecting, and it will try to grab the user provide value and write to a file. when users provides a sysdump file, we don't have to ask the user to submit a separate helm value file. 

in this implementation, we just grab the latest value and user provide values for cilium. in the future, there can be more enhancement for this.

1. we can provide more options to collect previous version or the rendered helm values for everything.

btw, this is my first golang/OSS PR. feel free to let me know what I can do to make this better

### DIFF
--- a/internal/cli/cmd/sysdump.go
+++ b/internal/cli/cmd/sysdump.go
@@ -74,6 +74,9 @@ func initSysdumpFlags(cmd *cobra.Command, options *sysdump.Options, optionPrefix
 	cmd.Flags().StringVar(&options.CiliumEnvoyLabelSelector,
 		optionPrefix+"cilium-envoy-label-selector", sysdump.DefaultCiliumEnvoyLabelSelector,
 		"The labels used to target Cilium Envoy pods")
+	cmd.Flags().StringVar(&options.CiliumHelmReleaseName,
+		"cilium-helm-release-name", sysdump.DefaultCiliumHelmReleaseName,
+		"The Cilium Helm release name for which to get values")
 	cmd.Flags().StringVar(&options.CiliumOperatorLabelSelector,
 		optionPrefix+"cilium-operator-label-selector", sysdump.DefaultCiliumOperatorLabelSelector,
 		"The labels used to target Cilium operator pods")
--- a/k8s/client.go
+++ b/k8s/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/distribution/distribution/reference"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/output"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
@@ -1082,6 +1083,31 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName 
 		Version: version,
 		Values:  values,
 	}, nil
+}
+
+// GetHelmValues is the function for cilium cli sysdump to collect the helm values from the release directly
+func (c *Client) GetHelmValues(_ context.Context, releaseName string, namespace string) (string, error) {
+	if c.RESTClientGetter == nil {
+		return "", fmt.Errorf("no RESTClientGetter for Helm Values")
+	}
+	helmDriver := ""
+	actionConfig := action.Configuration{}
+	logger := func(format string, v ...interface{}) {}
+	if err := actionConfig.Init(c.RESTClientGetter, namespace, helmDriver, logger); err != nil {
+		return "", err
+	}
+	helmGetValsClient := action.NewGetValues(&actionConfig)
+	vals, err := helmGetValsClient.Run(releaseName)
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve helm value from release %s: %w", releaseName, err)
+	}
+
+	valuesBuf := new(bytes.Buffer)
+	if err = output.EncodeYAML(valuesBuf, vals); err != nil {
+		return "", fmt.Errorf("unable to parse helm values from release %s: %w", releaseName, err)
+	}
+	return valuesBuf.String(), nil
+
 }
 
 func (c *Client) ListAPIResources(_ context.Context) ([]string, error) {
--- a/sysdump/client.go
+++ b/sysdump/client.go
@@ -41,6 +41,7 @@ type KubernetesClient interface {
 	GetSecret(ctx context.Context, namespace, name string, opts metav1.GetOptions) (*corev1.Secret, error)
 	GetCiliumVersion(ctx context.Context, p *corev1.Pod) (*semver.Version, error)
 	GetVersion(ctx context.Context) (string, error)
+	GetHelmValues(ctx context.Context, releaseName string, namespace string) (string, error)
 	ListCiliumBGPPeeringPolicies(ctx context.Context, opts metav1.ListOptions) (*ciliumv2alpha1.CiliumBGPPeeringPolicyList, error)
 	ListCiliumCIDRGroups(ctx context.Context, opts metav1.ListOptions) (*ciliumv2alpha1.CiliumCIDRGroupList, error)
 	ListCiliumClusterwideNetworkPolicies(ctx context.Context, opts metav1.ListOptions) (*ciliumv2.CiliumClusterwideNetworkPolicyList, error)
--- a/sysdump/constants.go
+++ b/sysdump/constants.go
@@ -74,6 +74,7 @@ const (
 	cniConfigMapFileName                     = "cni-configmap-<ts>.yaml"
 	cniConfigFileName                        = "cniconf-%s-%s-<ts>.txt"
 	eniconfigsFileName                       = "aws-eniconfigs-<ts>.yaml"
+	ciliumHelmValuesFileName                 = "cilium-helm-values-<ts>.yaml"
 	gopsFileName                             = "gops-%s-%s-<ts>-%s.txt"
 	hubbleDaemonsetFileName                  = "hubble-daemonset-<ts>.yaml"
 	hubbleFlowsFileName                      = "hubble-flows-%s-<ts>.json"
--- a/sysdump/defaults.go
+++ b/sysdump/defaults.go
@@ -6,6 +6,8 @@ package sysdump
 import (
 	"runtime"
 	"time"
+
+	"github.com/cilium/cilium-cli/defaults"
 )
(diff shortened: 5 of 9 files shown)


internal/cli/cmd/sysdump.go
```diff
@@ -74,6 +74,9 @@ func initSysdumpFlags(cmd *cobra.Command, options *sysdump.Options, optionPrefix
 	cmd.Flags().StringVar(&options.CiliumEnvoyLabelSelector,
 		optionPrefix+"cilium-envoy-label-selector", sysdump.DefaultCiliumEnvoyLabelSelector,
 		"The labels used to target Cilium Envoy pods")
+	cmd.Flags().StringVar(&options.CiliumHelmChartReleaseName,
+		"cilium-helm-chart-release-name", sysdump.DefaultCiliumHelmChartReleaseName,
+		"Cilium Helm chart release name to get the values from the sysdump")
```
asauber, line 79: This description is a bit redundant in context. A better one would be 

"The release name to use when retrieving Helm values"

Also, the word "chart" is a bit redundant here as well, as all releases are associated with a chart. So, this could be changed to `cilium-helm-release-name` and `CiliumHelmReleaseName`.

k8s/client.go
```diff
@@ -1084,6 +1085,22 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
 	}, nil
 }
 
+// GetHelmValues is the function for cilium cli sysdump to collect the helm values from the release directly
+func (c *Client) GetHelmValues(_ context.Context, releaseName string) (string, error) {
+	client := action.NewGetValues(c.HelmActionConfig)
```
asauber, line 1090: Rather than `client`, this is more accurately an `action`.

k8s/client.go
```diff
@@ -1084,6 +1085,22 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
...
+	vals, err := client.Run(releaseName)
+	if err != nil {
+		return "", fmt.Errorf("unable to retrieve the helm value from the release %s: %w", releaseName, err)
```
asauber, line 1093: This message is better as "unable to retrieve helm values from release %s: %w"

k8s/client.go
```diff
@@ -1084,6 +1085,22 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
...
+	}
+
+	valuesBuf := new(bytes.Buffer)
+	if err = output.EncodeYAML(valuesBuf, vals); err != nil {
+		return "", fmt.Errorf("unable to parse from helm values from release %s: %w", releaseName, err)
```
asauber, line 1098: This message is better as "unable to parse helm values from release %s: %w"

asauber (commented):
This looks great overall. I have a few comments regarding names and error messages.

Please re-request my review if you end up waiting on me!

k8s/client.go
liyihuang: https://github.com/helm/helm/blob/8854547d3518cf3c83eec80d9390a05c453f4b3e/cmd/helm/get_values.go#L42

This is called client in helm...

k8s/client.go
liyihuang: I thought this way we can better identify where the error is. May I know why we keep it the same for different err?

internal/cli/cmd/sysdump.go
liyihuang: will change in my next commit

k8s/client.go
liyihuang: will change in my next commit

k8s/client.go
liyihuang: will change in my next commit

k8s/client.go
liyihuang: it returns a `GetValues` struct, there are different structs in action pkg, and I will rename it to `helmGetValsClient` hope its more accurate.

mtt0:
If we want hornor `--cilium-namespace` sysdump flag ranther than the global flag `-n`, workaround:

Add:
```golang
helmDriver := ""
logger := func(format string, v ...interface{}) {}
k8sClient.RESTClientGetter.Namespace = sysdumpOptions.CiliumNamespace
if err := k8sClient.HelmActionConfig.Init(&k8sClient.RESTClientGetter, sysdumpOptions.CiliumNamespace, helmDriver, logger); err != nil {
  return fmt.Errorf("failed to init helm action config: %w", err)
}
```
below
https://github.com/cilium/cilium-cli/blob/ccc2fcb396d4706df68cb39b27d7882f0c193108/internal/cli/cmd/sysdump.go#L32-L35

liyihuang:
> If we want hornor `--cilium-namespace` sysdump flag ranther than the global flag `-n`, workaround:
> 
> Add:
> 
> ```go
> helmDriver := ""
> logger := func(format string, v ...interface{}) {}
> k8sClient.RESTClientGetter.Namespace = sysdumpOptions.CiliumNamespace
> if err := k8sClient.HelmActionConfig.Init(&k8sClient.RESTClientGetter, sysdumpOptions.CiliumNamespace, helmDriver, logger); err != nil {
>   return fmt.Errorf("failed to init helm action config: %w", err)
> }
> ```
> 
> below
> 
> https://github.com/cilium/cilium-cli/blob/ccc2fcb396d4706df68cb39b27d7882f0c193108/internal/cli/cmd/sysdump.go#L32-L35
> 
> * If `-n` present and `--cilium-namespace` not, then `-n` tack effect
> * If `--cilium-namespace` present, then `--cilium-namespace` tack effect and override `-n` （`k8sClient.RESTClientGetter.Namespace` and `k8sClient.HelmActionConfig` changed)
> 
> Not sure if there are other side effects, whether it will affect other commands, because I know nothing about cilium-cli, sorry😣

thanks for the reply. 

I don't quite understand why we need init the NS twice here and I see we dont need to worry about NS in the RESTClientGetter here https://github.com/cilium/cilium-cli/blob/b98ada85d891f27fbf6bf660d233d991145c5a41/k8s/client.go#L125-L130

btw, `RESTClientGetter` is the interface of the k8s client, I dont think I can access the struct directly, where I might need to init another struct.

It might be an ok for the current implementation.

k8s/client.go
asauber: Since there is an “action config”, it would makes sense to call this an “action” (does it satisfy an Action interface maybe?). Then it makes more sense for an action to “Run” rather than for a client to “Run”, usuall a client would “request” or “get” in this case.

Either way, if the official codebase is doing it, this is fine : )

liyihuang: I already updated it to `helmGetValsClient` in my latest commit and pushed my commit last week. 

btw, this is not something urgent, and I can wait for next week for more discussion or review

mtt0:
Oh...this project is using the `vendor` directory, so RESTClientGetter is indeed an interface...

[We need switch kubeClient namespace as well](https://github.com/helm/helm/issues/9256#issuecomment-763767654). Because in `action.Init()`, kubeClient comes from getter:

https://github.com/helm/helm/blob/37cc2fa5cefb7f5bb97905b09a2a19b8c05c989f/pkg/action/action.go#L370

See:
- https://github.com/helm/helm/issues/9256
- https://github.com/helm/helm/issues/8269#issuecomment-642060058
for why

k8s/client.go
```diff
@@ -1084,6 +1085,39 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
...
+func (c *Client) GetRESTClientGetter() (*genericclioptions.RESTClientGetter, error) {
+	if c.RESTClientGetter == nil {
+		return nil, fmt.Errorf("unable to get the RESTClientGetter")
+	}
+	return &c.RESTClientGetter, nil
+}
+
+// GetHelmValues is the function for cilium cli sysdump to collect the helm values from the release directly
+func (c *Client) GetHelmValues(_ context.Context, releaseName string, namespace string) (string, error) {
+	restClientGetter, err := c.GetRESTClientGetter()
```
asauber, line 1097: Rather than add this function, how about just performing the `nil` check here?

k8s/client.go
```diff
@@ -1084,6 +1085,39 @@ func (c *Client) GetHelmState(ctx context.Context, namespace string, secretName
...
+	if err != nil {
+		return "", fmt.Errorf("unable to get the RESTClientGetter from the client")
```
asauber, line 1099: If we do keep this function, you can `return "", err` here, since the error messages are essentially identical.

asauber (changes requested):
Looks great for the most part.

Can you squash the commits down to 1 or 2 commits? Maybe with supporting the additional namespace flag as the second commit.

It's also important that the CLI is fully usable after each commit, which might be one reason to squash down to 1.

k8s/client.go
liyihuang: My understanding is we have to keep this function since `RESTClientGetter` the interface and I can't just get the struct if I dont implement this function. Isn't it?

asauber: You're implementing `GetRESTClientGetter` with the receiver `(c *Client)`, so the only interface that you could be satisfying with that implementation would be an interface for `*Client`. However, you are not passing the value of `c` in this function as any interface that expects a `GetRESTClientGetter` method.

Therefore you can replace this line and the function above with:
```
	if c.RESTClientGetter == nil {
		return nil, fmt.Errorf("unable to get RESTClientGetter for Helm Values")
	}
```

And then later in this function use `&c.RESTClientGetter` wherever you are using `restClientGetter`.

k8s/client.go
asauber: With the above section, you can remove this block

internal/cli/cmd/sysdump.go
```diff
@@ -74,6 +74,9 @@ func initSysdumpFlags(cmd *cobra.Command, options *sysdump.Options, optionPrefix
 	cmd.Flags().StringVar(&options.CiliumEnvoyLabelSelector,
 		optionPrefix+"cilium-envoy-label-selector", sysdump.DefaultCiliumEnvoyLabelSelector,
 		"The labels used to target Cilium Envoy pods")
+	cmd.Flags().StringVar(&options.CiliumHelmReleaseName,
+		"cilium-helm-release-name", sysdump.DefaultCiliumHelmReleaseName,
+		"Cilium Helm release name to get the values from the sysdump")
```
asauber, line 79: "The Cilium Helm release name for which to get values"

(referring to the sysdump here is redundant)

asauber (approved):
One style comment, otherwise LGTM

asauber (approved):

sayboras (approved):
LGTM ✅

PR #1893, opened 2023-08-04
connectivity: get nodename on which a pod is running via PodSpec

jschwinger233:
Previously we tried to get it via PodStatus.NominatedNodeName, which is usually set to empty.

Signed-off-by: Zhichuan Liang <gray.liang@isovalent.com>


jschwinger233:
cc @brb

brb:
Thanks for the fix!

Last time I've checked, the `pod.Spec.NodeName` was empty, and only set by a user - https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes/#create-a-pod-that-gets-scheduled-to-specific-node.

jschwinger233:
@brb I noticed this issue at our tests-ipsec-upgrade CI jobs: https://github.com/cilium/cilium/actions/runs/5746964893/job/15577230167

```
# 417L

[=] Test [no-ipsec-xfrm-errors]

  [-] Scenario [no-ipsec-xfrm-errors/no-ipsec-xfrm-error]
  🟥 Node  xfrm errors were changed (previous errors: , current errors: inbound_no_state:6)
```

The log format is 
https://github.com/cilium/cilium-cli/blob/7bcee268141748473ae7d92126c703f3aa90326e/connectivity/tests/ipsec_xfrm.go#L52
so we can tell the nodename (from `nominatedNodeName`) was empty over there.

brb (approved):

tklauser (approved):

Issue #1702, opened 2023-06-05
helm mode: `🔮 Auto-detected` log messages don't take `--helm-*` flags into account


michi-covalent:
some of these `🔮 Auto-detected` log messages don't make sense when you use `--helm-*` flags (`--helm-set=cluster.name=foobar` for example):

- https://github.com/cilium/cilium-cli/blob/9f9f6148409c2da428cb3a667e3703d2034a23a7/install/autodetect.go#L118
- https://github.com/cilium/cilium-cli/blob/9f9f6148409c2da428cb3a667e3703d2034a23a7/install/autodetect.go#L156

a simple fix would be to remove these log messages. in helm mode there are `--dry-run` / `--dry-run-helm-values` flags, so these log messages are redundant.

PR #1918, opened 2023-08-15
install: Remove some auto-detection log messages

michi-covalent:
Remove some auto-detection log messages from the install command output. There is --dry-run-helm-values flag to see exactly what Helm values are being set.

Fixes: #1702


install/autodetect.go
```diff
@@ -84,7 +84,6 @@ func (k *K8sInstaller) detectDatapathMode(withKPR bool) error {
 		k.params.DatapathMode = DatapathTunnel
 
 		if withKPR && k.params.KubeProxyReplacement == "" {
-			k.Log("ℹ️  kube-proxy-replacement disabled")
```
tklauser, line 87: Should we remove the same log message for AKS [on line 110](https://github.com/cilium/cilium-cli/pull/1918/files#diff-70c8da281ad8db17039130d2e7ad09aa2fc7db3f6d44217e6db0764bef731fffR110) as well?

michi-covalent: yeah let's do that 👍 i'll update the PR today

tklauser (approved):

PR #1938, opened 2023-08-25
clustermesh: Add support for kubernetes service annotations

joaoubaldo:
Add a `--service-annotation` argument so that custom annotations can be added to the Kubernetes Service when clustermesh is being enabled.  
A use case for this flag is to create a private NLB in AWS instead of a public classic LB:
> cilium clustermesh enable --service-type LoadBalancer --service-annotation 'service.beta.kubernetes.io/aws-load-balancer-internal=0.0.0.0/0' --service-annotation 'service.beta.kubernetes.io/aws-load-balancer-type=nlb'


thorn3r:
@joaoubaldo looks like you're hitting some CI failures that were fixed last month but your branch base is out of date. Could you rebase onto latest?

tklauser:
@thorn3r could you please take another look?

thorn3r (approved):
thanks, looks great!

PR #1947, opened 2023-09-01
Add pod-to-pod strict mode tests

3u13r:
This PR implements the pod-to-pod strict mode tests in the cilium-cli test suite. Currently, there are similar tests inside the legacy vagrant test suite (https://github.com/cilium/cilium/blob/836598a317565d4c53c678bb09173ae9fee5f54b/test/k8s/datapath_configuration.go#L387). 

With this PR we could deprecate and remove the vagrant tests if we add strict mode versions to Cilium's conformance CI. The difference is that the vagrant tests control the Kubernetes environment and cilium configuration for each test. Therefore the test coverage regarding different cilium configurations is easier to achieve. 

My question is: If the vagrant tests are "legacy", what is the best way to implement potentially disrupting tests which need to perform changes to the K8s environment? It seems like that this test suite was originally developed to implement tests which the user can execute in their production environment without any disruption. I guess this is the discussion you have right now: https://github.com/cilium/design-cfps/pull/9#discussion_r1218651510.

### DIFF
--- a/connectivity/tests/encryption.go
+++ b/connectivity/tests/encryption.go
@@ -119,9 +119,13 @@ func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
 	})
 }
 
-func testNoTrafficLeak(ctx context.Context, t *check.Test, s check.Scenario,
+
+// startTcpdump starts tcpdump in the background, and returns a cancel function
+// to stop it, and a channel which is closed when tcpdump has exited.
+// It writes captured pkts to /tmp/$TEST_NAME.pcap.
+func startTcpdump(ctx context.Context, t *check.Test,
 	client, server, clientHost *check.Pod, reqType requestType, ipFam check.IPFamily,
-) {
+) (context.CancelFunc, chan struct{}) {
 	dstAddr := server.Address(ipFam)
 	iface := getInterNodeIface(ctx, t, clientHost, dstAddr)
 	srcAddr := getSourceAddress(ctx, t, client, clientHost, ipFam, dstAddr)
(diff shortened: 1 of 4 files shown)


margamanterola:
@brb Could you elaborate on the flakiness risk? Is there any way we could avoid that risk that doesn't require the control plane tests?

brb:
@margamanterola The test needs to scale down cilium-operator replicas count to 0, then removes some CRDs, runs tests, and brings back cilium-operator replica count to a previous. The problem I see is that some CLI test cases might be missing proper wait statements which could result in flakiness (the missing waits are notorious in Cilium / CLI). Another issue is that we assume the CLI connectivity tests won't modify Cilium instances. But if you think that control-plane e2e doesn't happen soon, then we should consider that option.

3u13r:
I see potential flakiness reasons in:
* the test not resetting the operator scale correctly
* not being able to concurrently execute tests, which is also not possible for the other network policy tests I guess. The tests in this PR just also ignore any K8s namespace boundary. 

If there are waits missing in the other tests we always add verification that everything was restored correctly. We kinda do this already since we first test that the traffic is blocked and then also test that after the operator was scaled up again and the endpoint we previously deleted has been restored, the connection succeeds. 

To summarize: we delete one endpoint and explicitly wait for its regeneration. We can still have missing endpoints but I don't think the risk is _much_ higher than when e.g. new pods are spawned. 

I mostly ported the test:
* to migrate away from the legacy tests should they be shut down in the future
* to get a feel how to develop strict tests to develop a node-to-node strict mode via TDD

If there's another test framework/suite which better fits the requirements of those tests I'm happy to wait.

connectivity/tests/encryption.go
```diff
@@ -224,6 +254,125 @@ func testNoTrafficLeak(ctx context.Context, t *check.Test, s check.Scenario,
...
+		}
+		savedReplicas := scale.Spec.Replicas
+		scale.Spec.Replicas = replicas
+		if _, err := clientHost.K8sClient.Clientset.AppsV1().Deployments(t.Context().Params().CiliumNamespace).UpdateScale(ctx, "cilium-operator", scale, metav1.UpdateOptions{}); err != nil {
+			t.Fatalf("Failed to scale cilium-operator: %s", err)
+		}
+		return savedReplicas
+	}
+	waitForIPCacheEntry := func(clientPod, dstPod *check.Pod) {
+		timeout := time.After(20 * time.Second)
```
derailed, line 294: nit: Perhaps using consts here for the timeout and sleep durations would make it easier to adjust in the future?

derailed (commented):
@3u13r Nice work! Thank you for this PR! Just a few small picks...

brb:
@3u13r Are you in Cilium's Slack? Let's discuss it in #development channel.

connectivity/tests/encryption.go
```diff
@@ -115,13 +116,55 @@ func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
...
 
 	t.ForEachIPFamily(func(ipFam check.IPFamily) {
-		testNoTrafficLeak(ctx, t, s, client, &server, &clientHost, requestHTTP, ipFam)
+		testNoTrafficLeak(ctx, t, client, &server, &clientHost, requestHTTP, ipFam, false)
 	})
 }
 
-func testNoTrafficLeak(ctx context.Context, t *check.Test, s check.Scenario,
+// PodToPodEncryption is a test case which checks the following:
+//   - There is a connectivity between pods on different nodes when any
```
ti-mo, line 124:
```suggestion
//   - There is connectivity between pods on different nodes when any
```

connectivity/tests/encryption.go
```diff
@@ -115,13 +116,55 @@ func (s *podToPodEncryption) Run(ctx context.Context, t *check.Test) {
...
+//     encryption mode is on (either WireGuard or IPsec).
+//   - No unencrypted packet is leaked.
```
ti-mo, line 126: Could you clarify what is meant by 'leaked' here?

PR #1952, opened 2023-09-07
connectivity: Delete test-conn-disrupt pods after running tests

brb:
In the CI upgrade tests we sometimes see the following:

    The failure connectivity test failed: timeout reached waiting for
    deployment cilium-test/test-conn-disrupt-client to become ready
    (last error: only 3 of 5 replicas are available)

@jschwinger233 pointed out that it might be related to "--flush-ct", because according to some output of "kubectl get po", the test-conn-disrupt client pods crashed at the end of last connectivity test, when the "--flush-ct" took place.

Tested with https://github.com/cilium/cilium/pull/27966

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -414,6 +414,17 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 		wg.Wait()
 	}
 
+	// Delete conn disrupt pods after running tests. Otherwise, after calling
+	// --flush-ct they might get into crashloop state from which it takes time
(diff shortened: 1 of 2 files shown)


jschwinger233 (approved):

jschwinger233:
Do you think we can merge this first? I saw many were suffering from ipsec-upgrade-test https://github.com/cilium/cilium/actions/workflows/tests-ipsec-upgrade.yaml

nebril (approved):

PR #1955, opened 2023-09-08
connectivity: Decrease test-conn-disrupt-client interval to 10ms

brb:
Depends on https://github.com/cilium/cilium-cli/pull/1952

cc @jschwinger233

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -613,7 +613,7 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 			testConnDisruptServerDeployment := newDeployment(deploymentParameters{
 				Name:           testConnDisruptServerDeploymentName,
 				Kind:           KindTestConnDisrupt,
-				Image:          "quay.io/cilium/test-connection-disruption:v0.0.4",
+				Image:          "quay.io/cilium/test-connection-disruption:v0.0.5",
 				Replicas:       3,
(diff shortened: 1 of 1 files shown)


jschwinger233 (approved):

connectivity/check/deployment.go
```diff
@@ -665,12 +665,13 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
 				Name:     testConnDisruptClientDeploymentName,
 				Kind:     KindTestConnDisrupt,
-				Image:    "quay.io/cilium/test-connection-disruption:v0.0.4",
+				Image:    "quay.io/cilium/test-connection-disruption:v0.0.5",
 				Replicas: 5,
 				Labels:   map[string]string{"app": "test-conn-disrupt-client"},
 				Port:     8000,
 				Command: []string{
 					"tcd-client",
+					"--dispatch-interval", "10", // 10ms
```
michi-covalent, line 674: unrelated to this pr, but in general it's better to use duration flags so that you can simply say `"--dispatch-interval", "10ms"`

michi-covalent (approved):
🍍

nebril (approved):

connectivity/check/deployment.go
brb: :+1: will do it as a follow-up

PR #1968, opened 2023-09-12
connectivity: Add retries to PodToPod curl commands

learnitall:
This commit adds retries to the curl commands in the PodToPod scenario. The goal of adding this change is to try and address the Cilium CI flake cilium/cilium#27762.


christarazi:
I'm not sure we want to accept retries for pod-to-pod traffic because that's a recipe for hiding bugs. Retries to external targets like 1.1.1.1 are ok for retries because they are entities outside of our control.

learnitall:
Ah ok that makes sense. I'll close this one out.

PR #1975, opened 2023-09-15
check: add methods to export TestNamespace and CiliumNamespace

jibi:


michi-covalent:
just to add a bit of context, `ConnectivityTest` has `Params()` method that returns `Parameters`:

https://github.com/cilium/cilium-cli/blob/bb0ef930330231c69c091962048720865541e28c/connectivity/check/context.go#L876-L878

so you don't need to add separate methods to access individual fields inside `Parameters` 📝

PR #1986, opened 2023-09-26
utils: export WaitObserver methods into wait package

jibi:
to make them available to hook extensions

PR #1987, opened 2023-09-26
golangci-lint: enable goheader check

tklauser:
To make sure all Go files have a proper copyright header. Also add the copyright header to files where it was previously missing.

Ref. https://github.com/cilium/cilium-cli/pull/1986/files#r1337029656


sayboras (approved):
🚢

PR #1988, opened 2023-09-26
connectivity: test referencing nodes by CIDR

squeed:
We now support selecting cluster nodes in policy by CIDR. Previously, only the `remote-node` and `host` entities could select nodes. So, add a corresponding feature and test case for this.

This required adding a function to autodetect node CIDRs.


tklauser (approved):

nathanjsweet (approved):
Looks good!

Issue #1962, opened 2023-09-12
Quieten expected warnings


joestringer:
When running `cilium sysdump` in a kind dev environment with a relatively simple deployment, the command complains with many warnings that do not need to be warned about:

```
...
I0911 16:35:09.580937 3224081 request.go:697] Waited for 1.167974212s due to client-side throttling, not priority and fairness, request: GET:https://127.0.0.1:41805/api/v1/namespaces/kube-system/pods?labelSelector=k8s-app%3Dcilium
⚠️ ConfigMap "spire-agent" not found in namespace "kube-system" - this is expected if SPIRE installation is not enabled
⚠️ ConfigMap "spire-server" not found in namespace "kube-system" - this is expected if SPIRE installation is not enabled
I0911 16:35:19.780446 3224081 request.go:697] Waited for 2.945800215s due to client-side throttling, not priority and fairness, request: GET:https://127.0.0.1:41805/api/v1/namespaces/kube-system/pods/cilium-d7rr7/log?container=mount-bpf-fs&limitBytes=1073741824&sinceTime=2022-09-11T23%3A35%3A12Z&timestamps=true
I0911 16:35:29.780784 3224081 request.go:697] Waited for 3.779628289s due to client-side throttling, not priority and fairness, request: GET:https://127.0.0.1:41805/api/v1/namespaces/kube-system/pods/kube-controller-manager-kind-control-plane/log?container=kube-controller-manager&limitBytes=1073741824&sinceTime=2022-09-11T23%3A35%3A12Z&timestamps=true 
⚠️ The following tasks failed, the sysdump may be incomplete:
⚠️ [12] Collecting Cilium egress NAT policies: failed to collect Cilium egress NAT policies: the server could not find the requested resource
⚠️ [13] Collecting Cilium Egress Gateway policies: failed to collect Cilium Egress Gateway policies: the server could not find the requested resource (get ciliumegressgatewaypolicies.cilium.io)
⚠️ [15] Collecting Cilium local redirect policies: failed to collect Cilium local redirect policies: the server could not find the requested resource (get ciliumlocalredirectpolicies.cilium.io)
⚠️ [17] Collecting Cilium endpoint slices: failed to collect Cilium endpoint slices: the server could not find the requested resource (get ciliumendpointslices.cilium.io)
[...]

PR #1990, opened 2023-09-26
Clean up the node feature detection logic

michi-covalent:
- Remove nodesWithoutCilium list as it is redundant, and rename nodesWithoutCiliumMap to nodesWithoutCilium.
- Initialize both nodes and nodesWithoutCilium in getNodes to avoid listing nodes twice.
- Associate extractFeaturesFromNodes with FeatureSet instead of ConnectivityTest to decouple FeatureSet from ConnectivityTest.

Ref: #1962

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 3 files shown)


tklauser (approved):
🧹

PR #1992, opened 2023-09-28
sysdump: Cache Cilium pods and cilium-config ConfigMap

michi-covalent:
Get Cilium pods and cilium-config ConfigMap once at the beginning to reduce the number of Kubernetes API calls.

Ref: #1962

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -147,6 +147,10 @@ type Collector struct {
 	allNodes *corev1.NodeList
 	// NodeList is a list of nodes to collect sysdump information from.
 	NodeList []string
+	// CiliumPods is a list of Cilium agent pods running on nodes in NodeList.
+	CiliumPods []*corev1.Pod
+	// CiliumConfigMap is a pointer to cilium-config ConfigMap.
+	CiliumConfigMap *corev1.ConfigMap
 	// additionalTasks keeps track of additional tasks added via AddTasks.
 	additionalTasks []Task
 }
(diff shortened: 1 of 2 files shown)


sysdump/sysdump.go
```diff
@@ -738,11 +760,10 @@ func (c *Collector) Run() error {
 			Description: "Collecting the Cilium configuration",
 			Quick:       true,
 			Task: func(ctx context.Context) error {
-				v, err := c.Client.GetConfigMap(ctx, c.Options.CiliumNamespace, ciliumConfigMapName, metav1.GetOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to collect the Cilium configuration: %w", err)
+				if c.CiliumConfigMap == nil {
+					return nil
```
derailed, line 764: Should we return an error here as we did prior to the change if the cache cm is nil?

sysdump/sysdump.go
```diff
@@ -963,13 +984,7 @@ func (c *Collector) Run() error {
...
 			Quick:           true,
 			Task: func(ctx context.Context) error {
-				p, err := c.Client.ListPods(ctx, c.Options.CiliumNamespace, metav1.ListOptions{
-					LabelSelector: c.Options.CiliumLabelSelector,
-				})
-				if err != nil {
-					return fmt.Errorf("failed to list Cilium pods: %w", err)
-				}
-				if err := c.SubmitCniConflistSubtask(FilterPods(p, c.NodeList), ciliumAgentContainerName); err != nil {
+				if err := c.SubmitCniConflistSubtask(c.CiliumPods, ciliumAgentContainerName); err != nil {
```
derailed, line 987: Should we raise an error if the filtered list of pods is empty?

sysdump/sysdump.go
```diff
@@ -998,13 +1013,7 @@ func (c *Collector) Run() error {
...
 			Quick:           true,
 			Task: func(ctx context.Context) error {
-				p, err := c.Client.ListPods(ctx, c.Options.CiliumNamespace, metav1.ListOptions{
-					LabelSelector: c.Options.CiliumLabelSelector,
-				})
-				if err != nil {
-					return fmt.Errorf("failed to get Cilium pods: %w", err)
-				}
-				if err := c.SubmitGopsSubtasks(FilterPods(p, c.NodeList), ciliumAgentContainerName); err != nil {
+				if err := c.SubmitGopsSubtasks(c.CiliumPods, ciliumAgentContainerName); err != nil {
```
derailed, line 1016: Same comment as above and subsequent if no cilium pods are present

derailed (changes requested):
@michi-covalent Nice work! A few small picks...

sysdump/sysdump.go
michi-covalent: i think what we should do is to not even start these tasks if there are no cilium pods anyways.

that would be a bigger change than i'd like to make in this pull request. let me do that as a follow up PR 🚀🙏

sysdump/sysdump.go
michi-covalent: in general i'm trying to avoid raising errors. the reason is that the sysdump is also used to collect tetragon-related resources. in that case it is expected that cilium-config configmap is not found.

note that we still log when sysdump fails to get cilium-config configmap in https://github.com/cilium/cilium-cli/pull/1992/files#diff-e9c40a8e7876ce0237b03aba3f2b4f81ad70f07ae0b8ff1cffa57e76ef5eec8fR260-R265

michi-covalent (commented):
hello @derailed thanks for your feedback, i replied to your comments. could you take a look again ✨🙏

derailed (approved):
@michi-covalent  Thanks for the clarification!

PR #2004, opened 2023-10-04
sysdump: Detect features from cilium-config ConfigMap

michi-covalent:
- Call FeatureSet.ExtractFromConfigMap to extract features from cilium-config.
- Add tasks related to SPIRE server if and only if "mutual-auth-spiffe" feature is enabled.

Ref: #1962

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -29,6 +29,7 @@ import (
 	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/internal/utils"
 	"github.com/cilium/cilium-cli/k8s"
+	"github.com/cilium/cilium-cli/utils/features"
 )
 
 const sysdumpLogFile = "cilium-sysdump.log"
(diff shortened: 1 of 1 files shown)


michi-covalent:
requesting review from service mesh team 🚀🙏

sysdump/sysdump.go
```diff
@@ -263,6 +267,14 @@ func NewCollector(k KubernetesClient, o Options, startTime time.Time, cliVersion
 			}
 			c.log("ℹ️  %s ConfigMap not found in %s namespace", ciliumConfigMapName, c.Options.CiliumNamespace)
 		}
+		if c.CiliumConfigMap != nil && len(c.CiliumPods) > 0 {
+			ciliumVersion, err := c.Client.GetCiliumVersion(context.Background(), c.CiliumPods[0])
```
derailed, line 271: Yikes! hopefully we'll have a better way to retrieve version info... perhaps helm values??

derailed (approved):
@michi-covalent Nice work!

sysdump/sysdump.go
michi-covalent: yeah this is not great, but it needs to work even when you install cilium without doing `helm install`, like `helm template | kubectl apply -` or similar.

meyskens (approved):
LGTM from service mesh

PR #2008, opened 2023-10-05
helm-upgrade-clustermesh: Free up some disk space

michi-covalent:
helm-upgrade-clustermesh recently started running out of disk space
[^1]. Run "docker system prune -fa" to make some space before running
cilium connectivity test.

[^1]: https://github.com/cilium/cilium-cli/actions/runs/6423071562/job/17440825179

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -328,6 +328,8 @@ jobs:
 
       - name: Run the multicluster connectivity tests
         run: |
+          # Remove unused docker stuff to avoid running out of disk space.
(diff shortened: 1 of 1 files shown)


michi-covalent:
ping @christarazi @joestringer for the U.S timezone coverage 🚀🙏

christarazi (approved):
It's a bit strange why this is happening all of a sudden and needing the `docker system prune` line in just this workflow file is not ideal either, but approving to unblock in the case that the issue is a temporary blip on GH's side.

PR #2025, opened 2023-10-09
sysdump: Detect if Gateway API support is enabled

michi-covalent:
Check if "enable-gateway-api" is set to "true" in cilium-config ConfigMap, and collect resources related to Gateway API if and only if the feature is enabled.

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -1274,126 +1274,6 @@ func (c *Collector) Run() error {
 				return nil
 			},
 		},
-		{
-			Description: "Collecting GatewayClass entries",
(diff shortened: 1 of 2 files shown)


michi-covalent:
do we need to collect GRPCRoute / TCPRoute / UDPRoute? they are not listed in https://docs.cilium.io/en/latest/network/servicemesh/gateway-api/gateway-api/, so these tasks will always fail.

sayboras (approved):
Thanks ✅

michi-covalent:
@asauber ping!

asauber (approved):

Issue #2029, opened 2023-10-11
Sysdump fails if one agent is unhealthy


joestringer:
## Bug report

**General Information**

- Cilium CLI version: v0.15.10
- Orchestration system version in use: k8s v1.27.3
- Platform / infrastructure information: kind on LVH
- Link to relevant artifacts: cilium/cilium CI failure: https://github.com/cilium/cilium/actions/runs/6487092803/job/17616656283
- Generate and upload a system zip: No sysdump due to the CLI bug

**How to reproduce the issue**

1. Break Cilium in a way that causes a Cilium pod to be unavailable
2. `cilium sysdump`

**Output example from test**
```
    /¯¯\
 /¯¯\__/¯¯\    Cilium:             2 errors
 \__/¯¯\__/    Operator:           OK
 /¯¯\__/¯¯\    Envoy DaemonSet:    disabled (using embedded mode)
 \__/¯¯\__/    Hubble Relay:       disabled
    \__/       ClusterMesh:        disabled

DaemonSet              cilium             Desired: 3, Ready: 2/3, Available: 2/3, Unavailable: 1/3
Deployment             cilium-operator    Desired: 1, Ready: 1/1, Available: 1/1
Containers:            cilium-operator    Running: 1
                       cilium             Running: 3
Cluster Pods:          8/8 managed by Cilium
Helm chart version:    1.15.0-dev
Image versions         cilium-operator    quay.io/cilium/operator-generic-ci:516f401ad369de9993428a28b66a4fc0d37a11[89](https://github.com/cilium/cilium/actions/runs/6487092803/job/17616656283#step:14:91): 1
                       cilium             quay.io/cilium/cilium-ci:516f401ad369de99[93](https://github.com/cilium/cilium/actions/runs/6487092803/job/17616656283#step:14:95)428a28b66a4fc0d37a1189: 3
Errors:                cilium             cilium          1 pods of DaemonSet cilium are not ready
                       cilium             cilium-8c5m4    unable to retrieve cilium status: container cilium-agent is in CrashLoopBackOff, exited with code 2:     /go/src/github.com/cilium/cilium/pkg/egressgateway/manager.
🔍 Collecting sysdump with cilium-cli version: v0.15.10, args: [sysdump --output-filename cilium-sysdump-9-final]
🔮 Detected Cilium installation in namespace "kube-system"
🔮 Detected Cilium operator in namespace "kube-system"
[...]

PR #2032, opened 2023-10-11
sysdump feature detection: Don't depend on Cilium version

michi-covalent:
Detecting Cilium version can get arbitrary complicated. Let's not even try during sysdump feature detection since sysdump doesn't depend on any version-specific features yet.

This commit separates ExtractFromConfigMap into two functions:

- ExtractFromConfigMap to detect version-independent features.
- ExtractFromVersionedConfigMap to detect version-specific features.

Then sysdump uses ExtractFromConfigMap to extract enabled features it needs to know about without having to detect Cilium version. Problem solved, at least for now...

Fixes: #2029

### DIFF
--- a/connectivity/check/features.go
+++ b/connectivity/check/features.go
@@ -278,7 +278,8 @@ func (ct *ConnectivityTest) detectFeatures(ctx context.Context) error {
 		// If unsure from which source to retrieve the information from,
 		// prefer "CiliumStatus" over "ConfigMap" over "RuntimeConfig".
 		// See the corresponding functions for more information.
-		features.ExtractFromConfigMap(ct.CiliumVersion, cm)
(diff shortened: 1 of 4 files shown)


nebril (approved):

squeed (approved):
Seems reasonable, but why does this actually fix the mentioned issue?

michi-covalent:
> Seems reasonable, but why does this actually fix the mentioned issue?

oh yeah so the original issue was caused by this line:

```
			ciliumVersion, err := c.Client.GetCiliumVersion(context.Background(), c.CiliumPods[0])
```

`GetCiliumVersion` goes inside a cilium pod and run `cilium version` to get the cilium version, so it fails if cilium is not running 🙀 

https://github.com/cilium/cilium-cli/blob/7bb30e6a068c88e1e4f60724075bfe9e596c5f05/k8s/client.go#L967-L988

Issue #2033, opened 2023-10-11
cilium connectivity test --pause-on-fail doesn't work


joestringer:
## Bug report

**General Information**

- Cilium CLI version: v0.15.10
- Orchestration system version in use (e.g. `kubectl version`, ...): N/A
- Platform / infrastructure information (e.g. AWS / Azure / GCP, image / kernel versions): N/A
- Link to relevant artifacts (policies, deployments scripts, ...): N/A
- Generate and upload a system zip: `cilium sysdump`: N/A

**How to reproduce the issue**

```
$ ~/git/cilium-cli/cilium connectivity test  --secondary-network-iface=eth1 --include-unsafe-tests --test egress-gateway --pause-on-fail
⌛ [kind-kind] Waiting for deployment cilium-test/client to become ready...
⌛ [kind-kind] Waiting for deployment cilium-test/client2 to become ready...
⌛ [kind-kind] Waiting for deployment cilium-test/echo-same-node to become ready...
⌛ [kind-kind] Waiting for deployment cilium-test/echo-other-node to become ready...
⌛ [kind-kind] Waiting for deployment cilium-test/echo-external-node to become ready...
⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/client-84bfddc76b-gxtd9 to appear...
⌛ [kind-kind] Waiting for CiliumEndpoint for pod cilium-test/client2-764b565764-k84ph to appear...
⌛ [kind-kind] Waiting for pod cilium-test/client-84bfddc76b-gxtd9 to reach DNS server on cilium-test/echo-same-node-76847c8f99-ldgbd pod...
⌛ [kind-kind] Waiting for pod cilium-test/client2-764b565764-k84ph to reach DNS server on cilium-test/echo-same-node-76847c8f99-ldgbd pod...
⌛ [kind-kind] Waiting for pod cilium-test/client-84bfddc76b-gxtd9 to reach DNS server on cilium-test/echo-other-node-76c8b47f7c-8qxcr pod...
⌛ [kind-kind] Waiting for pod cilium-test/client2-764b565764-k84ph to reach DNS server on cilium-test/echo-other-node-76c8b47f7c-8qxcr pod...
⌛ [kind-kind] Waiting for pod cilium-test/client-84bfddc76b-gxtd9 to reach default/kubernetes service...
⌛ [kind-kind] Waiting for pod cilium-test/client2-764b565764-k84ph to reach default/kubernetes service...
[...]

PR #2036, opened 2023-10-12
Fix --pause-on-fail

joestringer:
`cilium connectivity test --pause-on-fail` wasn't working before because the
individual tests would fail out of the goroutine and the wrapping logic
wouldn't properly detect and enforce the pause. Add it into the fail/fatal
commands to make sure it takes effect, even if that means blocking an
individual test goroutine on user input.

Fixes: #2033


michi-covalent:
i mildly prefer deferring the whole post-failure logic like this https://github.com/cilium/cilium-cli/commit/8000548cd89fa7de02cc12f4daf9bd2c124e0949 so that we also get printing flows and capturing a sysdump on failure 💭

joestringer:
:+1: I'm not opinionated at all. That commit doesn't work for me, but happy to try any alternative proposal.

michi-covalent:
alright let me try and figure out why it's not working for you 💭

connectivity/check/logging.go
```diff
@@ -281,13 +281,21 @@ func (t *Test) Infof(format string, a ...interface{}) {
 	t.logf(info+" "+format, a...)
 }
 
+func (t *Test) failCommon() {
+	t.failed = true
+	t.flush()
+	if t.ctx.params.PauseOnFail {
+		t.log("Pausing after action failure, press the Enter key to continue:")
+		fmt.Scanln()
```
joestringer, line 289: Incidentally this is not great, because if you do ctrl+c it won't terminate the execution until you press enter to continue.

joestringer: Considering to add the following on top to solve this issue:

```
diff --git a/connectivity/check/logging.go b/connectivity/check/logging.go
index 230f0c196345..c1f18982f11f 100644
--- a/connectivity/check/logging.go
+++ b/connectivity/check/logging.go
@@ -4,9 +4,13 @@
 package check

 import (
+       "context"
        "fmt"
        "io"
+       "os"
+       "os/signal"
        "runtime"
+       "syscall"
        "time"
 )

@@ -286,7 +290,16 @@ func (t *Test) failCommon() {
        t.flush()
        if t.ctx.params.PauseOnFail {
                t.log("Pausing after action failure, press the Enter key to continue:")
-               fmt.Scanln()
+               cont := make(chan struct{})
+               go func() {
+                       fmt.Scanln()
+                       close(cont)
+               }()
+               ctx, _ := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+               select {
+               case <-cont:
+               case <-ctx.Done():
+               }
        }
 }

```

michi-covalent: magical

joestringer:
Hmm, I think it's pausing twice with that proposal:

```
[=] Test [to-cidr-external]
  ℹ️  📜 Applying CiliumNetworkPolicy 'client-egress-to-cidr' to namespace 'cilium-test'..
  [-] Scenario [to-cidr-external/pod-to-cidr]
  [.] Action [to-cidr-external/pod-to-cidr/external-1111-0: cilium-test/client-84bfddc76b-njg5j (10.244.3.22) -> external-1111 (1.1.1.1:443)]
  Pausing after action failure, press the Enter key to continue:

  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 --retry 3 --retry-all-errors --retry-delay 3 https://1.1.1.1:443" failed: command terminated with exit code 28
  ℹ️  curl output:


  📄 No flows recorded for peer cilium-test/client-84bfddc76b-njg5j during action external-1111-0
  📄 No flows recorded for peer external-1111 during action external-1111-0
  Pausing after action failure, press the Enter key to continue:
```

joestringer:
(updated to follow your latest feedback)

This first pause is actually not great, because it's pausing before printing the debuggability steps:

```
  [.] Action [to-cidr-external/pod-to-cidr/external-1111-0: cilium-test/client-84bfddc76b-njg5j (10.244.3.22) -> external-1111 (1.1.1.1:443)]
  Pausing after action failure, press the Enter key to continue:

  ❌ command "curl -w %{local_ip}:%{local_port} -> %{remote_ip}:%{remote_port} = %{response_code} --silent --fail --show-error --output /dev/null --connect-timeout 2 --max-time 10 --retry 3 --retry-all-errors --retry-delay 3 https://1.1.1.1:443" failed: command terminated with exit code 28
  ℹ️  curl output:
```

michi-covalent:
oh yeah i think we can just delete https://github.com/cilium/cilium-cli/pull/2036/commits/ace4127abb7a86f966c64a9469c84ac5ea75ebc0#diff-e3c06da3c6d8b3d471002757d5e992af824a412431f22e7f18f5f245f35552adR217. no need to call failCommon() from there 💡

joestringer:
Hmm, unfortunately that's the one that is printing after printing the other output that's relevant to debugging the test. Deleting that one will prevent there being double pause, but it also deletes the pause at the most useful time..

joestringer:
For what it's worth, I think that this model where individual tests know about the control flow of the overall testing and run `Fatal()` to try to interrupt the test flow breaks the abstraction a bit.. a test should either fail or not and then leave it up to the CLI infrastructure to appropriately terminate and implement all the other bells and whistles (like the pause on failure, sysdump, etc.)

michi-covalent:
would it help if we swap these 2 lines => https://github.com/cilium/cilium-cli/pull/2036/files#diff-66a6f82a060f753d91cfb453548d1f41fd695abde4debdb8cc5b982b6b1260beR314-R315

joestringer:
I'm not sure about the significance of moving the flush until after the log, but that seems right to me.

michi-covalent (approved):
✨

connectivity/check/logging.go
```diff
@@ -281,41 +285,58 @@ func (t *Test) Infof(format string, a ...interface{}) {
...
+	t.failed = true
+	t.flush()
+	if t.ctx.params.PauseOnFail {
+		t.log("Pausing after action failure, press the Enter key to continue:")
+		cont := make(chan struct{})
+		go func() {
+			fmt.Scanln()
+			close(cont)
+		}()
+		ctx, _ := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
```
joestringer, line 298: Incidentally I think this should ideally check against some broader CLI-wide context that already subscribes to this signal instead of establishing a new signal context here, but the most obvious way I could think to do this is to force all `Fail()` and variations to pass the `ctx` down which is not ideal from a developer perspective.

joestringer: We _could_ store a context in an object somewhere or have a channel that's closed to signal that the test and/or its parents are being terminated.. not sure the best approach.

joestringer: Storing contexts is generally frowned upon as I understand it.

michi-covalent: > the most obvious way I could think to do this is to force all Fail() and variations to pass the ctx down

yeah i think this is the way, maybe we can just open a separate good first issue 🥰

nebril (approved):

connectivity/check/logging.go
joestringer: Filed https://github.com/cilium/cilium-cli/issues/2042

PR #2040, opened 2023-10-13
Replace --helm-set with --set

michi-covalent:
The --helm-set flag has been deprecated.

### DIFF
--- a/internal/cli/cmd/hubble.go
+++ b/internal/cli/cmd/hubble.go
@@ -93,11 +93,6 @@ func newCmdHubbleEnable() *cobra.Command {
 
 	for flagName := range hubble.FlagsToHelmOpts {
 		// TODO(aanm) Do not mark the flags has deprecated for now.
-		// msg := fmt.Sprintf("use --helm-set=%s<=value> instead", helmOpt)
-		// err := cmd.Flags().MarkDeprecated(flagName, msg)
-		// if err != nil {
-		// 	panic(err)
-		// }
 		hubble.FlagValues[flagName] = cmd.Flags().Lookup(flagName).Value
 	}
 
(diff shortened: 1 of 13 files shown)


internal/cli/cmd/hubble.go
```diff
@@ -93,11 +93,6 @@ func newCmdHubbleEnable() *cobra.Command {
 
 	for flagName := range hubble.FlagsToHelmOpts {
 		// TODO(aanm) Do not mark the flags has deprecated for now.
```
viktor-kurchenko, line 95: Do we still need this `TODO`?

viktor-kurchenko:
JFYI: There are few places with `helm-set` (not sure if they should be replaced as well):
- internal/cli/cmd/hubble.go
- internal/cli/cmd/install.go
- internal/cli/cmd/install_test.go

viktor-kurchenko (approved):
LGTM.

michi-covalent:
>    internal/cli/cmd/hubble.go
>    internal/cli/cmd/install.go
>    internal/cli/cmd/install_test.go

yeah these are actual flag definitions and tests for normalizing flags. we are not removing these flags, just yet...

internal/cli/cmd/hubble.go
michi-covalent: not sure, i just didn't want to remove somebody else's TODOs. let's see what @aanm says ✨

aanm: @michi-covalent @viktor-kurchenko if we are deprecating `--helm-set` then we should warn users about it, which is what the commented code is for. So we can remove the TODO as long we uncomment the following code.

michi-covalent: alright then i'll keep the TODO for now ✅

aanm: The TODO needs to be kept alongside the code that was removed :sweat_smile:

michi-covalent: you mean the code that was commented out? sure i can put it back

thorn3r (approved):
amazing

PR #2035, opened 2023-10-12
Make encryption-related connectivity tests more reliable

giorio94:
Please refer to the individual commit messages for additional details.

PR #2045, opened 2023-10-16
gha: cleanup stale AWS-CNI iptables rules in EKS tunnel workflow

giorio94:
Extend the eks-tunnel script to additionally flush the stale AWS-CNI iptables rules left by the aws-node DaemonSet, which break host to pod connectivity. Using the cleanup commands from the documentation [1].

[1]: https://github.com/cilium/cilium/blob/cc2051eec0ba5e57842d8a7d4b4b33f31b12509f/Documentation/installation/k8s-install-helm.rst?plain=1#L208-L211

For https://github.com/cilium/cilium-cli/pull/2035.

### DIFF
--- a/.github/in-cluster-test-scripts/eks-tunnel.sh
+++ b/.github/in-cluster-test-scripts/eks-tunnel.sh
@@ -25,6 +25,19 @@ cilium status --wait
 # Make sure the 'aws-node' DaemonSet exists but has no scheduled pods
 [[ $(kubectl -n kube-system get ds/aws-node -o jsonpath='{.status.currentNumberScheduled}') == 0 ]]
 
+# Clean up stale AWS-CNI iptables rules, which break host to pod connectivity (cilium/cilium#25804)
+CLEANUP_CMD="\
+    iptables -t nat -F AWS-SNAT-CHAIN-0 && \
+    iptables -t nat -F AWS-SNAT-CHAIN-1 && \
+    iptables -t nat -F AWS-CONNMARK-CHAIN-0 && \
+    iptables -t nat -F AWS-CONNMARK-CHAIN-1"
+kubectl get pod \
+    -n kube-system \
+    -l app.kubernetes.io/name=cilium-agent \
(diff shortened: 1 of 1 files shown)


giorio94:
Link to successful run: https://github.com/cilium/cilium-cli/actions/runs/6529848997/job/17728137356?pr=2045
Dropping the temporary commit.

pchaigno (approved):
I thought we already did that :thinking:

nebril:
This should be done by this post-start hook: https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/files/agent/poststart-eni.bash

Should we update this script to also remove `AWS-CONNMARK-CHAIN` rules?

giorio94:
> This should be done by this post-start cilium Daemonset hook: https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/files/agent/poststart-eni.bash
> 
> Should we update this script to also remove `AWS-CONNMARK-CHAIN` rules?
> 
> @giorio94 ^^

I don't know the details of the poststart-eni script, but makes definitely sense to me (it is very easy to miss the documentation and hit this issue). I edited this workflow because from a quick look at https://github.com/cilium/cilium/issues/25804 it seemed that the script was not always executed.

Even if we extend the `poststart-eni.bash` script, I guess we'll still need this change for the moment, until a new release is cut.

giorio94:
> it seemed that the script was not always executed.

Just double-checked, and it is not run when running on EKS in tunnel mode, because `eni.mode` is not set.

/cc @nebril

nebril (approved):

giorio94:
AKS failure is unrelated, marking ready to merge

joestringer:
This seems like it's fixing CI, but isn't the same problem true in production systems?

giorio94:
> This seems like it's fixing CI, but isn't the same problem true in production systems?

Unfortunately yes. More info in https://github.com/cilium/cilium/issues/25804

PR #2047, opened 2023-10-16
echo-external-node: Check if NodeWithoutCilium is enabled

michi-covalent:
Create echo-external-node deployment if and only if there is a node with "cilium.io/no-schedule=true" label. Otherwise Kubernetes cannot schedule this deployment.

Fixes: d541be5e7ce5 ("connectivity: add echo-external deployment")
Fixes: cilium/cilium#27009

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -953,34 +953,38 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 			}
 		}
 
-		_, err = ct.clients.src.GetDeployment(ctx, ct.params.TestNamespace, echoExternalNodeDeploymentName, metav1.GetOptions{})
-		if err != nil {
-			ct.Logf("✨ [%s] Deploying echo-external-node deployment...", ct.clients.src.ClusterName())
-			containerPort := 8080
-			echoExternalDeployment := newDeployment(deploymentParameters{
-				Name:           echoExternalNodeDeploymentName,
-				Kind:           kindEchoExternalNodeName,
-				Port:           containerPort,
-				NamedPort:      "http-8080",
-				HostPort:       8080,
-				Image:          ct.params.JSONMockImage,
-				Labels:         map[string]string{"external": "echo"},
-				Annotations:    ct.params.DeploymentAnnotations.Match(echoExternalNodeDeploymentName),
-				NodeSelector:   map[string]string{"cilium.io/no-schedule": "true"},
-				ReadinessProbe: newLocalReadinessProbe(containerPort, "/"),
-				HostNetwork:    true,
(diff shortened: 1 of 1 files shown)


joestringer:
Does this automatically disable the tests that rely on this deployment?

michi-covalent:
> Does this automatically disable the tests that rely on this deployment?

yeah, it's a bit implicit but it checks if `ExternalEchoPods` is empty:

- https://github.com/cilium/cilium-cli/blob/951eb9108277f4ffeb86d047bd492b6b96a52542/connectivity/tests/egressgateway.go#L237
- https://github.com/cilium/cilium-cli/blob/951eb9108277f4ffeb86d047bd492b6b96a52542/connectivity/tests/egressgateway.go#L253
- https://github.com/cilium/cilium-cli/blob/951eb9108277f4ffeb86d047bd492b6b96a52542/connectivity/tests/egressgateway.go#L278
- https://github.com/cilium/cilium-cli/blob/951eb9108277f4ffeb86d047bd492b6b96a52542/connectivity/tests/egressgateway.go#L369

so these tests won't run if there is no `echo-external-node` pod.

connectivity/check/deployment.go
```diff
@@ -953,33 +953,35 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
-				Annotations:    ct.params.DeploymentAnnotations.Match(echoExternalNodeDeploymentName),
-				NodeSelector:   map[string]string{"cilium.io/no-schedule": "true"},
-				ReadinessProbe: newLocalReadinessProbe(containerPort, "/"),
-				HostNetwork:    true,
-				Tolerations: []corev1.Toleration{
-					{Operator: corev1.TolerationOpExists},
-				},
-			})
-			_, err = ct.clients.src.CreateServiceAccount(ctx, ct.params.TestNamespace, k8s.NewServiceAccount(echoExternalNodeDeploymentName), metav1.CreateOptions{})
+		if ct.Features[features.NodeWithoutCilium].Enabled {
```
nebril, line 956: Can we get a log line in `else` block that will indicate that we will not run these tests? I feel like this would save headache of whoever mistakenly tries to run these tests on a misconfigured cluster and won't know why they are not running.

nebril (changes requested):
Looks good, one quality of life comment left inline.

connectivity/check/deployment.go
michi-covalent: yeah sure any specific phrasing for the log message? maybe something like:
 
```
🔮 Skipping tests that require a node without Cilium
```
?

michi-covalent: ok added an info log:
```
ℹ️  Skipping tests that require a node Without Cilium
```

PR #2048, opened 2023-10-16
fix: properly propagate flags used for kubeproxyreplacement

bittermandel:
After the switch to Helm install & upgrade mode rather than classic mode, the flags for kubeProxyReplacement, k8sServiceHost & k8sServicePort was not propagated properly.

These changes makes sure that there's no auto detection done if the `kubeProxyReplacement` flag has been set, and if it's not set, make sure to prioritze the k8sServiceHost / k8sServicePort flags over getting it from the current context.

```release-note
fix: properly propagate helm values used for kubeProxyReplacement
```

### DIFF
--- a/install/autodetect.go
+++ b/install/autodetect.go
@@ -198,32 +198,50 @@ func (k *K8sInstaller) autodetectAndValidate(ctx context.Context, helmValues map
 		return fmt.Errorf("invalid encryption mode")
 	}
 
-	k.autodetectKubeProxy(ctx)
+	apiServerHost, apiServerPort, err := k.autodetectKubeProxy(ctx)
+	if err != nil {
+		return err
+	}
+
+	if host, ok := helmValues["k8sServiceHost"].(string); ok {
+		apiServerHost = host
+	}
+	if apiServerHost != "" {
+		k.params.HelmOpts.Values = append(k.params.HelmOpts.Values, fmt.Sprintf("k8sServiceHost=%s", apiServerHost))
+	}
+
+	if port, ok := helmValues["k8sServicePort"].(string); ok {
+		apiServerPort = port
+	}
+	if apiServerPort != "" {
+		k.params.HelmOpts.Values = append(k.params.HelmOpts.Values, fmt.Sprintf("k8sServicePort=%s", apiServerPort))
+	}
+
+	if k.params.UserSetKubeProxyReplacement && (apiServerHost == "" || apiServerPort == "") {
+		k.Log("❌ Could not autodetect API server endpoint. Either set a valid kubernetes context or set k8sServerHost & k8sServerPort.")
+		return fmt.Errorf("could not autodetect API server endpoint, cannot replace kubeproxy")
+	}
 	return k.autoEnableBPFMasq()
 }
 
-func (k *K8sInstaller) autodetectKubeProxy(ctx context.Context) error {
-	if k.params.UserSetKubeProxyReplacement {
-		return nil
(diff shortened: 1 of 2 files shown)


install/autodetect.go
```diff
@@ -223,7 +232,7 @@ func (k *K8sInstaller) autodetectKubeProxy(ctx context.Context) error {
 			return nil
 		}
 	}
-	apiServerHost, apiServerPort := k.client.GetAPIServerHostAndPort()
```
squeed, line 226: Hmm, why is it safe to remove this line? I agree that we must prefer user-provided inputs over autodetection, but why remove autodetection entirely?

bittermandel: I've read through it a couple of more times and it's not safe. Also I don't think the flow of the autodetect stuff makes much sense with these changes. I will check tomorrow how it can be improved, regardless setting the helm values in the end of autodetect feels wrong.

bittermandel:
@squeed with the changes from the last commit, the following flow is enabled:
* If kube-proxy is installed, no autodetection of apiserver endpoints is needed and helm values are set as expected.
* If kube-proxy is not installed or replacement flag is set:
  * kubeProxyReplacement is set to true as expected. 
  * k8sServiceHost and k8sServicePort is set either from the current kubernetes context or --set flags.
  * if both values could not be determined, return an error and log.

This flow makes sense to me and is compatible with previous versions.

squeed (approved):
looks great, thanks!

asauber (changes requested):
@bittermandel, thanks for the PR.

You have some gofmt and revive errors in your latest commit. See the "Go / build (pull_request)" workflow logs. Would you be able to resolve those?

Could you also squash your latest changes into the earlier commits?

In addition, could you reformat your commit messages with a shorter summary line and some additional description of the changes? (See this PR for a good example of our commit message standards https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)

bittermandel:
@asauber I made sure to squash the commits, put a more descriptive commit message and a description!
I am not sure why these CI tests fail though. Doesn't seem related to my changes?

asauber (commented):
Re-running CI and taking another look

internal/cli/cmd/install.go
```diff
@@ -266,6 +266,11 @@ cilium install
 cilium install --context kind-cluster1 --set cluster.id=1 --set cluster.name=cluster1
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.Flags().Visit(func(f *pflag.Flag) {
+				if f.Name == "set" && strings.Contains(f.Value.String(), "kubeProxyReplacement") {
```
asauber, line 270: This isn't sufficient as a test for `kubeProxyReplacement`, as the value could be `kubeProxyReplacement=false`. In this case we do not want to set `params.UserSetKubeProxyReplacement = true`.

asauber (changes requested):

asauber (commented):
The kind workflow is running into an issue with the latest logic:

```
🔮 Auto-detected kube-proxy has been installed
Error: Unable to install Cilium: could not autodetect API server endpoint, cannot replace kubeproxy
❌ Could not autodetect API server endpoint. Either set a valid kubernetes context or set k8sServerHost & k8sServerPort.
```

Could you look into why this is the case?

ti-mo:
@bittermandel Looks like this is failing CI, PTAL.

ti-mo:
Looks like the author has abandoned the PR, closing for now.

PR #2052, opened 2023-10-17
sysdump: Add "serial" tasks support to enable trace data collection

pippolo84:
- Add support for "serial" tasks in the sysdump subcommand.  Serial tasks are the ones that cannot be executed concurrently with other tasks.
- Add the collection of the data from the Go execution tracer as a "serial" sysdump task.

See the individual commit messages for further details.

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -66,6 +66,8 @@ type Options struct {
 	Debug bool
 	// Whether to enable scraping profiling data.
 	Profiling bool
+	// Whether to enable scraping tracing data.
+	Tracing bool
 	// The labels used to target additional pods
 	ExtraLabelSelectors []string
 	// The labels used to target Hubble pods.
@@ -1263,6 +1265,21 @@ func (c *Collector) Run() error {
 
 	if c.Options.CiliumNamespace != "" && c.Options.CiliumOperatorNamespace != "" {
 		tasks = append(tasks, ciliumTasks...)
+
+		serialTasks = append(serialTasks, Task{
+			CreatesSubtasks: true,
+			Description:     "Collecting tracing data from Cilium pods",
+			Quick:           false,
+			Task: func(ctx context.Context) error {
+				if !c.Options.Tracing {
+					return nil
+				}
+				if err := c.SubmitTracingGopsSubtask(c.CiliumPods, ciliumAgentContainerName); err != nil {
+					return fmt.Errorf("failed to collect tracing data from Cilium pods: %w", err)
+				}
+				return nil
+			},
+		})
 	}
 
 	tetragonTasks := []Task{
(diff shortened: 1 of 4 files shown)


pippolo84:
On a second thought, it'd be better to collect the trace data while all the other sysdump tasks have finished. This avoids polluting the tracing data from the agent. I'm gonna rework the PR to run the tracing task in a workerpool of its own. Moving back to draft for now.

sysdump/sysdump.go
```diff
@@ -1448,6 +1452,53 @@ func (c *Collector) Run() error {
...
+			}
+			c.logTask(t.Description)
+			defer c.logDebug("Finished %q", t.Description)
+			return t.Task(ctx)
+		}); err != nil {
+			return fmt.Errorf("failed to submit task to the worker pool: %w", err)
+		}
+
+		// Wait for the subtasks to be submitted and then call 'Drain' to wait for them to finish.
+		wg.Wait()
```
asauber, line 1506: It's a bit sad that the worker pool requires a separate WG in order to operate properly. Can't we treat submissions to the Pool as blocking?

I see that this is the pattern that we use elsewhere. Seems fine.

sysdump/sysdump.go
```diff
@@ -1448,6 +1452,53 @@ func (c *Collector) Run() error {
 		return fmt.Errorf("failed to close the worker pool: %w", err)
 	}
 
+	// Run each serial task in its own workerpool.
+	for i, t := range serialTasks {
```
asauber, line 1473: Should we execute the serialTasks before any other type?

sysdump/sysdump.go
```diff
@@ -2116,6 +2133,44 @@ func (c *Collector) SubmitProfilingGopsSubtasks(pods []*corev1.Pod, containerNam
...
+func (c *Collector) SubmitTracingGopsSubtask(pods []*corev1.Pod, containerName string) error {
+	for _, p := range pods {
+		p := p
+		if err := c.Pool.Submit(fmt.Sprintf("gops-%s-%s", p.Name, gopsTrace), func(ctx context.Context) error {
+			agentPID, err := c.getGopsPID(ctx, p, containerName)
+			if err != nil {
+				return err
+			}
+			o, err := c.Client.ExecInPod(ctx, p.Namespace, p.Name, containerName, []string{
+				gopsCommand,
```
asauber, line 2146: I think this would be more readable as

```
gopsPath,    // replaces "Command"
gopsCommand, // replaces "Trace"
```

asauber (changes requested):

sysdump/sysdump.go
```diff
@@ -26,6 +25,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
+	"github.com/cilium/cilium/pkg/versioncheck"
```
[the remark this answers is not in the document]
pippolo84: I think it should stay in the previous section because it doesn't have a prefix "github.com/cilium/cilium" (in fact, workerpool lives in its own [repository](https://github.com/cilium/workerpool)).
At least this is the default settings we use for organizing imports in cilium (my IDE is automatically moving it). Are we are using something different in cilium-cli?

sysdump/sysdump.go
pippolo84: I'm afraid we need this because of the subtasks. Each task in `serialTasks` is executed asynchronously, and they too might submit a new subtask to the same pool. (In fact, we are doing this in the `gops trace` case, inserting a subtask for each agent to trace).
But calling `Drain` prevents subtasks to do that:

```
// Drain waits until all tasks are completed. This operation prevents
// submitting new tasks to the worker pool. Drain returns the results of the
// tasks that have been processed.
```

So, without the wg this could happen:

```
1) submit main task
2) call Drain -> wait for main task to finish
3) execute the task
  3a) `c.Pool.Submit` in `SubmitTracingGopsSubtask` is gonna fail because `wp` is in draining mode
```

I agree with you that this is a bit cumbersome, probably we could probably use another method in the workerpool to wait for tasks termination without entering draining mode.
I see that the same pattern has been used for previous `tasks` too.

sysdump/sysdump.go
pippolo84: Might be a good idea in order to avoid triggering the Go runtime memory preallocation from completing the other tasks. I'll move them! :+1:

sysdump/sysdump.go
pippolo84: Actually, we have other gops sub-commands like `gopsProfiling`. An alternative would be to just swap `gopsCommand` for `gopsPath`, but I'm not sure this is truly better, as gops is used also as a command itself to detect the PID: https://github.com/cilium/cilium-cli/blob/0245001fd6fd9c907780ef0b78547caf118c72eb/sysdump/sysdump.go#L2046-L2050

WDYT?

sysdump/sysdump.go
asauber: Interleaving the calls to `Submit` seems like a similarly poor API to me. I agree that this is the established pattern and it's certainly fine for now.

sysdump/sysdump.go
asauber: Given the different named, subcommands, I think "Command" is fine. Resolving :)

sysdump/sysdump.go
```diff
@@ -1404,6 +1421,54 @@ func (c *Collector) Run() error {
...
+			return fmt.Errorf("failed to drain the serial tasks worker pool: %w", err)
+		}
+		// Close the worker Pool.
+		if err := c.Pool.Close(); err != nil {
+			return fmt.Errorf("failed to close the serial tasks worker pool: %w", err)
+		}
+
+		r = append(r, results...)
+	}
+
```
tklauser, line 1471: It looks like the first commit is introducing this bit of the code and the second commit is moving it. Should it be added in this place already by the first commit to avoid churn in the second commit?

sysdump/sysdump.go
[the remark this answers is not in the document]
tklauser: Nit: in cilium/cilium and cilium/cilium-cli we AFAIK only use three import groups:

1. standard libary
2. other 3rd party imports
3. imports from the same module

Following the above, I think the `github.com/cilium/cilium/pkg/versioncheck` import should stay where it currently is because that package is a 3rd party import from the perspective of this module (`github.com/cilium/cilium-cli/sysdump`)

sysdump/sysdump.go
pippolo84: That's right, I've inadvertently messed up the commits with interactive rebase. Fixing it!

sysdump/sysdump.go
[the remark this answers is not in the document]
pippolo84: Might be some misconfig in my IDE, then. I'll move it back there :+1:

tklauser (approved):
Thanks

PR #2054, opened 2023-10-17
sysdump: Check if enable-envoy-config is set

michi-covalent:
Collect CiliumClusterwideEnvoyConfigs and CiliumEnvoyConfigs custom resources if and only if enable-envoy-config is set to "true".


michi-covalent:
@asauber ping!

sayboras (approved):

PR #2056, opened 2023-10-18
go.mod: Let Renovate manage Go version

michi-covalent:
Configure Renovate to update Go version in go.mod, and modify workflow files to use go.mod to get the Go version to reduce the number of places Renovate needs to update.

Ref: https://github.com/cilium/tetragon/pull/1579

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,8 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
         with:
-          # renovate: datasource=golang-version depName=go
-          go-version: 1.21.3
+          go-version-file: 'go.mod'
 
       - name: Generate the artifacts
         run: make release
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,6 @@
 module github.com/cilium/cilium-cli
 
+// renovate: datasource=golang-version depName=go
 go 1.21.1
 
 // Replace directives from github.com/cilium/cilium. Keep in sync when updating Cilium!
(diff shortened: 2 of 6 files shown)


nbusseneau (approved):

go.mod
```diff
@@ -1,5 +1,6 @@
 module github.com/cilium/cilium-cli
 
+// renovate: datasource=golang-version depName=go
 go 1.21.1
```
rolinh, line 4: We may not want renovate to bump the patch version in the go.mod, see https://github.com/cilium/cilium/pull/28686

michi-covalent: oh interesting could we just specify `go 1.21` here?

michi-covalent: anyways i'll just wait for https://github.com/cilium/cilium/pull/28686 to get merged and do the same thing. converting to draft.

rolinh: I think that you can specify `go 1.21` as long as you add a toolchain directive. Maybe that's what we should be doing, ie:
```
go 1.21

// renovate: datasource=golang-version depName=go
toolchain 1.21.3
```

michi-covalent: fascinating

aanm: cc @joestringer :point_up:

.github/workflows/release.yaml
```diff
@@ -18,8 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
         with:
-          # renovate: datasource=golang-version depName=go
-          go-version: 1.21.3
+          go-version-file: 'go.mod'
```
joestringer, line 21: I would suggest *not* to tie release workflows to `go.mod` version. Release workflows should always use the latest version, whereas I think that developer toolchains can lag behind within the same minor version and that's OK.

joestringer: In practice that means that the approach I'm exploring in the other linked PR is to have two renovate configs: One for go.mod following the `X.Y.0` releases and a different one for everything else following every `X.Y.Z` patch release.

joestringer: The actual implementation ended up different: There's a general Go version bump rule, then I added a specific rule for `go.mod` to disable patch updates for the toolchain version in the go.mod file.

michi-covalent:
closing this for now. i guess we could do something similar to https://github.com/cilium/cilium/pull/28686 💭

PR #2059, opened 2023-10-19
Add SPIRE server entries to sysdump

meyskens:
This adds an export of the SPIRE server entries to the sysdump. This can be used to detect any initial Cilium access entry issues as well as any sync issues with the operator.

### DIFF
--- a/sysdump/constants.go
+++ b/sysdump/constants.go
@@ -33,6 +33,7 @@ const (
 	hubbleRelayContainerName           = defaults.RelayContainerName
 	hubbleRelayDeploymentName          = defaults.RelayDeploymentName
 	hubbleUIDeploymentName             = defaults.HubbleUIDeploymentName
+	spireServerContainerName           = "spire-server"
 	redacted                           = "XXXXXX"
 )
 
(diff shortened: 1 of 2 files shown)


sysdump/constants.go
```diff
@@ -34,6 +34,7 @@ const (
 	hubbleRelayDeploymentName          = defaults.RelayDeploymentName
 	hubbleUIDeploymentName             = defaults.HubbleUIDeploymentName
 	redacted                           = "XXXXXX"
+	spireServerContainerName           = "spire-server"
```
mhofstetter, line 37: nit: would keep `redacted` at the end

sysdump/constants.go
```diff
@@ -51,6 +52,7 @@ const (
 	ciliumSPIREAgentConfigMapFileName        = "cilium-spire-agent-configmap-<ts>.yaml"
 	ciliumSPIREServerStatefulSetFileName     = "cilium-spire-server-statefulset-<ts>.yaml"
 	ciliumSPIREServerConfigMapFileName       = "cilium-spire-server-configmap-<ts>.yaml"
+	ciliumSPIREServerEntriesFileName         = "cilium-spire-server-entries-%s-<ts>.log"
```
mhofstetter, line 55: file ending `log` seems not to fit here.

what do you think of changing the output format to json (by adding `--output json` to the command) and changing the file ending accordingly? Should also be easier to process and analyze the file.

mhofstetter (commented):
Thanks @meyskens 

Looks good, only some comments inline.

sysdump/constants.go
meyskens: Good one! Didn't think it would have JSON output, I changed it!

mhofstetter (approved):

Issue #2070, opened 2023-10-27
CI: external workloads workflow consistently fails in "Verify DNS on VM" step


tklauser:
The [external workloads workflow](https://github.com/cilium/cilium-cli/actions/workflows/externalworkloads.yaml) seems to be failing consistently:

https://github.com/cilium/cilium-cli/actions/runs/6622170504

It looks like `nslookup` is timing out trying to reach the DNS server:

```
;; connection timed out; no servers could be reached
```

This seems to have started around 2023-10-21.

* First scheduled run that failed: https://github.com/cilium/cilium-cli/actions/runs/6593096759
* Last scheduled run that passed: https://github.com/cilium/cilium-cli/actions/runs/6590038993

nbusseneau:
External workfloads on main Cilium repo is working, perhaps something was changed in there that should have been changed on CLI as well but did not?

michi-covalent:
this might be relevant. from the sysdump in https://github.com/cilium/cilium-cli/actions/runs/6725580157 [cilium-sysdump-out.zip.zip](https://github.com/cilium/cilium-cli/files/13233460/cilium-sysdump-out.zip.zip)
```
% grep 'level=error' ./logs-clustermesh-apiserver-694765f445-9z5d6-apiserver-20231101-221630.log
2023-11-01T22:13:59.507218421Z level=error msg="CEW: Invalid identity 1 in &{{cilium-cilium-cli-6725580157-classic-vm1 cilium-cilium-cli-6725580157-classic-vm1 [{InternalIP 10.168.0.5} {InternalIP fc00::10ca:1} {CiliumInternalIP 10.192.1.2} {CiliumInternalIP f00d::a05:0:0:9f6f}] 10.192.1.0/30 [] f00d::a05:0:0:0/96 [] <nil> <nil> <nil> <nil> 0 local 0 map[io.cilium.k8s.policy.cluster:cilium-cilium-cli-6725580157-classic-vm1 io.kubernetes.pod.name:cilium-cilium-cli-6725580157-classic-vm1 io.kubernetes.pod.namespace:default] map[] 1 }}" subsys=clustermesh-apiserver
```

giorio94:
> this might be relevant. from the sysdump in https://github.com/cilium/cilium-cli/actions/runs/6725580157 [cilium-sysdump-out.zip.zip](https://github.com/cilium/cilium-cli/files/13233460/cilium-sysdump-out.zip.zip)
> 
> ```
[...]

PR #2073, opened 2023-11-01
connectivity: don't use an ephemeral port for the echo server

jibi:
as otherwise replies to traffic SNATed with that same source port will get rejected by the HostPort service


brb (approved):
Thanks! As a follow-up, we could add a sanity check.

tklauser (approved):

tklauser:
External workloads workflow is failing due to #2070. All other workflows passed, merging.

PR #2094, opened 2023-11-07
Connectivity Test: Add latency measurement

darox:


learnitall (commented):
One suggestion here is to add the ability to see percentile results in addition to the average. Having 50, 90, and 99th percentiles can be really helpful when trying to understand the average.

darox:
> One suggestion here is to add the ability to see percentile results in addition to the average. Having 50, 90, and 99th percentiles can be really helpful when trying to understand the average.

Implemented

connectivity/tests/perfpod.go
```diff
@@ -69,48 +73,134 @@ func (s *netPerfPodtoPod) Run(ctx context.Context, t *check.Test) {
...
+			if err != nil {
+				a.Fatal("Unable to parse netperf result")
+			}
+			latency["p90Latency"] = append(latency["p90Latency"], p90Latency)
+
+			p99Latency, err := strconv.ParseFloat(d[6], 64)
+			if err != nil {
+				a.Fatal("Unable to parse netperf result")
+			}
+			latency["p99Latency"] = append(latency["p99Latency"], p99Latency)
```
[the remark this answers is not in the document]
darox: I agree. This is much cleaner. Thank you

derailed (commented):
@darox Nice work!

darox:
> @darox Nice work!

Thank you. I implemented all of your proposals.

markpash (approved):

christarazi (approved):
LGTM. It might make sense at some point to consider collapsing all of the `--perf-*` flags into one `--perf` flag that accepts enums such as `--perf=latency` or `--perf=net`. The former runs latency tests and the latter only runs the base "network" tests. We could have `--perf=all` to run all perf-related tests. This would reduce the need to keep adding more and more flags. Not required for this PR though.

darox:
> LGTM. It might make sense at some point to consider collapsing all of the `--perf-*` flags into one `--perf` flag that accepts enums such as `--perf=latency` or `--perf=net`. The former runs latency tests and the latter only runs the base "network" tests. We could have `--perf=all` to run all perf-related tests. This would reduce the need to keep adding more and more flags. Not required for this PR though.

Thank you for the feedback. Absolutely, this is something I wanna work on after this PR is merged.

connectivity/check/context.go
```diff
@@ -553,16 +554,36 @@ func (ct *ConnectivityTest) report() error {
...
-		ct.Logf("📋 %-15s | %-50s | %-15s | %-15s | %-15s | %-15s", "Scenario", "Pod", "Test", "Num Samples", "Duration", "Avg value")
-		ct.Logf("%s", strings.Repeat("-", 145))
-		for p, d := range ct.PerfResults {
-			ct.Logf("📋 %-15s | %-50s | %-15s | %-15d | %-15s | %.2f (%s)", d.Scenario, p.Pod, p.Test, d.Samples, d.Duration, d.Avg, d.Metric)
-			ct.Debugf("Individual Values from run : %s", d.Values)
+
+		ct.Headerf("🔥 Latency Test Summary: ")
+
+		// Report Performance results for latency
+		if ct.params.PerfLatency {
```
tklauser, line 561: nit: don't we only want to print that header in case of latency tests and `ct.Headerf("🔥 Performance Test Summary: ")` in the `else` case below (see other review remark)?

```suggestion
		// Report Performance results for latency
		if ct.params.PerfLatency {
			ct.Headerf("🔥 Latency Test Summary: ")
```

connectivity/check/context.go
```diff
@@ -553,16 +554,36 @@ func (ct *ConnectivityTest) report() error {
...
+					d.Scenario, p.Pod, p.Test, d.Samples, d.Duration,
+					d.Latency["min"][0], d.Metric,
+					d.Latency["mean"][0], d.Metric,
+					d.Latency["max"][0], d.Metric,
+					d.Latency["p50"][0], d.Metric,
+					d.Latency["p90"][0], d.Metric,
+					d.Latency["p99"][0], d.Metric)
+			}
+			ct.Logf("%s", strings.Repeat("-", 233))
+		} else {
```
tklauser, line 576: per the comment above:

```suggestion
		} else {
			ct.Headerf("🔥 Performance Test Summary: ")
```

darox:
> LGTM. One non-blocking nit regarding summary headers for latency tests vs. existing performance tests. Can be addressed as a follow-up. Thanks.

Awesome catch. I didn't see that.

PR #2096, opened 2023-11-08
connectivity: Add GetTestOrDie()

michi-covalent:
Add a convenient function GetTestOrDie(). There isn't much you can do if GetTest() fails anyways.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -293,6 +293,16 @@ func (ct *ConnectivityTest) GetTest(name string) (*Test, error) {
 	panic("missing test descriptor for a registered name")
 }
 
+// GetTestOrDie returns the test scope for test named "name" if found,
+// or panics otherwise.
+func (ct *ConnectivityTest) GetTestOrDie(name string) *Test {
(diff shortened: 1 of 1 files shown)


christarazi (approved):

connectivity/check/context.go
```diff
@@ -293,6 +293,16 @@ func (ct *ConnectivityTest) GetTest(name string) (*Test, error) {
 	panic("missing test descriptor for a registered name")
 }
 
+// GetTestOrDie returns the test scope for test named "name" if found,
+// or panics otherwise.
+func (ct *ConnectivityTest) GetTestOrDie(name string) *Test {
```
tklauser, line 298: nit: the upstream convention for functions that panic on failure is usually `MustFoo` (see e.g. [`netip.MustParseAddr`](https://pkg.go.dev/net/netip#MustParseAddr)), so this could be named `MustGetTest` instead.

michi-covalent: ✨ https://github.com/cilium/cilium-cli/pull/2098 ✨

PR #2097, opened 2023-11-09
Make --nodes-without-cilium work with SPIRE

meyskens:
This change sets the same affinity on the SPIRE agent as CIlium has when` --nodes-without-cilium` is used, this prevents it from being scheduled where Cilium agent is not present.


squeed:
I suggested using PodAffinity in https://github.com/cilium/cilium/pull/28947; that should solve this without a CLI change needed.

meyskens:
Sadly that will not work see https://github.com/cilium/cilium/pull/28947#issuecomment-1823069364

squeed (approved):

PR #2111, opened 2023-11-15
Connectivity tests counter.

viktor-kurchenko:
Connectivity tests counter added to make it more user friendly.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -402,10 +402,10 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 	}
 
 	// Newline denoting start of test output.
-	ct.Log("🏃 Running tests...")
+	ct.Logf("🏃 Running %d tests ...", len(ct.tests))
 
 	// Execute all tests in the order they were registered by the test suite.
-	for _, t := range ct.tests {
+	for i, t := range ct.tests {
 		if err := ctx.Err(); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


tklauser (commented):
Should we report test counter when skipping a test as well? i.e. in the `t.Context().skip` and `t.skip` calls?

viktor-kurchenko:
> Should we report test counter when skipping a test as well? i.e. in the `t.Context().skip` and `t.skip` calls?

I thought about it and I didn't do that because skip action doesn't consume time (happens immediately).
However from a consistency point of view maybe it's better to add a counter.

tklauser:
> > Should we report test counter when skipping a test as well? i.e. in the `t.Context().skip` and `t.skip` calls?
> 
> I thought about it and I didn't do that because skip action doesn't consume time (happens immediately). However from a consistency point of view maybe it's better to add a counter.

Yeah, I think it would be better for consistency.

viktor-kurchenko:
> > > Should we report test counter when skipping a test as well? i.e. in the `t.Context().skip` and `t.skip` calls?
> > 
> > 
> > I thought about it and I didn't do that because skip action doesn't consume time (happens immediately). However from a consistency point of view maybe it's better to add a counter.
> 
> Yeah, I think it would be better for consistency.

Done.

tklauser (approved):
Thanks

nebril (approved):

PR #2126, opened 2023-11-24
connectivity: Add more tests for Ingress Controller

sayboras:
This commit is to cover the cases which the traffic is sent via external node client (i.e. from node without Cilium) to Ingress service.

### DIFF
--- a/connectivity/tests/service.go
+++ b/connectivity/tests/service.go
@@ -297,3 +297,35 @@ func (s *outsideToNodePort) Run(ctx context.Context, t *check.Test) {
 		}
 	}
 }
+
+// OutsideToIngressService sends an HTTP request from client pod running on a node w/o
+// Cilium to NodePort services.
+func OutsideToIngressService() check.Scenario {
+	return &outsideToIngressService{}
+}
+
+type outsideToIngressService struct{}
+
+func (s *outsideToIngressService) Name() string {
+	return "outside-to-ingress-service"
+}
+
+func (s *outsideToIngressService) Run(ctx context.Context, t *check.Test) {
+	clientPod := t.Context().HostNetNSPodsByNode()[t.NodesWithoutCilium()[0]]
+	i := 0
+
+	for _, svc := range t.Context().IngressService() {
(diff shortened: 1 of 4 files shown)


connectivity/check/deployment.go
```diff
@@ -32,9 +32,10 @@ const (
 
 	perfHostNetNamingSuffix = "-host-net"
 
-	clientDeploymentName  = "client"
-	client2DeploymentName = "client2"
-	clientCPDeployment    = "client-cp"
+	clientDeploymentName         = "client"
+	client2DeploymentName        = "client2"
+	clientCPDeployment           = "client-cp"
+	clientExternalNodeDeployment = "client-external-node"
```
brb, line 38: Do we actually need this external client? We already have a pod running on that external node. You can check the `outsideToNodePort` test case.

sayboras: ah thanks :heavy_check_mark:

brb (approved):
Thanks!

aspsk (approved):

connectivity/tests/service.go
```diff
@@ -297,3 +297,35 @@ func (s *outsideToNodePort) Run(ctx context.Context, t *check.Test) {
...
+func (s *outsideToIngressService) Name() string {
+	return "outside-to-ingress-service"
+}
+
+func (s *outsideToIngressService) Run(ctx context.Context, t *check.Test) {
+	clientPod := t.Context().HostNetNSPodsByNode()[t.NodesWithoutCilium()[0]]
+	i := 0
+
+	for _, svc := range t.Context().IngressService() {
+		t.NewAction(s, fmt.Sprintf("curl-%d", i), &clientPod, svc, features.IPFamilyAny).Run(func(a *check.Action) {
```
tklauser, line 318: Please use a slightly more descriptive name for the action, otherwise it will be hard to distinguish:

```suggestion
		t.NewAction(s, fmt.Sprintf("curl-ingress-service-%d", i), &clientPod, svc, features.IPFamilyAny).Run(func(a *check.Action) {
```

tklauser (approved):

PR #2144, opened 2023-12-04
clustermesh: fix pattern to match IPv4 address

mhofstetter:
Executing `cilium clustermesh vm install install-external-workload.sh` fails if the randomly generated IP (v4) contains a zero (`0`) in its first byte.

Error: `Malformed CLUSTER_ADDR: 104.198.85.126:2379`

Therefore, this commit changes the pattern that is used to detect an IPv4 address (with port).

Until now i don't understand why it's working with the prefix `[^[]` for all other cases but not if the first byte is containing a zero :man_shrugging: I just think that this part isn't necessary - and shouldn't break anything :pray: 

Detected on a cilium/cilium CI run: https://github.com/cilium/cilium/actions/runs/7067930793/job/19241908576


mhofstetter:
Thanks @marseel 

I think you meant `any character except for `[` ?

Now that this is clear, the reason for this might have been to prevent matching any ipv6 address (without `[..]` with this match :thinking:

michi-covalent (approved):
a very innocent regex

PR #2145, opened 2023-12-04
connectivity test: introduce connectivity test suite timeout flag

mhofstetter:
Currently, the connectivity test suite command doesn't have a timeout. Its context only gets cancelled due to interrupt of SIGTERM signal.

For most connectivity test scenarios and actions, this isn't a problem as they have individual fine-grained timeouts (e.g. request timeout).

But there are test scenarios (e.g. `health ` - `Cilium Health Probe`) that might run infinitely on failiure (e.g. if a Cilium Agent Pod no longer is available).

Therefore, this commit introduces a new flag `--timeout` that provides the possibility to configure an overall timeout for the connectivity test suite. It defaults to `0` - meaning no timeout by default (current behavior).

Alternative: Introduce timeout for health probe test (I thought it's better to solve with a global timeout - because other test scenarios / actions have the same problem)

Example of an infinite health probe attempt (until GitHub action timeout) in `cilium/cilium`: https://github.com/cilium/cilium/actions/runs/7060997016/job/19221753163 (Of course it would also be possible to only solve it on the usage side - e.g. by setting a timeout in the connectivity test step. But i thought it might be worth for other use-cases too.)

### DIFF
--- a/defaults/defaults.go
+++ b/defaults/defaults.go
@@ -3,7 +3,9 @@
 
 package defaults
 
-import "time"
+import (
+	"time"
+)
 
 const (
 	// renovate: datasource=github-releases depName=cilium/cilium
@@ -135,6 +137,9 @@ const (
 	// ClustermeshMaxConnectedClusters is the default number of the maximum
 	// number of clusters that should be allowed to connect to the Clustermesh.
 	ClustermeshMaxConnectedClusters = 255
+
+	// Default timeout for Connectivity Test Suite (disabled by default)
+	ConnectivityTestSuiteTimeout = 0 * time.Minute
 )
 
 var (
--- a/internal/cli/cmd/connectivity.go
+++ b/internal/cli/cmd/connectivity.go
@@ -44,6 +44,7 @@ var params = check.Parameters{
 		Writer:                   os.Stdout,
 	},
 }
+
 var tests []string
 
 func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
@@ -76,11 +77,17 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 				return err
 			}
 
-			ctx, _ := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+
+			if params.Timeout > 0 {
+				timeoutCtx, cancelFunc := context.WithTimeoutCause(ctx, params.Timeout, fmt.Errorf("connectivity test suite timeout (%s) reached", params.Timeout))
+				defer cancelFunc()
+				ctx = timeoutCtx
+			}
 
 			go func() {
 				<-ctx.Done()
-				cc.Log("Interrupt received, cancelling tests...")
+				cc.Logf("Cancellation request (%s) received, cancelling tests...", context.Cause(ctx))
 			}()
 
 			done := make(chan struct{})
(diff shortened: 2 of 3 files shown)


internal/cli/cmd/connectivity.go
```diff
@@ -76,11 +77,17 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
...
+			if params.Timeout > 0 {
+				timeoutCtx, cancelCtx := context.WithTimeoutCause(ctx, params.Timeout, fmt.Errorf("connectivity test suite timeout (%s) reached", params.Timeout))
+				defer cancelCtx()
+				ctx = timeoutCtx
+			}
 
 			go func() {
 				<-ctx.Done()
-				cc.Log("Interrupt received, cancelling tests...")
+				cc.Logf("Cancellation request (%s) received, cancelling tests...", context.Cause(ctx))
```
mhofstetter, line 90: In case of an interrupt, we might lose a little bit of detail in the log. This is because `signal.NotifyContext` doesn't provide the actual cause. But this should be fixed once https://github.com/golang/go/issues/60756 is fixed.

For the time being we would have to live with what `context.Cause(ctx)` provides -> `ctx.Error()`

Interrupt: `Cancellation request (context canceled) received, cancelling tests...`
Timeout: `Cancellation request (connectivity test suite timeout (5s) reached) received, cancelling tests...`

internal/cli/cmd/connectivity.go
```diff
@@ -76,11 +77,17 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 				return err
 			}
 
-			ctx, _ := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+			ctx, _ := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+
+			if params.Timeout > 0 {
+				timeoutCtx, cancelCtx := context.WithTimeoutCause(ctx, params.Timeout, fmt.Errorf("connectivity test suite timeout (%s) reached", params.Timeout))
```
derailed, line 83: nit: Suffix confusing. cancelCtx is not a context but a CancelFunc.

derailed (approved):
@mhofstetter LGTM

internal/cli/cmd/connectivity.go
mhofstetter: renamed - thanks @derailed

christarazi (commented):
I agree with the difficulty in selecting a default timeout value because most of these tests run with a GH Actions timeout anyway. With that said, what's the point of this PR if the default is going to be 0 anyway? When is it intended to set the timeout to a non-zero value?

defaults/defaults.go
```diff
@@ -135,6 +137,9 @@ const (
 	// ClustermeshMaxConnectedClusters is the default number of the maximum
 	// number of clusters that should be allowed to connect to the Clustermesh.
 	ClustermeshMaxConnectedClusters = 255
+
+	// Default timeout for Connectivity Test Suite (disabled by default)
```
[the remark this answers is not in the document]
tklauser: I think it's fine to leave the default timeout disable by default. Otherwise users outside of GHA might easily hit it, depending on their infrastructure.

tklauser (approved):
LGTM, I think this might benefit users running connectivity tests outside of GHA. I like that the default timeout is disabled by default and we don't change existing behavior.

christarazi (approved):
I'm OK with this change. I have a slight worry in the back of my mind that when this timeout is set, if we are not carefully enough we'll run into weird situations where it might overlap just enough with the GH Actions timeout. 

Maybe it's worth just mentioning in the help msg that this is not recommended to set in GH Actions?

PR #2146, opened 2023-12-04
connectivity test: check for deleted cilium agent pod in health probe

mhofstetter:
Currently, the health probe connectivity test fails forever if an Cilium Agent Pod that existed when starting the connectivity tests no longer exists.

The reason is that the health probes uses the list of Cilium Agent pods that got fetched at the beginning of the connectivity test run.

Therefore, this commit adds a check whether the Pod still exists. If not, the health probe check fails.

The underlying reason is often that a underlying K8s node has been deleted in the meantime (since starting the tests).

Example of an infinite health probe attempt (until GitHub action timeout) in cilium/cilium due to Cilium Agent Pod deletion (Node has been deleted): https://github.com/cilium/cilium/actions/runs/7060997016/job/19221753163 

Related PR (connectivity test timeout): https://github.com/cilium/cilium-cli/pull/2145

**Alternative**

An alternative would be to not rely on the Cilium Agent Pods that have been gathered at the beginning of the test run - and instead re-fetch them at the beginning of the Health Probe test scenario. 

This might be something for a follow up PR. In this case i think it would be worth to keep the part about informing the user that a Cilium Agent Pod has been deleted. Maybe in a separate test that only checks for this.

Kind of boils down to the question what the health probe test should actually check for.

### DIFF
--- a/connectivity/tests/health.go
+++ b/connectivity/tests/health.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/jsonpath"
 
 	"github.com/cilium/cilium-cli/connectivity/check"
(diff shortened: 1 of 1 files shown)


tklauser (approved):
LGTM, thanks

PR #2151, opened 2023-12-05
connectivity: Check for unexpected packet drops

pchaigno:
This pull requests ensures we fail the connectivity tests if we hit any unexpected packet drops. The first commit transforms the existing missed-tail-call drops check into a new check for unexpected packet drops. The second commit defines a new flag to set the list of expected packet drop reasons. The last commit adds a bunch of packet drop reasons.

### DIFF
--- a/internal/cli/cmd/connectivity.go
+++ b/internal/cli/cmd/connectivity.go
@@ -190,6 +190,10 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 	cmd.Flags().StringVar(&params.ConnDisruptTestRestartsPath, "conn-disrupt-test-restarts-path", "/tmp/cilium-conn-disrupt-restarts", "Conn disrupt test temporary result file (used internally)")
 	cmd.Flags().StringVar(&params.ConnDisruptTestXfrmErrorsPath, "conn-disrupt-test-xfrm-errors-path", "/tmp/cilium-conn-disrupt-xfrm-errors", "Conn disrupt test temporary result file (used internally)")
 	cmd.Flags().DurationVar(&params.ConnDisruptDispatchInterval, "conn-disrupt-dispatch-interval", 0, "TCP packet dispatch interval")
+
+	cmd.Flags().StringSliceVar(&params.ExpectedDropReasons, "expected-drop-reasons", defaults.ExpectedDropReasons, "List of expected drop reasons")
+	cmd.Flags().MarkHidden("expected-drop-reasons")
+
 	cmd.Flags().BoolVar(&params.FlushCT, "flush-ct", false, "Flush conntrack of Cilium on each node")
 	cmd.Flags().MarkHidden("flush-ct")
 	cmd.Flags().StringVar(&params.SecondaryNetworkIface, "secondary-network-iface", "", "Secondary network iface name (e.g., to test NodePort BPF on multiple networks)")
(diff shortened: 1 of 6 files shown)


brb:
Nice intention, but IIRC "invalid source ip", "unknown l3 target address", "auth required", etc are always present.

internal/cli/cmd/connectivity.go
```diff
@@ -190,6 +190,9 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 	cmd.Flags().StringVar(&params.ConnDisruptTestRestartsPath, "conn-disrupt-test-restarts-path", "/tmp/cilium-conn-disrupt-restarts", "Conn disrupt test temporary result file (used internally)")
 	cmd.Flags().StringVar(&params.ConnDisruptTestXfrmErrorsPath, "conn-disrupt-test-xfrm-errors-path", "/tmp/cilium-conn-disrupt-xfrm-errors", "Conn disrupt test temporary result file (used internally)")
 	cmd.Flags().DurationVar(&params.ConnDisruptDispatchInterval, "conn-disrupt-dispatch-interval", 0, "TCP packet dispatch interval")
+
+	cmd.Flags().StringSliceVar(&params.ExpectedDropReasons, "expected-drop-reasons", defaults.ExpectedDropReasons, "List of expected drop reasons")
```
tklauser, line 194: Is this a flag that we expect users outside of cilium/cilium to modify? If not, I'd probably mark it as hidden. In case we want to expose it to users, I'd propose to add a bit more context to the usage message, e.g. that this applies to the `no-unexpected-packet-drops` test.

tklauser (commented):
LGTM apart from an open question on the new drop reasons CLI flag.

internal/cli/cmd/connectivity.go
pchaigno: I'm unsure, so it's probably best to hide it now and unhide if the use case ever comes up.

pchaigno:
> Nice! Have you tried running `/test` on cilium/cilium after running `git grep -l cilium_cli_ci_version: | xargs sed -i "s/cilium_cli_ci_version: .*/cilium_cli_ci_version: $CI_BUILD_VSN/g"`?

No. But we have a flag to set the list of expected reasons, so we can adjust if specific tests need additional reasons.

brb:
> No. But we have a flag to set the list of expected reasons, so we can adjust if specific tests need additional reasons.

From my personal experience, I'd use it as a last resort. It's difficult to keep a mental map in a head of all options which need to be customized in each job X branches. 

Also, if you don't test it now, then it might become someone else's problem once a new CLI version is released, and cilium/cilium's CLI dependency is bumped.

tklauser (approved):

brb (changes requested):
I tried running the change myself in https://github.com/cilium/cilium/pull/29718, and most of the red builds are due to this check failing.

If we merge it now (or w/o trying it out), then this change will block other changes (which will be released together) from being used in the cilium/cilium CI.

pchaigno:
> I tried running the change myself in [cilium/cilium#29718](https://github.com/cilium/cilium/pull/29718), and most of the red builds are due to this check failing.

I added those drop reasons.

> If we merge it now (or w/o trying it out), then this change will block other changes (which will be released together) from being used in the cilium/cilium CI.

That's FUD given we have a flag to add other drop reason exceptions.

brb (approved):

michi-covalent:
the kind failure is tracked in https://github.com/cilium/cilium-cli/issues/2162. merging 🚀🙏

PR #2159, opened 2023-12-07
Remove classic mode tests from the main branch

michi-covalent:
Let's remove classic mode tests from the main branch. In a very unlikely event of having to fix a bug in the classic mode for Cilium v1.12 or v1.13, we can create v0.14 branch from v0.14.8 tag and patch it there.

Ref: https://github.com/cilium/cilium-cli#releases


michi-covalent:
ok let's ship this

PR #2166, opened 2023-12-08
Don't modify running cluster in dry-run mode

### LINKED ISSUE #2161: Dangerous effect of using --dry-run: makes unannounced changes to EKS cluster DaemonSets
## Bug report
using --dry-run or --dry-run-helm-values does changes to eks aws-node. This should not be done, and while it is outputted to logs in code the logs are discarded when using --dry-run.

**General Information**
```
cilium-cli: v0.15.17 compiled with go1.21.4 on darwin/arm64
cilium image (default): v1.14.4
cilium image (stable): v1.14.4
```

Problematic code:
https://github.com/cilium/cilium-cli/blob/036d768db9768be6292ea8eb2cefd38e6485d277/install/install.go#L681C1-L694C4

1. Use EKS cluster
2. Run something similar to `cilium install --version 1.14.4 --values values-migration.yaml --dry-run-helm-values`
3. Watch how the node selector is changed

tklauser:
In EKS mode, don't stop the AWS node DaemonSet when running in dry-run mode. In AKS mode, don't create any cloud resources.

Fixes #2161

### DIFF
--- a/install/azure.go
+++ b/install/azure.go
@@ -193,7 +193,7 @@ func (k *K8sInstaller) azureSetupServicePrincipal() error {
 		}
 	}
 
-	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" {
+	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" && !k.params.IsDryRun() {
 		k.Log("🚀 Creating Azure Service Principal for Cilium Azure operator...")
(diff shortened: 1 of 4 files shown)


gandro (approved):

install/azure.go
```diff
@@ -193,7 +193,7 @@ func (k *K8sInstaller) azureSetupServicePrincipal() error {
 		}
 	}
 
-	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" {
+	if k.params.Azure.TenantID == "" && k.params.Azure.ClientID == "" && k.params.Azure.ClientSecret == "" && !k.params.IsDryRun() {
```
derailed, line 196: nit: no directly part of this pr. Would be nice to define a helper func here ie needsAzServicePrincipal(params) bool.

derailed (approved):
@tklauser Nice catch Tobias!

tommyp1ckles (approved):

PR #2183, opened 2023-12-14
check: add third client scheduled on a different node

jibi:
for some upcoming tests we will need a new client scheduled on a different node than the first 2 clients. This commit adds this new client.

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -143,7 +143,7 @@ jobs:
 
       - name: Wait for job
         env:
-          timeout: 20m
+          timeout: 30m
         run: |
           # Background wait for job to complete or timeout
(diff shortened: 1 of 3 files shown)


christarazi (commented):
Why is a third client needed?

jibi:
> Why is a third client needed?

as I added on the commit message, for some upcoming tests we will need a new client scheduled on a _different_ node than the first 2 clients (and these 2 clients always run on the same node, as client2 uses pod affinity to run on the same node as client1)

jibi:
eks is failing due to a timeout, link to the green workflow after bumping it and running the workflow from PR https://github.com/cilium/cilium-cli/actions/runs/7220606369/job/19673889321?pr=2183

brlbil (approved):

PR #2184, opened 2023-12-14
Enable no-errors-in-logs check by default, and extend it to all Cilium components

giorio94:
Currently, we check for errors in Cilium agent logs only when unsafe tests are enabled. Let's promote this check to be always performed for Cilium v1.14 and subsequent (earlier versions are not compatible with the Cilium CLI v0.15 series), to ensure that we catch unexpected errors early (as unsafe tests are only enabled in a limited set of CI workflows) given that it does not perform any unsafe operation. Additionally, let's extend it to also cover the init containers in the cilium agent pods, as well as other cilium components, such as the operator and the clustermesh-apiserver.

### DIFF
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -1232,7 +1232,7 @@ func Run(ctx context.Context, ct *check.ConnectivityTest, extra Hooks) error {
 		return err
 	}
 
-	if ct.Params().IncludeUnsafeTests {
+	if versioncheck.MustCompile(">=1.14.0")(ct.CiliumVersion) || ct.Params().IncludeUnsafeTests {
 		ct.NewTest("check-log-errors").WithScenarios(tests.NoErrorsInLogs(ct.CiliumVersion))
 	}
 
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -1233,7 +1233,9 @@ func Run(ctx context.Context, ct *check.ConnectivityTest, extra Hooks) error {
 	}
 
 	if versioncheck.MustCompile(">=1.14.0")(ct.CiliumVersion) || ct.Params().IncludeUnsafeTests {
-		ct.NewTest("check-log-errors").WithScenarios(tests.NoErrorsInLogs(ct.CiliumVersion))
+		ct.NewTest("check-log-errors").
+			WithSysdumpPolicy(check.SysdumpPolicyOnce).
+			WithScenarios(tests.NoErrorsInLogs(ct.CiliumVersion))
 	}
 
 	return ct.Run(ctx)
--- a/connectivity/tests/errors.go
+++ b/connectivity/tests/errors.go
@@ -10,9 +10,13 @@ import (
 	"time"
 
 	"github.com/blang/semver/v4"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/cilium-cli/connectivity/check"
 	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/k8s"
+	"github.com/cilium/cilium-cli/sysdump"
 	"github.com/cilium/cilium-cli/utils/features"
 )
 
@@ -61,17 +65,30 @@ func (n *noErrorsInLogs) Name() string {
 	return "no-errors-in-logs"
 }
 
+type podID struct{ Cluster, Namespace, Name string }
+type podInfo struct {
+	containers []string
+	client     *k8s.Client
+}
+
 func (n *noErrorsInLogs) Run(ctx context.Context, t *check.Test) {
 	var since time.Time
-	ct := t.Context()
 
-	for _, pod := range ct.CiliumPods() {
-		pod := pod
-		logs, err := pod.K8sClient.CiliumLogs(ctx, pod.Pod.Namespace, pod.Pod.Name, since, nil)
-		if err != nil {
-			t.Fatalf("Error reading Cilium logs: %s", err)
+	pods, err := n.allCiliumPods(ctx, t.Context())
+	if err != nil {
+		t.Fatalf("Error retrieving Cilium pods: %s", err)
+	}
+
+	for pod, info := range pods {
+		client := info.client
+		for _, container := range info.containers {
+			id := fmt.Sprintf("%s/%s/%s (%s)", pod.Cluster, pod.Namespace, pod.Name, container)
+			logs, err := client.GetLogs(ctx, pod.Namespace, pod.Name, container, since, sysdump.DefaultLogsLimitBytes, false)
+			if err != nil {
+				t.Fatalf("Error reading Cilium logs: %s", err)
+			}
+			n.checkErrorsInLogs(id, logs, t)
 		}
-		n.checkErrorsInLogs(logs, t)
 	}
 
 }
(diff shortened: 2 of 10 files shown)


giorio94:
Additionally tested in https://github.com/cilium/cilium/pull/30604 and https://github.com/cilium/cilium/pull/29890

k8s/client.go
```diff
@@ -929,15 +929,15 @@ func (c *Client) ListCiliumPodIPPools(ctx context.Context, opts metav1.ListOptio
 	return c.CiliumClientset.CiliumV2alpha1().CiliumPodIPPools().List(ctx, opts)
 }
 
-func (c *Client) GetLogs(ctx context.Context, namespace, name, container string, sinceTime time.Time, limitBytes int64, previous bool) (string, error) {
+func (c *Client) GetLogs(ctx context.Context, namespace, name, container string, sinceTime time.Time, limitBytes int64, previous, timestamps bool) (string, error) {
```
derailed, line 932: I understand if this might be out of scope for this pr but it seems that this call might as well take in PodLogOptions?

connectivity/suite.go
```diff
@@ -1232,8 +1232,10 @@ func Run(ctx context.Context, ct *check.ConnectivityTest, extra Hooks) error {
 		return err
 	}
 
-	if ct.Params().IncludeUnsafeTests {
-		ct.NewTest("check-log-errors").WithScenarios(tests.NoErrorsInLogs(ct.CiliumVersion))
+	if versioncheck.MustCompile(">=1.14.0")(ct.CiliumVersion) || ct.Params().IncludeUnsafeTests {
```
derailed, line 1235: This could be an expensive call if an operator runs this wide open on a larger cluster. ie scanning logs for errors = #cluster*#pod*#cos*#lines. Perhaps this should be an opt-in flag?

connectivity/tests/errors.go
```diff
@@ -61,17 +65,32 @@ func (n *noErrorsInLogs) Name() string {
...
-			t.Fatalf("Error reading Cilium logs: %s", err)
+	pods, err := n.allCiliumPods(ctx, t.Context())
+	if err != nil {
+		t.Fatalf("Error retrieving Cilium pods: %s", err)
+	}
+
+	for pod, info := range pods {
+		client := info.client
+		for _, container := range info.containers {
+			id := fmt.Sprintf("%s/%s/%s (%s)", pod.Cluster, pod.Namespace, pod.Name, container)
```
derailed, line 85: Since we have a podID type perhaps a convenience Stringer on this type would lean this out?

connectivity/tests/errors.go
```diff
@@ -61,17 +65,32 @@ func (n *noErrorsInLogs) Name() string {
...
+			t.NewGenericAction(n, id).Run(func(a *check.Action) {
+				logs, err := client.GetLogs(ctx, pod.Namespace, pod.Name, container, since, sysdump.DefaultLogsLimitBytes, false, false)
```
derailed, line 87: Might as well pass in a PodLogOpts?

connectivity/tests/errors.go
```diff
@@ -126,7 +145,52 @@ func computeExpectedDropReasons(defaultReasons, inputReasons []string) string {
...
+		id := podID{Cluster: pod.K8sClient.ClusterName(), Namespace: pod.Namespace(), Name: pod.NameWithoutNamespace()}
+		if _, ok := output[id]; !ok {
+			output[id] = podInfo{client: pod.K8sClient, containers: n.podContainers(pod.Pod)}
+		}
+	}
+
+	return output, nil
+}
+
+func (n *noErrorsInLogs) podContainers(pod *corev1.Pod) (containers []string) {
```
derailed, line 181: nit: No prealloc of cos slice. 
Also wondering if these are all agent pods or operator pods likely the container list should always be the same?

derailed (changes requested):
@giorio94 Nice work Marco!

k8s/client.go
giorio94: Sounds reasonable to me. I've updated it to take `PodLogOptions` instead of adding the extra parameter.

connectivity/tests/errors.go
giorio94: I'd personally keep it easy, given that this is used only once.

connectivity/tests/errors.go
giorio94: I'd personally prefer keep this generic and simple, avoiding to hardcode the knowledge about the number of containers associated with any pods (as it may change between versions). I'd also not overoptimize at this point, given that this logic is executed only once, and the collection of logs is orders of magnitude more expensive in any case.

tklauser (approved):

connectivity/suite.go
giorio94: I agree that this test is potentially expensive. The reasoning for enabling it by default (opposed to it being behind a special flag) is that we want to run it for all CI workflows (which are the main consumers of the connectivity tests in any case), and making it opt-in would make that really tricky to achieve. That said, it is still possible for a user to either disable this check through the `--test` flag, or specify a custom label selection which matches only a subset of nodes. This is consistent with the behavior of other expensive checks, like the health one.

derailed (approved):
@giorio94 Thanks for the updates!

connectivity/tests/errors.go
```diff
@@ -61,17 +65,31 @@ func (n *noErrorsInLogs) Name() string {
...
-			t.Fatalf("Error reading Cilium logs: %s", err)
+	opts := corev1.PodLogOptions{LimitBytes: ptr.To[int64](sysdump.DefaultLogsLimitBytes)}
+	for pod, info := range pods {
+		client := info.client
+		for _, container := range info.containers {
+			id := fmt.Sprintf("%s/%s/%s (%s)", pod.Cluster, pod.Namespace, pod.Name, container)
+			t.NewGenericAction(n, id).Run(func(a *check.Action) {
+				logs, err := client.GetLogs(ctx, pod.Namespace, pod.Name, container, opts)
+				if err != nil {
+					a.Fatalf("Error reading Cilium logs: %s", err)
```
tommyp1ckles, line 88: Just wondering, do we want to fatal out here? seems like we could still proceed with other checks even if this fails?

connectivity/tests/errors.go
```diff
@@ -146,14 +209,13 @@ func (n *noErrorsInLogs) checkErrorsInLogs(logs string, t *check.Test) {
...
-			failures = append(failures, f)
-
-			t.Logf("Found %q in logs %d times\n", f, c)
+			failures.WriteRune('\n')
+			failures.WriteString(f)
+			failures.WriteString(fmt.Sprintf(" (%d occurrences)", c))
 		}
-		failureMsgs := strings.Join(failures, "\n")
-		t.Failf("Found %d logs matching list of errors that must be investigated:\n%s", len(uniqueFailures), failureMsgs)
+		a.Failf("Found %d logs in %s matching list of errors that must be investigated:%s", len(uniqueFailures), id, failures.String())
```
tommyp1ckles, line 218: q: do we actually match this anywhere, it appears like we just report on unique errors.

tommyp1ckles (approved):
Very valuable improvement, looks good aside from some questions.

connectivity/tests/errors.go
giorio94: I don't have strong opinions honestly. The current implementation mimics the previous one in this respect, which already fataled out in case we failed reading the logs (with the caveat that the test suite then recovers from the panic, and continues with the execution of the other scenarios). And this is also consistent with other checks which also fatal out in case something goes wrong during retrieval of information (e.g., drop counts). 

IMHO, we could likely improve this logic to be more resilient in case of agent restart and/or spot instances rotation, but that looks a more generic and long term effort to me, which is unrelated from this specific PR.

connectivity/tests/errors.go
giorio94: Sorry, I'm not sure I'm fully understanding your question here. If you refer to the *matching list* wording, that refers to the fact that we check each log entry against a set of [known *unexpected* messages](https://github.com/giorio94/cilium-cli/blob/17bb0867880ca0d9b06efcaa9388b631a85e5c37/connectivity/tests/errors.go#L30-L52) (with possible exceptions), and trigger a failure if any of them is found. Multiple occurrences of the same entry are then coalescenced for the sake of brevity.

PR #2168, opened 2023-12-08
Move perf test command to separate subcommand

marseel:
I really tried to be mindful of splitting that into commits, but unfortunately, I ended up refactoring the whole thing :cry: 

Summary:
- Move `cilium connectivity test --perf` to separate subcommand `cilium connectivity perf`
- Allow to run host/pod tests together (before you could only run one of), while also allow to test host-to-pod, pod-to-host etc.
- Before NodeSelector was totally not working in perf test and we were forcing users to use one zone for nodes, now nodeSelector works and we just warn in case nodes are in different zones (so now we can test both in-zone and across-zones)
-  Added possibility to export results in json format that can be visualized with perfdash - this will be used for tracking datapath performance in CI
- Visualize operation rate next to latency in console output (as they were from the same netperf test)
- Always cleanup perf pods before performing tests - to make sure we respect nodeSelector
- Refactored netperf integration, it was full of hacks.

Help for perf subcommand:
```
cilium connectivity perf --help
Test network performance

Usage:
  cilium connectivity perf [flags]

Flags:
  -d, --debug                             Show debug messages
      --deployment-pod-annotations json   Add annotations to the connectivity pods, e.g. '{"client":{"foo":"bar"}}' (default {})
      --duration duration                 Duration for the Performance test to run (default 10s)
  -h, --help                              help for perf
      --host-net                          Test host network
      --node-selector stringToString      Restrict connectivity pods to nodes matching this label (default [])
      --performance-image string          Image path to use for performance (default "quay.io/cilium/network-perf:a816f935930cb2b40ba43230643da4d5751a5711@sha256:679d3a370c696f63884da4557a4466f3b5569b4719bb4f86e8aac02fbe390eea")
      --pod-net                           Test pod network (default true)
      --report-dir string                 Directory to save perf results in json format
      --samples int                       Number of Performance samples to capture (how many times to run each test) (default 1)
      --test-namespace string             Namespace to perform the connectivity in (default "cilium-test")

Global Flags:
      --context string     Kubernetes configuration context
  -n, --namespace string   Namespace Cilium is running in (default "kube-system")

```

Example output of full tests:
```
🔥 Network Performance Test Summary:
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
📋 Scenario        | Node       | Test            | Duration        | Min             | Mean            | Max             | P50             | P90             | P99             | Transaction rate OP/s
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #2185, opened 2023-12-14
Fixes cilium clustermesh disconnect

littlejo:
`cilium clustermesh disconnect` never work with helm version:
`remote cluster has cluster ID %d out of acceptable range (1-0)`

So i fix it.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -2254,9 +2254,6 @@ func (k *K8sClusterMesh) DisconnectWithHelm(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	if err = k.validateInfoForConnect(aiLocal, aiRemote); err != nil {
-		return err
-	}
 
 	// Modify the clustermesh config to remove the intended cluster if any
 	localHelmValues, err := removeFromClustermeshConfig(localRelease.Config, aiRemote.ClusterName)


clustermesh/clustermesh.go
```diff
@@ -2243,7 +2272,7 @@ func (k *K8sClusterMesh) DisconnectWithHelm(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	if err = k.validateInfoForConnect(aiLocal, aiRemote); err != nil {
```
thorn3r, line 2246: I think rather than creating a new validation function, we can safely remove this validation from disconnect. The validation in `validateInfoForConnect` is there to prevent creating a broken clustermesh configuration, so we shouldn't need to re-validate that before removing the configuration as well.

The reason why this is breaking is because we're using `shallowExtractAccessInfo` rather than the full `extractAccessInfo`, but using the same validate function (which is assuming all of the data has been extracted).

thorn3r (changes requested):
Thanks for the submission @littlejo! Left a suggestion here on a simpler approach to fixing this bug

clustermesh/clustermesh.go
littlejo: @thorn3r do you have news?

thorn3r (approved):
Looks good! Sorry for the delay, I was away for the holidays.

thorn3r:
@littlejo Looks like the perf tests were moved to their own subcommand in https://github.com/cilium/cilium-cli/pull/2168. The GH workflows run on `main` which have those changes, but your branch doesn't. If you rebase your PR it should fix these failures

kariya-mitsuru:
I am eagerly awaiting this fix to be merged, what are the blocking factors?

littlejo:
@thorn3r do you have news about this PR? Thank you.

thorn3r:
Hey @littlejo it looks like your failure is related to [this issue](https://github.com/cilium/cilium-cli/issues/2222), which was fixed [here](https://github.com/cilium/cilium-cli/pull/2226).

Sorry to ask for this again, but can you rebase once more? The `upload-artifact` action we're using in that workflow was updated to v4.0.0, which included breaking changes. It's now on v4.3.0, which I believe includes resolution for this - but we've also given the artifacts unique names as per the PR I linked above.

michi-covalent:
this will be included in the upcoming v0.15.22 release 🚀

PR #2190, opened 2023-12-15
connectivity: Ignore expected XFRM errors

pchaigno:
First commit refactors the existing logic for the `--expected-drop-reasons`. The second adds a new `--expected-xfrm-errors`.


michi-covalent (approved):

nbusseneau (approved):

jschwinger233 (approved):
If `len(inputExceptions) == 0`, `ComputeFailureExceptions` returns an empty slice regardless of `defaultExceptions`. But the original logic appears to be same, there should be not a problem.

PR #2195, opened 2023-12-19
Make connectivity test pods terminate quickly

giorio94:
Let's use the proper pause entrypoint, instead of sleep, so that the container reacts to signals and terminates immediately when the pod is deleted, instead of waiting for the expiration of the grace period.

Related: https://github.com/cilium/alpine-curl/pull/97


tklauser (approved):
Thanks

PR #2205, opened 2023-12-21
Drop cilium endpoint verification.

viktor-kurchenko:
Looks like Cilium endpoint verification is no longer needed because Cilium should be already provisioned correctly in the environment before deploying the connectivity apps, and these tests already validate that the k8s deployment resource is ready before continuing.

This should also unblock connectivity tests for Customers that have `disableEndpointCRD: "true"` property configured.

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -1099,8 +1099,13 @@ func (ct *ConnectivityTest) DeleteConnDisruptTestDeployment(ctx context.Context,
 
 // validateDeployment checks if the Deployments we created have the expected Pods in them.
 func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
-
 	ct.Debug("Validating Deployments...")
+
+	disableEndpointCRD := false
+	if status, ok := ct.Feature(features.DisableEndpointCRD); ok && status.Enabled {
+		disableEndpointCRD = true
+		ct.Info("CiliumEndpoint CRD check disabled")
+	}
 
 	srcDeployments, dstDeployments := ct.deploymentList()
 	for _, name := range srcDeployments {
@@ -1122,7 +1127,7 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 		}
 		for _, perfPod := range perfPods.Items {
 			// Individual endpoints will not be created for pods using node's network stack
-			if !perfPod.Spec.HostNetwork {
+			if !perfPod.Spec.HostNetwork && !disableEndpointCRD {
 				if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, perfPod.Name); err != nil {
 					return err
 				}
@@ -1157,8 +1162,10 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	}
 
 	for _, pod := range clientPods.Items {
-		if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, pod.Name); err != nil {
-			return err
+		if !disableEndpointCRD {
+			if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, pod.Name); err != nil {
+				return err
+			}
 		}
 
 		if strings.Contains(pod.Name, clientCPDeployment) {
@@ -1244,8 +1251,10 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 			return fmt.Errorf("unable to list echo pods: %w", err)
 		}
 		for _, echoPod := range echoPods.Items {
-			if err := WaitForCiliumEndpoint(ctx, ct, client, echoPod.GetNamespace(), echoPod.GetName()); err != nil {
-				return err
+			if !disableEndpointCRD {
+				if err := WaitForCiliumEndpoint(ctx, ct, client, echoPod.GetNamespace(), echoPod.GetName()); err != nil {
+					return err
+				}
 			}
 
 			ct.echoPods[echoPod.Name] = Pod{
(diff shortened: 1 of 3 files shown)


connectivity/check/wait.go
```diff
@@ -53,9 +53,13 @@ func WaitForDeployment(ctx context.Context, log Logger, client *k8s.Client, name
 }
 
 // WaitForCiliumEndpoint waits until the specified cilium endpoint gets created.
-func WaitForCiliumEndpoint(ctx context.Context, log Logger, client *k8s.Client, namespace, name string) error {
-	log.Logf("⌛ [%s] Waiting for CiliumEndpoint for pod %s/%s to appear...", client.ClusterName(), namespace, name)
+func WaitForCiliumEndpoint(ctx context.Context, log Logger, client *k8s.Client, namespace string, ciliumEndpointDisabled bool, name string) error {
+	if ciliumEndpointDisabled {
```
asauber, line 57: Adding a flag which makes this function a no-op doesn't make much sense. Checking this bool should be the responsibility of the caller, not the function.

Consider the following

```
func doSomething(dontDoIt bool) {
   if dontDoIt {
       return
   }
   actuallyDoIt()
}
```

compared to

```
func doSomething() {
   actuallyDoIt()
}

// ...

if !dontDoIt {
   doSomething()
}
```

Regardless of how we discover the value of `ciliumEndpointDisabled`. It should not be a parameter of this function.

internal/cli/cmd/connectivity.go
```diff
@@ -157,6 +157,7 @@ func newCmdConnectivityTest(hooks Hooks) *cobra.Command {
 	cmd.Flags().MarkHidden("include-unsafe-tests")
 	cmd.Flags().BoolVar(&params.K8sLocalHostTest, "k8s-localhost-test", false, "Include tests which test for policy enforcement for the k8s entity on its own host")
 	cmd.Flags().MarkHidden("k8s-localhost-test")
+	cmd.Flags().BoolVar(&params.CiliumEndpointDisabled, "skip-cilium-endpoint-check", false, "Skip Cilium endpoint CRD check")
```
asauber, line 160: Rather than add a flag, can we read the value of `disableEndpointCRD: "true"` from Cilium's live config.

This would be more accurate and easier to use.

asauber (changes requested):

connectivity/check/wait.go
viktor-kurchenko: Fixed.

internal/cli/cmd/connectivity.go
viktor-kurchenko: Done, thanks!

asauber (approved):
Looks great, thanks for the updates

nathanjsweet (approved):

brlbil (approved):

connectivity/check/deployment.go
```diff
@@ -1157,8 +1162,10 @@ func (ct *ConnectivityTest) validateDeployment(ctx context.Context) error {
 	}
 
 	for _, pod := range clientPods.Items {
-		if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, pod.Name); err != nil {
-			return err
+		if !disableEndpointCRD {
+			if err := WaitForCiliumEndpoint(ctx, ct, ct.clients.src, ct.Params().TestNamespace, pod.Name); err != nil {
```
joestringer, line 1166: Is the intent here to better support testing Cilium that's configured in kvstore mode?

I'm tracing this back to https://github.com/cilium/cilium/pull/5933 , which given the timing may be when we were initially introducing CEP support so most users were on kvstore.

The main concern I have with this approach is the potential for race conditions. Currently, this check seems to ensure that some state is configured in the cluster before continuing. If we skip that check entirely and don't create an equivalent check for kvstore state, then this could cause tests to pass even if the cluster is not ready (because Cilium just isn't handling the traffic due to the race), or cause failures because the test didn't wait long enough (even if the test would otherwise pass if it _did_ wait long enough).

It's not obvious to me why this PR proposal is safe and why it won't introduce flakiness into the tests in this deployment scenario.

joestringer (changes requested):
I have a concern below about the behaviour of the testsuite with this proposal.

connectivity/check/deployment.go
joestringer: I discussed this a bit with @marseel out-of-band and he brought up that these deployments are deployed by the CLI tool as part of `SetupAndValidate()`, explicitly *after* ensuring that Cilium is deployed correctly, so there's no pattern here where the apps are deployed without Cilium being set up correctly. As long as the Deployment is ready, this should provide a strong enough guarantee that the CEPs (or equivalent state) is also propagated.

In that case I have no remaining reservations. The main question to me would be whether we need this option at all - would it be simpler to just remove these wait conditions altogether instead of making them conditional?

viktor-kurchenko: Thank you, Joe!
I don't have a strong opinion here... Maybe let's keep the condition at least for a while.

joestringer: The cost of "keeping it for a while" is that there's no obvious time to revisit this. By deleting it in only one specific test scenario, it makes it more likely we won't notice problems, and in the mean time the code is more complex. I'd rather keep the code simpler and remove it entirely if it's unnecessary. Then if it turns out we're wrong and it's needed, then we will also learn about that sooner and in all testing cases instead of just in this one new testing mode.

viktor-kurchenko: Agree, I've removed the function `WaitForCiliumEndpoint`.

CC: @asauber @nathanjsweet

joestringer (approved):

joestringer:
I think it would be useful to include the context in the description that: This is no longer needed because Cilium should be already provisioned correctly in the environment before deploying the connectivity apps, and these tests already validate that the k8s deployment resource is ready before continuing.

Issue #2222, opened 2024-01-10
CI: JUnit upload failure in kind workflow: artifact with this name already exists on the workflow run


tklauser:
```
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

Seen on https://github.com/cilium/cilium-cli/actions/runs/7469244368/job/20326056243?pr=2221

tklauser:
Another occurence on #2224: https://github.com/cilium/cilium-cli/actions/runs/7473789182/job/20338628475?pr=2224

PR #2224, opened 2024-01-10
ci: remove unnecessary steps unlabeling nodes in kind workflow

tklauser:
The nodes are no longer labeled in the kind workflow since commit e1b543279e1c ("kind: Configure external targets inside the cluster") so they also don't need to be unlabeled on cleanup.


viktor-kurchenko (approved):

tklauser:
Kind workflow failed on #2222: https://github.com/cilium/cilium-cli/actions/runs/7473789182/job/20338628475?pr=2224

PR #2233, opened 2024-01-12
Fix status exit code

### LINKED ISSUE #2232: cilium status: exit-code wrong in case of errors
## Bug report

**General Information**

I noticed that the `cilium status` command does always return `0` as exit-code in case of failures, instead of `1`.
So I will fix it.

1. `cilium status` on a failing/non-cilium cluster
2. `echo $?` returns 0

ttretau:
`cilium status` does currently return exit-code 0 in case of errors.

Adding an error here in case exists in the sub command, so it will result in correct exit-code `1`.

Fixes #2232

### DIFF
--- a/internal/cli/cmd/status.go
+++ b/internal/cli/cmd/status.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 
@@ -46,6 +47,10 @@ func newCmdStatus() *cobra.Command {
 				fmt.Println(string(jsonStatus))
 			} else {
 				fmt.Print(s.Format())
+			}
+
+			if err == nil && len(s.CollectionErrors) > 0 {
+				err = errors.New("status check failed")
 			}
 			return err
 		},


internal/cli/cmd/status.go
```diff
@@ -47,6 +48,10 @@ func newCmdStatus() *cobra.Command {
 			} else {
 				fmt.Print(s.Format())
 			}
+
+			if err == nil && len(s.CollectionErrors) > 0 {
```
derailed, line 52: I think we could do something like:
```go
  return errors.Join(err, errors.Join(s.CollectionErrors...))
```

ttretau: @derailed Thanks for the idea!

Unfortunately this would interfere a little bit with the [overall output](https://github.com/ttretau/cilium-cli/blob/89f65e3e1019be58f5bf00385745045bc50df955/cmd/cilium/main.go#L23) of errors. For example in Status [errors are already in output](https://github.com/ttretau/cilium-cli/blob/89f65e3e1019be58f5bf00385745045bc50df955/status/status.go#L371). Therefore the simple error message.

example output with the join:
```
Containers:            cilium
                       cilium-operator
Cluster Pods:          0/41 managed by Cilium
Helm chart version:
Errors:                cilium    cilium    daemonsets.apps "cilium" not found
daemonsets.apps "cilium" not found

derailed (approved):
@ttretau Thank you for looking into it! 
Likely a separate pr to surface and format/print the actual error(s).

PR #2249, opened 2024-01-19
Retrieve CiliumNodes in connectivity tests context, and use them to determine PodCIDRs

giorio94:
Add CiliumNodes to the list of resources which are retrieved during connectivity tests initialization, and stored for later access from individual tests. Indeed, CiliumNodes contain additional information compared to plain Kubernetes Node objects, especially around IPAM. Additionally, leverage them to determine the PodCIDRs used for certain (unsafe) tests, so that we retrieve the correct CIDR both in case of kubernetes and cluster-pool IPAM modes, whereas now it only works in the first case.


giorio94:
Converting back to draft as one of the CI failures seems legitimate.

giorio94:
CI is green. Marking ready for review again.

tklauser (approved):

PR #2251, opened 2024-01-19
bgp: routes CLI - update min. cilium version, use `cilium` binary

rastislavs:
As the necessary cilium-dbg functionality for retrieving BGP routes was backported to cilium v1.14.6 (https://github.com/cilium/cilium/pull/30205), we have to use the `cilium` binary instead of `cilium-dbg` to fetch routes from cilium pods, as the rename from `cilium` to `cilium-dbg` happened in 1.15.

This also updates the minimal cilium version for this command in the command description.

### DIFF
--- a/bgp/routes.go
+++ b/bgp/routes.go
@@ -87,7 +87,7 @@ func (s *Status) fetchRoutesConcurrently(ctx context.Context, args []string) (ma
 	wg.Add(len(s.ciliumPods))
 
 	// compute the command for fetching the routes from a cilium pod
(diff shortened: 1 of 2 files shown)


harsimran-pabla (approved):
Thanks, we should think about detecting cilium version and use corresponding cilium binary name. That's bit complicated and should probably be generalized across cli. When `cilium` alias to `cilium-dbg` gets deprecated from newer releases - we will have to figure this out. For now, this lgtm - thanks :)

derailed (approved):
@rastislavs LGTM

PR #2261, opened 2024-01-24
Display formatted status during status wait

raphink:
Signed-off-by: Raphaël Pinson <raphael@isovalent.com>


status/k8s.go
```diff
@@ -357,9 +358,18 @@ func (k *K8sStatusCollector) Status(ctx context.Context) (*Status, error) {
 		}
 		if !k.statusIsReady(s) && k.params.Wait {
 			time.Sleep(defaults.WaitRetryInterval)
+			statusFmt := s.Format()
+			for i := 1; i < lines; i++ {
```
derailed, line 362: Wondering if we could use a progress bar for this. Likely there might be other places that could benefit with user feedback.

We should probably use a help func to clear out the output if any.
nit: Docs might help here to indicate what this sequence does...

derailed (commented):
@raphink Thanks for this update Raphael!

status/k8s.go
raphink: Ah yes, the sequence is to set the cursor one line up.

raphink: As for the progress bar, I'm not a big fan because:

1.  we can't really know the % progress
2. as a user, I'd rather have details of what is going on (and see if something is stuck) in the same format as `cilium status`)

status/k8s.go
```diff
@@ -357,13 +358,28 @@ func (k *K8sStatusCollector) Status(ctx context.Context) (*Status, error) {
 		}
 		if !k.statusIsReady(s) && k.params.Wait {
 			time.Sleep(defaults.WaitRetryInterval)
+			if k.params.Output != "json" {
```
derailed, line 361: I think we should use the predefined const. Also perhaps best to check `summary` vs `json` in case other fmts are exposed in the future.

status/k8s.go
```diff
@@ -357,13 +358,28 @@ func (k *K8sStatusCollector) Status(ctx context.Context) (*Status, error) {
...
+				statusFmt := s.Format()
+				cursorUp(lines)
+				lines = len(strings.Split(statusFmt, "\n"))
+				fmt.Print(statusFmt)
+			}
 			continue
 		}
 
+		for i := 1; i < lines; i++ {
+			fmt.Print("\033[A\033[2K")
```
derailed, line 371: How does this sequence fare on different platforms?

derailed (changes requested):
@raphink Thanks for the updates!

status/k8s.go
raphink: Yes. The const is in a package I can't import though, because that creates a dependency loop. Should we move the const to another package?

status/k8s.go
raphink: It should be pretty safe on Linux and MacOS. On Windows, it ANSI codes seem to be well supported for Windows 10+.

raphink:
@derailed I believe I've addressed all comments!

derailed (approved):
@raphink Thank you for the updates Raphael!

ti-mo:
I think all jobs need to be rerun, looks like there was an issue with `helm.cilium.io` yesterday.

raphink:
OK, let's run again 👍🏼

raphink:
```
looks like "https://helm.cilium.io/" is not a valid chart repository or cannot be reached: Get "https://helm.cilium.io/index.yaml": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

still broken 😢

raphink:
That's really weird though, that repo works on my machine 🤔

PR #2267, opened 2024-01-28
connectivity: add endpointslice clustermesh sync test

MrFreezeex:
This adds endpointslice synchronization testing inside a clustermesh. It does that by checking that we can get a DNS answer with a global headless service with endpointslice synchronization enabled across two clusters.

Testing it via DNS ensures that the created EndpointSlice are correctly linked to the headless service.

Related to https://github.com/cilium/cilium/pull/28440 e2e testing

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -43,15 +43,16 @@ const (
 
 	DNSTestServerContainerName = "dns-test-server"
 
-	echoSameNodeDeploymentName     = "echo-same-node"
-	echoOtherNodeDeploymentName    = "echo-other-node"
-	echoExternalNodeDeploymentName = "echo-external-node"
-	corednsConfigMapName           = "coredns-configmap"
-	corednsConfigVolumeName        = "coredns-config-volume"
-	kindEchoName                   = "echo"
-	kindEchoExternalNodeName       = "echo-external-node"
-	kindClientName                 = "client"
-	kindPerfName                   = "perf"
+	echoSameNodeDeploymentName                 = "echo-same-node"
+	echoOtherNodeDeploymentName                = "echo-other-node"
+	EchoOtherNodeDeploymentHeadlessServiceName = "echo-other-node-headless"
+	echoExternalNodeDeploymentName             = "echo-external-node"
+	corednsConfigMapName                       = "coredns-configmap"
+	corednsConfigVolumeName                    = "coredns-config-volume"
+	kindEchoName                               = "echo"
+	kindEchoExternalNodeName                   = "echo-external-node"
+	kindClientName                             = "client"
+	kindPerfName                               = "perf"
 
 	hostNetNSDeploymentName          = "host-netns"
 	hostNetNSDeploymentNameNonCilium = "host-netns-non-cilium" // runs on non-Cilium test nodes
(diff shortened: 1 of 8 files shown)


MrFreezeex:
I am converting this to draft as I guess we need merge https://github.com/cilium/cilium/pull/28440 first? Unsure of the exact timing here :sweat_smile: and not sure if I added

I tested that locally and it seems to work fine but not sure how to trigger this in the CI as well as this test an ongoing PR...

```
[=] Test [endpointslice-mesh] [66/66]
..

✅ All 1 tests (2 actions) successful, 65 tests skipped, 0 scenarios skipped.
```

MrFreezeex:
> as the newly introduce headless service should be skipped in a few more places.

Ah yes now I need to ignore this from some tests and not only from the "deployment" phase, will look into it

bimmlerd (approved):
CODEOWNERS lgtm

connectivity/tests/endpointslice-clustermesh-sync.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (s *endpointSliceClusterMeshSync) Name() string {
+	return "endpointslice-clustermesh-sync"
+}
+
+func (s *endpointSliceClusterMeshSync) Run(ctx context.Context, t *check.Test) {
+	ct := t.Context()
+	client := ct.RandomClientPod()
+
+	service, ok := ct.EchoServicesAll()["echo-other-node-headless service"]
```
giorio94, line 28: I guess this should be `...[check.EchoOtherNodeDeploymentHeadlessServiceName]` (you'd need to export that variable though).

connectivity/check/deployment.go
```diff
@@ -531,6 +532,16 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
+
+			svcHeadless := svc.DeepCopy()
+			svcHeadless.Name = echoOtherNodeDeploymentHeadlessServiceName
+			svcHeadless.Spec.ClusterIP = corev1.ClusterIPNone
+			svcHeadless.Spec.Type = corev1.ServiceTypeClusterIP
+			svcHeadless.ObjectMeta.Annotations["service.cilium.io/global-sync-endpoint-slices"] = "true"
+			_, err = ct.clients.src.CreateService(ctx, ct.params.TestNamespace, svcHeadless, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
```
giorio94, line 544: Ideally, this should be guarded by the `GetService` check as above, to prevent failing if the service already exists. That said, this logic is already flawed in general, as we don't perform an upgrade in that case, hence it doesn't matter that much. Same for the one below.

connectivity/tests/endpointslice-clustermesh-sync.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func EndpointSliceClusterMeshSync() check.Scenario {
+	return &endpointSliceClusterMeshSync{}
+}
+
+type endpointSliceClusterMeshSync struct{}
+
+func (s *endpointSliceClusterMeshSync) Name() string {
+	return "endpointslice-clustermesh-sync"
```
giorio94, line 21: I've personally a small preference towards `clustermesh-endpointslice-sync` (same for the variable names), but no strong feelings. Feel free to keep it as it is now.

giorio94 (commented):
A couple of final comments inline, then I'll open the PR to test it on cilium/cilium.

connectivity/check/deployment.go
MrFreezeex: Indeed! I had to change the logic a bit for the copy to be a bit more convenient

christarazi (approved):

giorio94 (approved):
Thanks!

Additionally tested in https://github.com/cilium/cilium/pull/31551 as well.

giorio94 (approved):

connectivity/tests/clustermesh-endpointslice-sync.go
```diff
@@ -0,0 +1,38 @@
...
+	return "clustermesh-endpointslice-sync"
+}
+
+func (s *clusterMeshEndpointSliceSync) Run(ctx context.Context, t *check.Test) {
+	ct := t.Context()
+	client := ct.RandomClientPod()
+
+	service, ok := ct.EchoServicesAll()[check.EchoOtherNodeDeploymentHeadlessServiceName]
+	if !ok {
+		t.Fatal("Cannot get echo-other-node-headless service")
```
asauber, line 30: The const could be used in this error message, rather than a string literal with the name of the service.

asauber (approved):

connectivity/tests/clustermesh-endpointslice-sync.go
MrFreezeex: Indeed, thanks :D

MrFreezeex:
This should be ready :tada:, the Multicluster action failing seems to be an unrelated flake that could be retried I believe

giorio94:
> the Multicluster action failing seems to be an unrelated flake that could be retried I believe

Yep, I've triggered a retry

PR #2270, opened 2024-01-29
hooks: split feature detection from setup and validation

giorio94:
Currently, the SetupAndValidate extension hook is executed right after feature detection, but before the setup and validation of the different connectivity test pods and artifacts. While this has the advantage of nicely integrating with the logging of the detected features, it does not allow the hook function to rely on the existence of the pods created immediately afterwards. To enable both functionalities, let's split the hook into two -- one first function to detect extra features, and a second to perform the actual setup and validation of extra pods.

### DIFF
--- a/connectivity/suite.go
+++ b/connectivity/suite.go
@@ -195,9 +195,15 @@ var (
 	client2Label = map[string]string{"name": "client2"}
 )
 
-func Run(ctx context.Context, ct *check.ConnectivityTest, addExtraTests func(*check.ConnectivityTest) error,
-	addExtraSetup func(context.Context, *check.ConnectivityTest) error) error {
-	if err := ct.SetupAndValidate(ctx, addExtraSetup); err != nil {
+// Hooks defines the extension hooks provided by connectivity tests.
+type Hooks interface {
+	check.SetupHooks
+	// AddConnectivityTests is an hook to register additional connectivity tests.
+	AddConnectivityTests(ct *check.ConnectivityTest) error
+}
+
+func Run(ctx context.Context, ct *check.ConnectivityTest, extra Hooks) error {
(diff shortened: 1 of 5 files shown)


christarazi (approved):

connectivity/suite.go
```diff
@@ -195,9 +195,14 @@ var (
 	client2Label = map[string]string{"name": "client2"}
 )
 
-func Run(ctx context.Context, ct *check.ConnectivityTest, addExtraTests func(*check.ConnectivityTest) error,
-	addExtraSetup func(context.Context, *check.ConnectivityTest) error) error {
-	if err := ct.SetupAndValidate(ctx, addExtraSetup); err != nil {
+type hooks interface {
+	DetectFeatures(ctx context.Context, ct *check.ConnectivityTest) error
```
derailed, line 199: Wondering if we could  compose off interface defined above?

nit: Docs on ifc methods

internal/cli/cmd/hooks.go
```diff
@@ -20,6 +20,7 @@ type Hooks interface {
 
 // ConnectivityTestHooks to extend cilium-cli with additional connectivity tests and related flags.
 type ConnectivityTestHooks interface {
+	DetectFeatures(ctx context.Context, ct *check.ConnectivityTest) error
```
derailed, line 23: Same here. interface composition?

derailed (commented):
@giorio94 Thanks for the update Marco!
Perhaps leveraging interface composition (if possible) could dry this up a bit more?

michi-covalent (approved):

giorio94:
Rebased to pick the fix for the EKS tunnel workflow

derailed (approved):
@giorio94 LGTM

PR #2275, opened 2024-01-30
Remove no longer necessary step from the external workloads installation script generation process

giorio94:
Currently, the `cilium clustermesh external-workload install` command attempts to retrieve the Cilium version from either the secret created by the CLI (when running in classic mode) or from the helm release. The retrieved version is then used to tune the one of the options passed to the Cilium container on the external workload, which at a certain point got renamed. Given that all supported Cilium stable versions do support the new flag, and considering that the retrieval logic is potentially fragile (e.g., it fails if Cilium is installed with a non-default release name), let's just drop all of this logic. Regardless, we retrieve the Cilium image from the existing DaemonSet.


michi-covalent (approved):
let's do it

nathanjsweet (approved):

thorn3r (approved):
excellent :+1:

PR #2285, opened 2024-02-05
gateway: Upgrade API version

sayboras:
This is to avoid the below warning

```
W0205 12:29:06.703289  146168 warnings.go:70] The v1alpha2 version of ReferenceGrant has been deprecated and will be removed in a future release of the API. Please upgrade to v1beta1.
```

### DIFF
--- a/sysdump/constants.go
+++ b/sysdump/constants.go
@@ -158,25 +158,25 @@ var (
 	gatewayClass = schema.GroupVersionResource{
 		Group:    "gateway.networking.k8s.io",
 		Resource: "gatewayclasses",
-		Version:  "v1beta1",
+		Version:  "v1",
 	}
 
 	gateway = schema.GroupVersionResource{
 		Group:    "gateway.networking.k8s.io",
 		Resource: "gateways",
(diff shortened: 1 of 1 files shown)


mhofstetter (commented):
AFAIU this will break backward compatibility support for clusters with Cilium v1.13 installed. This version still installs [Gateway API in version 0.5.1 ](https://docs.cilium.io/en/v1.13/network/servicemesh/gateway-api/gateway-api/)that doesn't contain the new api versions.

Is this ok?

sayboras:
> AFAIU this will break backward compatibility support for clusters with Cilium v1.13 installed. This version still installs [Gateway API in version 0.5.1 ](https://docs.cilium.io/en/v1.13/network/servicemesh/gateway-api/gateway-api/)that doesn't contain the new api versions.
> 
> Is this ok?

I think it should be alright as long as we have another entry for 1.15+ in https://github.com/cilium/cilium-cli?tab=readme-ov-file#releases

mhofstetter (approved):

PR #2287, opened 2024-02-06
IPsec key status command implementation.

viktor-kurchenko:
`encryption key-status` command implementation to address: https://github.com/cilium/cilium/issues/28174.

### DIFF
--- a/encrypt/ipsec_key_status.go
+++ b/encrypt/ipsec_key_status.go
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package encrypt
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/status"
+)
+
+// IPsecKeyStatus displays IPsec key.
+func (s *Encrypt) IPsecKeyStatus(ctx context.Context) error {
+	ctx, cancelFn := context.WithTimeout(ctx, s.params.WaitDuration)
+	defer cancelFn()
+
+	secret, err := s.client.GetSecret(ctx, s.params.CiliumNamespace, defaults.EncryptionSecretName, metav1.GetOptions{})
+	if err != nil {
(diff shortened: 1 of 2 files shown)


encrypt/ipsec_key_status.go
```diff
@@ -0,0 +1,44 @@
...
+// IPsecKeyStatus displays IPsec key.
+func (s *Encrypt) IPsecKeyStatus(ctx context.Context) error {
+	ctx, cancelFn := context.WithTimeout(ctx, s.params.WaitDuration)
+	defer cancelFn()
+
+	secret, err := s.client.GetSecret(ctx, s.params.CiliumNamespace, defaults.EncryptionSecretName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch IPsec secret: %s", err)
+	}
+	return printIPsecKey(string(secret.Data["keys"]), s.params.Output)
```
derailed, line 26: nit: Just in case our mechanic changes it would be great with if we checked `keys` is in the map

derailed (changes requested):
@viktor-kurchenko Nice work Victor!

encrypt/ipsec_key_status.go
viktor-kurchenko: Nice catch, fixed! Thanks!

encrypt/ipsec_key_status.go
```diff
@@ -0,0 +1,48 @@
...
+
+	secret, err := s.client.GetSecret(ctx, s.params.CiliumNamespace, defaults.EncryptionSecretName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch IPsec secret: %s", err)
+	}
+
+	if key, ok := secret.Data["keys"]; ok {
+		return printIPsecKey(string(key), s.params.Output)
+	}
+	return fmt.Errorf("IPsec key not found in the secret: %s", defaults.EncryptionSecretName)
```
derailed, line 30: nit: key->keys

derailed (approved):
@viktor-kurchenko Thanks for the updates. Small pick on messaging

encrypt/ipsec_key_status.go
viktor-kurchenko: Fixed.

PR #2291, opened 2024-02-07
IPsec key rotation with algorithm change support.

viktor-kurchenko:


internal/cli/cmd/encrypt.go
```diff
@@ -57,14 +58,30 @@ func newCmdIPsecRotateKey() *cobra.Command {
...
+			}
 			s := encrypt.NewEncrypt(k8sClient, params)
 			if err := s.IPsecRotateKey(context.Background()); err != nil {
 				fatalf("Unable to rotate IPsec key: %s", err)
 			}
 			return nil
 		},
 	}
+	cmd.Flags().StringVarP(&params.IPsecKeyAuthAlgo, "auth-algo", "", "", "IPsec key authentication algorithm. One of: gcm-aes, hmac-md5, hmac-sha1, hmac-sha256, hmac-sha512")
+	cmd.Flags().StringVarP(&params.IPsecKeyPerNode, "key-per-node", "", "", "IPsec key per cluster node. One of: true, false")
```
jschwinger233, line 72: Having gone through the changes above, I believe empty values are also allowed for `--auth-algo` and `--key-per-node`: users don't have to specify these two flags if want to keep the same auth-algo and per-node settings as existing ones. Maybe worth mentioning in flag description that they are optional flags and empty values mean using the current settings.

jschwinger233 (approved):
Thanks!

internal/cli/cmd/encrypt.go
viktor-kurchenko: Makes sense! Added, thank you!

asauber (approved):

encrypt/ipsec_key_rotator.go
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package encrypt
+
+var rotators = map[string]func(key ipsecKey) (ipsecKey, error){
+	"":            func(key ipsecKey) (ipsecKey, error) { return key.rotate() },
+	"gcm-aes":     newGcmAesKey,
+	"hmac-md5":    newHmacMD5Key,
+	"hmac-sha1":   newHmacSHA1Key,
```
pchaigno, line 10: Are these two options really something we want to encourage using?

viktor-kurchenko: We just support them and it's up to the user which one should be used.
Do you think it's better to drop them?

pchaigno: What do you mean by "we support them"? Technically, we support all algorithms that Linux supports.

I'm mostly unsure we want to encourage MD5 and SHA1 when they are known broken in the general case (although maybe not in this specific authentication case).

viktor-kurchenko: Well, I put in the map all the algorithms that we used (during testing) for one of our Customers.
I'm happy to raise a PR and remove `hmac-md5` and `hmac-sha1`.

PR #2297, opened 2024-02-12
AWS OIDC instead of access key.

viktor-kurchenko:
Pass workflow examples:
- https://github.com/cilium/cilium-cli/actions/runs/7873661347/job/21481669941
- https://github.com/cilium/cilium-cli/actions/runs/7873661359/job/21481631164


brlbil (approved):
Failed AWS tests also work for creating the cluster.

PR #2300, opened 2024-02-13
Add sig/encryption as owner of /encrypt/ in CODEOWNERS

margamanterola:
With the new encryption commands added by Viktor, I believe we should add the relevant codeowners


pchaigno (approved):
Thanks!

tklauser:
CI failures are unrelated, this PR is only changing `CODEOWNERS`. Merging.

PR #2301, opened 2024-02-13
gha: increase GKE disk size in external workloads workflow to 15GB

giorio94:
Recently, I've been hitting [pretty consistent issues](https://github.com/cilium/cilium-cli/actions/runs/7884573763/job/21527869230?pr=2299) on this workflow caused by pods being evicted due to DiskPressure. Let's try increasing the disk size from 10GB to 15GB to see if the situation stabilizes.


giorio94:
The External Workloads workflow is green: https://github.com/cilium/cilium-cli/actions/runs/7898245112/job/21555270396?pr=2301
Dropping the temporary commit and marking ready for review.

brlbil (approved):
Thanks

PR #2306, opened 2024-02-15
Introduce host firewall connectivity tests

giorio94:
Introduce two new tests covering the host firewall functionality, i.e., asserting that both ingress and egress CiliumClusterwideNetworkPolicies specifying a NodeSelector correctly block the expected traffic. The
tests are executed only when the unsafe tests are enabled, as potentially disruptive if executed against a live cluster.

The initial five commits are refactor and generalize the logic in charge of parsing, registering and creating/updating the policies used during the tests, to uniform it, reduce code duplication, and in preparation for the support for CCNPs. The sixth commit introduces the actual support for CCNPs in connectivity test. The last commit introduces the actual tests.

Please review commit by commit, and refer to the commit messages for additional details.

### DIFF
--- a/connectivity/check/policy.go
+++ b/connectivity/check/policy.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


connectivity/check/policy.go
```diff
@@ -623,90 +665,52 @@ func (t *Test) CiliumLogs(ctx context.Context) {
 	}
 }
 
-// parseCiliumPolicyYAML decodes policy yaml into a slice of CiliumNetworkPolicies.
-func parseCiliumPolicyYAML(policy string) (cnps []*ciliumv2.CiliumNetworkPolicy, err error) {
-	if policy == "" {
+// ParsePolicyYAML decodes a yaml file into a slice of policies.
+func ParsePolicyYAML[T runtime.Object](input string, scheme *runtime.Scheme) (output []T, err error) {
```
viktor-kurchenko, line 669: Nice approach!

NIT: Why this function is public?

giorio94: I exported the generalized functions to make it possible to reuse them from external hooks as well.

viktor-kurchenko: Gotcha, thanks!

viktor-kurchenko (approved):
LGTM.

nathanjsweet (approved):

PR #2315, opened 2024-02-19
connectivity: improve reporting for no-unexpected-packet-drops check

giorio94:
Let's leverage the GenericAction construct to provide better progress and error reporting for the given test, displaying the node associated with any issue and ensuring that we continue checking for problems also after that a first one has been found. Let's also configure it to still capture a single sysdump regardless of the number of failures.


viktor-kurchenko (approved):
LGTM.

giorio94:
@danehans Gentle ping :pray:

tklauser (approved):

PR #2320, opened 2024-02-20
Ignore Cilium endpoints not found error for the status command.

viktor-kurchenko:
The fix is required for the Customers that have `disableEndpointCRD: "true"` in the Cilium configuration.

Status output example when `disableEndpointCRD: "true"`:
Screenshot 2024-02-20 at 11 29 00

The logic was stolen from: https://github.com/cilium/cilium-cli/blob/main/install/install.go#L582-L587

### DIFF
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -172,7 +172,11 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	ciliumEps, err := k.client.ListCiliumEndpoints(ctx, "", metav1.ListOptions{})
 	if err != nil {
-		return err
+		// When the CEP has not been registered yet, it's impossible
+		// for any pods to be managed by Cilium.
+		if err.Error() != "the server could not find the requested resource (get ciliumendpoints.cilium.io)" {
+			return err
+		}
 	}
 	if ciliumEps != nil {
 		numberCiliumPod = len(ciliumEps.Items)
--- a/status/k8s.go
+++ b/status/k8s.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"regexp"
 	"strings"
 	"sync"
@@ -170,11 +171,11 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 		}
 	}
 
+	// When the CEP has not been registered yet, it's impossible
+	// for any pods to be managed by Cilium. So continue, with 0 managed pods.
 	ciliumEps, err := k.client.ListCiliumEndpoints(ctx, "", metav1.ListOptions{})
 	if err != nil {
-		// When the CEP has not been registered yet, it's impossible
-		// for any pods to be managed by Cilium.
-		if err.Error() != "the server could not find the requested resource (get ciliumendpoints.cilium.io)" {
+		if err, ok := err.(*k8serrors.StatusError); ok && err.Status().Code != http.StatusNotFound {
 			return err
 		}
 	}


asauber (commented):
Is your screenshot with this fix or without? Is the intention to show 0 pods managed?

viktor-kurchenko:
> Is your screenshot with this fix or without? Is the intention to show 0 pods managed?

This is with fix. I think 0 pods is a side-effect of `disableEndpointCRD: "true"`.

margamanterola:
Some additional context on why this PR is needed:
https://github.com/cilium/cilium-cli/pull/2233 changed the exit code of the `status` subcommand so that if there was an error in any of the calls triggered to gather the status, the command finishes with error. However, when the cluster is configured with `disableEndpointCRD: "true"`, the "error" is expected and exiting with a non-zero value is actually incorrect.

A better solution would be for the CLI to support this option, but until we have that, this change allows the `status` subcommand to exit with zero when there aren't any actual errors.

status/k8s.go
```diff
@@ -172,7 +172,11 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	ciliumEps, err := k.client.ListCiliumEndpoints(ctx, "", metav1.ListOptions{})
 	if err != nil {
-		return err
+		// When the CEP has not been registered yet, it's impossible
+		// for any pods to be managed by Cilium.
+		if err.Error() != "the server could not find the requested resource (get ciliumendpoints.cilium.io)" {
```
asauber, line 177: Errors in Go should not be asserted by string matching, this is likely to lead to bugs when error strings change. Instead the `errors.As` (https://pkg.go.dev/errors#As) method can be used to search for a given error type in the wrapped error tree. 

In this case `http.StatusNotFound` is the error type that you want to assert with `errors.As`.

https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L446C7-L446C26

You can also move this predicate up into line 174 to avoid an extra level of indentation and simplify things. That is:

```
if err != nil and !errors.As(err, http.StatusNotFound) {
	return err
}
```

status/k8s.go
```diff
@@ -172,7 +172,11 @@ func (k *K8sStatusCollector) podCount(ctx context.Context, status *Status) error
 
 	ciliumEps, err := k.client.ListCiliumEndpoints(ctx, "", metav1.ListOptions{})
 	if err != nil {
-		return err
+		// When the CEP has not been registered yet, it's impossible
+		// for any pods to be managed by Cilium.
```
asauber, line 176: This comment could be expanded to explain what the intended resulting behavior is:

```
// When the CEP has not been registered yet, it's impossible
// for any pods to be managed by Cilium. So continue, with 0 managed pods.
```

If you collapse the predicates as suggested, then this comment block can be moved above line 174.

asauber (changes requested):
I see. I do think the overall `status` command should be changed to be made aware of `disableEndpointCRD`. 

It seems like the correct behavior would be to hide the `Cluster Pods:` section entirely. Is it true that Cilium is managing `"0/0"` Pods if Endpoints are disabled? I think yes.

margamanterola:
> I do think the overall status command should be changed to be made aware of disableEndpointCRD

Yes, but this is outside the scope of this PR, the goal of which is to make the status subcommand not return an error when nothing is wrong.

> It seems like the correct behavior would be to hide the Cluster Pods: section entirely. Is it true that Cilium is managing "0/0" Pods if Endpoints are disabled? I think yes.

I don't think so. I guess it depends on what is the expected meaning of "managed by Cilium". In setups that use kvstore mode it's common to disable the endpoint CRD to reduce unnecessary strain on the API server. From my point of view, this doesn't mean that the pods are not managed by Cilium, but only that there's no CiliumEndpoint CRD for those pods.

status/k8s.go
viktor-kurchenko: I fully agree that error string matching is a bad approach.
But I don't think we can use `errors.As` because we don't have error wrapping here.

So, I used type assertion and status code check.

asauber (approved):

asauber:
> I don't think so. I guess it depends on what is the expected meaning of "managed by Cilium". In setups that use kvstore mode it's common to disable the endpoint CRD to reduce unnecessary strain on the API server. From my point of view, this doesn't mean that the pods are not managed by Cilium, but only that there's no CiliumEndpoint CRD for those pods.

Understood, here's an Issue to track making this `status` output more accurate: https://github.com/cilium/cilium-cli/issues/2334

PR #2321, opened 2024-02-21
sniffer: improve handling of tcpdump termination

giorio94:
Let's improve the handling of tcpdump termination by the Sniff utility in two main directions:

* Don't return an error if tcpdump successfully exited before we determined it was ready. This can happen when running in sanity mode, as tcpdump exits after capturing a single packet matching the filter. Fixes the `Failed to execute tcpdump: %!w(<nil>)'` flake.

* Explicitly assert that tcpdump didn't terminate in advance when running in assert mode, as otherwise we may have missed packets.

### DIFF
--- a/utils/sniff/sniffer.go
+++ b/utils/sniff/sniffer.go
@@ -74,7 +74,7 @@ func Sniff(ctx context.Context, name string, target *check.Pod,
 			target.String(), target.NodeName(), mode, strings.Join(sniffer.cmd, " "))
 		err := target.K8sClient.ExecInPodWithWriters(ctx, cmdctx,
 			target.Pod.Namespace, target.Pod.Name, "", sniffer.cmd, &sniffer.stdout, io.Discard)
-		if err != nil && !errors.Is(err, context.Canceled) {
+		if err != nil {
 			sniffer.exited <- err
 		}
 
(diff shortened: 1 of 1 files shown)


giorio94:
AKS (BYOCNI) failure is unrelated. I'm stopping pressing the re-run button after hitting flakes five times in a row :sob:

asauber (approved):
Implementation looks good given the intention. I think it would be easier to read without the blank lines around line 166.

utils/sniff/sniffer.go
```diff
@@ -145,3 +152,35 @@ func (sniffer *Sniffer) Validate(ctx context.Context, a *check.Action) {
...
+		// at this point, as we may have missed packets. That's not the case in
+		// sanity mode, as configured to stop after receiving a single packet.
+		if err == nil && sniffer.mode == ModeAssert {
+			return errors.New("tcpdump terminated unexpectedly")
+		}
+
+		return err
+
+	default:
+		wctx, wcancel := context.WithTimeout(ctx, time.Second)
```
asauber, line 169: Does "w" stand for "wait"?

giorio94: Yep, that was the intention. I've just dropped it and aliased `ctx` though now, as it isn't used otherwise, and looks clearer.

PR #2322, opened 2024-02-21
Connectivity test factory component.

viktor-kurchenko:
Introducing connectivity tests factory component.

The PR goals are:
1. Extract connectivity test definitions in separate files.
2. Implement a factory component that is responsible for keeping test creation logic and order.
3. The first refactoring step related to the CFP: https://github.com/cilium/design-cfps/pull/15


connectivity/factory/factory.go
```diff
@@ -0,0 +1,244 @@
...
+	echoIngressFromCIDRYAML string
+)
+
+var (
+	clientLabel  = map[string]string{"name": "client"}
+	client2Label = map[string]string{"name": "client2"}
+
+	runAlways = func(semver.Version, check.Parameters) bool { return true }
+
+	parallelTests = []factory{
```
nickolaev, line 81: Please add a comment here that the order of these tests might be important.

viktor-kurchenko: Done.

tklauser (changes requested):
Could you please add entries to `CODEOWNERS` for the newly created files in `connectivity/facrtory` where applicable? Where the owner is not obvious, keeping @cilium/ci-structure is probably fine for now, but some tests are quite feature-specific so the relevant team should probably be in the loop for any future changes.

viktor-kurchenko:
> Could you please add entries to `CODEOWNERS` for the newly created files in `connectivity/facrtory` where applicable? Where the owner is not obvious, keeping @cilium/ci-structure is probably fine for now, but some tests are quite feature-specific so the relevant team should probably be in the loop for any future changes.

Done, thanks!

CODEOWNERS
```diff
@@ -31,6 +31,7 @@
 /connectivity/check/ipcache.go @cilium/ipcache
 /connectivity/check/metrics*.go @cilium/metrics
 /connectivity/check/policy.go @cilium/sig-policy
+/connectivity/factory/factory.go @cilium/ci-structure
```
nickolaev, line 34: I think @tklauser wanted to be more specific about the relevant tests. Like `/connectivity/factory/echo_ingress_knp.go`
I guess we can derive the names of the tests from git blame in the main branch.

viktor-kurchenko: Ah, my bad! Fixed.

tklauser: Yep, that's what I had in mind. Thanks for pointing it out @nickolaev and thanks for the fix @viktor-kurchenko.

tklauser: > I guess we can derive the names of the tests from git blame in the main branch.

Agree, I think we should already add the respective teams as owners where appropriate. Also, I think we don't need an entry for each file, we can just add `/connectivity/factory/* @cilium/ci-structure` as a catch-all for all files that are not covered by more specific entries.

tklauser: Both :) the catch all for any file that doesn't show a clear owner in `git blame`.

nickolaev: I think it's a good start though.

nickolaev (approved):

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+	//go:embed manifests/echo-ingress-from-cidr.yaml
+	echoIngressFromCIDRYAML string
+)
+
+var (
+	clientLabel  = map[string]string{"name": "client"}
+	client2Label = map[string]string{"name": "client2"}
+
+	// Test order is important
+	parallelTests = []factory{
```
asauber, line 80: Since this slice is called `parallelTests`, calling out that "order is important", is a bit strange for the reader here.

How about we call this `primaryTests` or similar until the parallelization is implemented?

viktor-kurchenko: Fixed.

michi-covalent:
i'll review this today

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+		toFqdns{},
+		podToControlplaneHost{},
+		podToK8sOnControlplane{},
+		podToControlplaneHostCidr{},
+		podToK8sOnControlplaneCidr{},
+	}
+)
+
+type factory interface {
+	name() string
```
michi-covalent, line 159: this method seems unnecessary if it's only called from build() 💭

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+	shouldRun(version semver.Version, params check.Parameters) bool
```
michi-covalent, line 160: `Test` struct already has `willRun` https://github.com/cilium/cilium-cli/blob/main/connectivity/check/test.go#L234 that checks for cilium version and such. it will be confusing to have both this method and willRUn.

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+	build(ct *check.ConnectivityTest, templates map[string]string)
```
michi-covalent, line 161: this is just a personal preference but i prefer the signature being:
```
build(templates map[string]string) *check.Test
```
building a test shouldn't depend on `ConnectivityTest`.

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+		dnsOnly{},
+		toFqdns{},
+		podToControlplaneHost{},
+		podToK8sOnControlplane{},
+		podToControlplaneHostCidr{},
+		podToK8sOnControlplaneCidr{},
+	}
+)
+
+type factory interface {
```
michi-covalent, line 158: totally subjective but `factory` might not be the right name here. maybe call it `Test` or `TestCase` or something?

michi-covalent (changes requested):
💯 for splitting tests into separate files 🥰 

at a very high level we need to think about how to express the order of tests. it will be difficult to maintain if we have a single `primaryTests`. we also need to think about tests being added from outside using the hook interface.

... stepping back even further, it doesn't have to be fixed in this PR but why does the order matter?

connectivity/factory/factory.go
nickolaev: @michi-covalent we are planning to expand on this a bit more. We had this session from @fgiloux during the hive time, where we discussed the possibilities to group the tests based on labeling them. Imagine have a label "single-node", or "clustermes" or whatever. Today this filtering is done based on naming assumptions, we want to change that and allow for more finer grained bisecting the whole test catalog.

connectivity/factory/factory.go
nickolaev: Oh we have enough things named after "test" here. And frankly at times it gets really confusing. I agree "factory" is not the best though, we can come up with a more appropriate name. @viktor-kurchenko came up with the factory pattern, that's why we have it here. I think we have changed the code in different direction so it's not really a factory anymore.

connectivity/factory/factory.go
viktor-kurchenko: Completely agree but unfortunately we already have this dependencies in expectations and scenarios:
- https://github.com/cilium/cilium-cli/blob/main/connectivity/suite.go#L352
- https://github.com/cilium/cilium-cli/blob/main/connectivity/suite.go#L1270
- https://github.com/cilium/cilium-cli/blob/main/connectivity/suite.go#L367

It should be improved but this PR is already large.
So, I would prefer to do it in follow up PRs.

connectivity/factory/factory.go
viktor-kurchenko: Maybe `builder`?

connectivity/factory/factory.go
viktor-kurchenko: Also, I don't want to do a logic refactoring in this PR. It's already big enough, isn't it?

connectivity/check/test.go
```diff
@@ -58,6 +58,29 @@ var (
 	egressGatewayPolicyYAML string
 )
 
+// NewTest factory function.
+func NewTest(name string, verbose bool, debug bool) *Test {
```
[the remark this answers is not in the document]
viktor-kurchenko: I don't have a strong opinion here )

connectivity/factory/factory.go
```diff
@@ -0,0 +1,248 @@
...
+
+	//go:embed manifests/echo-ingress-from-cidr.yaml
+	echoIngressFromCIDRYAML string
+)
+
+var (
+	clientLabel  = map[string]string{"name": "client"}
+	client2Label = map[string]string{"name": "client2"}
+
+	// Test order is important
```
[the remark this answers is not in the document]
viktor-kurchenko: My bad.
I put this comment for two reasons:
1. To highlight that we want to have a predictable test run order (which is already guaranteed by slice).
2. Because of comments like this: https://github.com/cilium/cilium-cli/blob/main/connectivity/suite.go#L1261

Perhaps it's better to remove this comment?

Also, as the next step(s) I was thinking about splitting all the tests into the following chunks:
1. `before`: tests that must be run before other tests (for some reason) and run only sequentially in a single namespace.
2. `parallel`: tests that can be run concurrently in separate namespaces.
4. `after`: tests that must be run before other tests (for some reason) and run only sequentially in a single namespace.

connectivity/factory/factory.go
viktor-kurchenko: Removed.

connectivity/factory/factory.go
michi-covalent: let's do `TestBuilder`

connectivity/check/test.go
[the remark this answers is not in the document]
michi-covalent: ok let's just keep one unless they are being used for different purposes 👍

connectivity/factory/factory.go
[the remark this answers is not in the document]
michi-covalent: yeah something along that line makes sense

connectivity/factory/factory.go
michi-covalent: ok yeah that's unfortunate. let's do a cleanup later

connectivity/factory/factory.go
michi-covalent: then i'd suggest doing necessary refactoring first in a separate PR. i think it's more natural for this `shouldRun` method to belong to `Test` instead of `TestBuilder` 💭

viktor-kurchenko: `factory` interface renamed to `testBuilder` and `factory` package renamed to `builder`.

viktor-kurchenko: I've removed `shouldRun` method and added `WithConditionFn` method to the Test struct.

connectivity/factory/factory.go
viktor-kurchenko: Done.

connectivity/check/test.go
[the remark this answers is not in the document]
viktor-kurchenko: Do you mean to drop function parameter or cli param at all?
@michi-covalent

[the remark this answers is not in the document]
michi-covalent: let's keep it as is for now, we can change it in a subsequent pr. initially i thought this pull request was adding those params.

connectivity/check/test.go
```diff
@@ -390,6 +420,13 @@ func configureNamespaceInPolicySpec(spec *api.Rule, namespace string) {
 	}
 }
 
+// WithConditionFn takes a function containing condition check logic that
+// decides if the test needs to be run.
+func (t *Test) WithConditionFn(fn func() bool) *Test {
```
michi-covalent, line 425: let's call it `WithCondition`:
```suggestion
// WithCondition takes a function containing condition check logic that
// returns true if the test needs to be run, and false otherwise.
func (t *Test) WithCondition(fn func() bool) *Test {
```

connectivity/check/test.go
```diff
@@ -129,6 +153,8 @@ type Test struct {
 	warnBuf *bytes.Buffer
 	verbose bool
 
+	conditionFn func() bool
```
michi-covalent, line 156:
```suggestion
        // conditionFn is a function that returns true if the test needs to run,
        // and false otherwise. By default, it's set to a function that returns
        // true.
	conditionFn func() bool
```

connectivity/factory/factory.go
michi-covalent: i love it

michi-covalent (approved):
thanks viktor it looks much better now. i have a couple of minor nits but i'm approving it now. feel free to merge it once these nits get addressed 🚀🙏

connectivity/factory/factory.go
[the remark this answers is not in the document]
michi-covalent: could you open a github issue to follow up on this so that we don't forget about it 🚀 🙏

connectivity/check/test.go
viktor-kurchenko: Fixed.

connectivity/check/test.go
viktor-kurchenko: Fixed.

CODEOWNERS
```diff
@@ -31,6 +31,85 @@
...
+/connectivity/builder/pod_to_ingress_service_deny_backend_service.go @cilium/ci-structure
+/connectivity/builder/pod_to_ingress_service_deny_ingress_identity.go @cilium/ci-structure
+/connectivity/builder/pod_to_k8s_on_controlplane.go @cilium/ci-structure
+/connectivity/builder/pod_to_k8s_on_controlplane_cidr.go @cilium/ci-structure
+/connectivity/builder/pod_to_node_cidrpolicy.go @cilium/ci-structure
+/connectivity/builder/pod_to_pod_encryption.go @cilium/sig-encryption
+/connectivity/builder/to_cidr_external.go @cilium/ci-structure
+/connectivity/builder/to_cidr_external_knp.go @cilium/ci-structure
+/connectivity/builder/to_entities_world.go @cilium/ci-structure
+/connectivity/builder/to_fqdns.go @cilium/ci-structure
```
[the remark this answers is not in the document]
viktor-kurchenko: Done.

nickolaev (approved):

connectivity/factory/factory.go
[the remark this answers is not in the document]
viktor-kurchenko: I have a roadmap issue in the private repo. Should I replicate it to the `cilium/cilium-cli` repo?

CODEOWNERS
```diff
@@ -31,6 +31,26 @@
...
+/connectivity/builder/echo_ingress.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_auth_always_fail.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_from_other_client_deny.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_from_outside.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_knp.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_l7.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_l7_named_port.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_mutual_auth_spiffe.go @cilium/sig-encryption
+/connectivity/builder/egress_gateway.go @cilium/sig-clustermesh
+/connectivity/builder/egress_gateway_excluded_cidrs.go @cilium/sig-clustermesh
```
tklauser, line 47: These should probably be owned by @cilium/egress-gateway.

CODEOWNERS
```diff
@@ -31,6 +31,26 @@
...
+/connectivity/builder/health.go @cilium/sig-clustermesh
+/connectivity/builder/host_entity_egress.go @cilium/sig-clustermesh
+/connectivity/builder/host_entity_ingress.go @cilium/sig-clustermesh
```
tklauser, line 50: Not sure why these are assigned to @cilium/sig-clustermesh. The features they're testing don't seem related. Probably best to drop these entries and let @cilium/ci-structure take care of them for now.

CODEOWNERS
```diff
@@ -31,6 +31,26 @@
...
+/connectivity/builder/cluster_entity_multi_cluster.go @cilium/sig-clustermesh
+/connectivity/builder/dns_only.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_auth_always_fail.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_from_other_client_deny.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_from_outside.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_knp.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_l7.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_l7_named_port.go @cilium/sig-clustermesh
+/connectivity/builder/echo_ingress_mutual_auth_spiffe.go @cilium/sig-encryption
```
tklauser, line 45: Based on the features these are testing, I think they should be assigned to @cilium/sig-servicemesh instead.

tklauser (changes requested):
Code changes LGTM.

Some suggestions for `CODEOWNERS` inline. Please also squash your commits into a single one.

CODEOWNERS
viktor-kurchenko: Fixed.

CODEOWNERS
viktor-kurchenko: I just deleted them because `@cilium/ci-structure` is the default for the package.

CODEOWNERS
viktor-kurchenko: Fixed, thanks!

viktor-kurchenko:
> Some suggestions for `CODEOWNERS` inline. Please also squash your commits into a single one.

Commits squashed, thank you!

tklauser (approved):
LGTM, thanks Viktor!

Issue #2327, opened 2024-02-25
delete classic mode


michi-covalent:
delete classic mode to get ready for v0.16.

PR #2328, opened 2024-02-25
Delete classic mode install/uninstall/upgrade commands

michi-covalent:
Ref: #2327


internal/cli/cmd/cmd.go
```diff
@@ -88,20 +87,10 @@ cilium connectivity test`,
 		newCmdStatus(),
 		newCmdSysdump(hooks),
 		newCmdVersion(),
+		newCmdInstallWithHelm(),
+		newCmdUninstallWithHelm(),
+		newCmdUpgradeWithHelm(),
 	)
-	if utils.IsInHelmMode() {
```
gandro, line 92: I think we need to update the README as well? https://github.com/cilium/cilium-cli?tab=readme-ov-file#helm-installation-mode

gandro (changes requested):
Overall this looks good to me, modulo some doc changes.

Would it help to do a cilium/cilium CI run with a version of cilium-cli with this patch applied? Or are we certain that the classic mode is not used anywhere?

internal/cli/cmd/cmd.go
michi-covalent: yeah i was going to update README.md in a subsequent PR after removing the classic mode from all the commands but i don't mind updating it in this PR.

i'm pretty sure the classic mode is not being used in cilium/cilium CI anymore, but i'll run the CI to make sure 🚀 🙏

michi-covalent:
running the CI here => https://github.com/cilium/cilium/pull/30978

PR #2331, opened 2024-02-26
Refactor: Move junit package under connectivity package

michi-covalent:
The junit package is only used by the connectivity package, so it can be under /connectivity/internal/ instead of /internal/.


brlbil (approved):

derailed (approved):
@michi-covalent LGTM

PR #2332, opened 2024-02-26
k8s: don't drop stdout of kubectl exec

bimmlerd:
A non-empty stderr is interpreted as a command failure. However, calling code still assumes the output it gets is the output of the command, whereas we currently just return a pristine buffer.

Fix it by returning the buffer.

An example of why this is annoying: https://github.com/cilium/cilium/actions/runs/8043919487/job/21966754656#step:17:328


michi-covalent (approved):
😇

joamaki (approved):

PR #2333, opened 2024-02-26
connectivity: fix encryption validation with wireguard encap + host fw

giorio94:
In Cilium v1.15 and later (as well as v1.14 if the --wireguard-encapsulate flag is set), when both tunnel encapsulation and WireGuard encryption are enabled, packets are first encapsulated and then encrypted. Let's fix the connectivity check inter-node interface determination when host firewall is additionally enabled, as currently always defaulting to the tunnel interface, which is incorrect (we would flag the packets as unencrypted, even though the encapsulated packets are encrypted afterwards).


brb (approved):
Thanks!

PR #2336, opened 2024-02-27
connectivity: Remove --delete-cilium-pod-on-nodes flag

michi-covalent:
The flag was added in #1131 but it looks like we ended up not using it. I grepped for 'delete-cilium-pod-on-nodes' in both cilium and cilium-cli repositories but there was no match.


asauber (approved):

christarazi (approved):

michi-covalent:
alright i'm going in. cc @tklauser just a heads up

PR #2337, opened 2024-02-27
docs: Update version matrix

sayboras:
This is to address the below comment.

https://github.com/cilium/cilium-cli/pull/2285#issuecomment-1926303693

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,13 @@ binary releases.
 
 | Release                                                                | Maintained | Compatible Cilium Versions |
 |------------------------------------------------------------------------|------------|----------------------------|
-| [v0.15.23](https://github.com/cilium/cilium-cli/releases/tag/v0.15.23) | Yes        | Cilium 1.14 and newer      |
+| [v0.15.23](https://github.com/cilium/cilium-cli/releases/tag/v0.15.23) | Yes        | Cilium 1.15 and newer      |
+| [v0.15.22](https://github.com/cilium/cilium-cli/releases/tag/v0.15.22) | Yes        | Cilium 1.14 (*)            |
 | [v0.14.8](https://github.com/cilium/cilium-cli/releases/tag/v0.14.8)   | Yes        | Cilium 1.13                |
+
+Note: 
+- [v0.15.22](https://github.com/cilium/cilium-cli/releases/tag/v0.15.22) can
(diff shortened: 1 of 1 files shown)


sayboras:
CI failure is not related for this doc change.

README.md
```diff
@@ -43,7 +43,8 @@ binary releases.
 
 | Release                                                                | Maintained | Compatible Cilium Versions |
 |------------------------------------------------------------------------|------------|----------------------------|
-| [v0.15.23](https://github.com/cilium/cilium-cli/releases/tag/v0.15.23) | Yes        | Cilium 1.14 and newer      |
+| [v0.15.23](https://github.com/cilium/cilium-cli/releases/tag/v0.15.23) | Yes        | Cilium 1.15 and newer      |
+| [v0.15.22](https://github.com/cilium/cilium-cli/releases/tag/v0.15.22) | Yes        | Cilium 1.14                |
 | [v0.14.8](https://github.com/cilium/cilium-cli/releases/tag/v0.14.8)   | Yes        | Cilium 1.13                |
```
michi-covalent, line 49: could you maybe add a sentence or two here and describe why `>=v0.15.23` is incompatible with cilium v1.14?

sayboras: here is the reason https://github.com/cilium/cilium-cli/pull/2285#issuecomment-1926303693.

Do you want to mention in the docs as well ?

michi-covalent: yeah i think it makes sense to add it in README.md

sayboras: ah right, I have added a small footnote for v0.15.22.

michi-covalent (approved):

tklauser (commented):
@sayboras this change is a bit unfortunate because we try to support all currently supported stable releases on the latest cilium-cli release. If I read the original change https://github.com/cilium/cilium-cli/pull/2285#issuecomment-1926303693 correctly, couldn't we instead just also capture `v1beta1` version of the changed resources in addition to the `v1` ones so we could at least still support Cilium 1.14 and Cilium 1.15 in the latest cilium-cli version.

sayboras:
> couldn't we instead just also capture v1beta1 version of the changed resources in addition to the v1 ones so we could at least still support Cilium 1.14 and Cilium 1.15 in the latest cilium-cli version.

Yes, we can. There might be the case that we missed some attributes (i.e. available in v1, but not in v1beta1) in the future.

PR #2339, opened 2024-02-27
sniff: avoid race condition in tcpdump execution

bimmlerd:
It is possible, and in sanity check mode even likely, that the tcpdump invocation terminates successfully before Sniff returns. Current code handels this badly, as a race condition occurs. Depending on who wins, we either continue normally, or fatally log 'Failed to execute tcpdump: %!w(<nil>)'.


bimmlerd:
Oh nevermind this was fixed in #2321 :grin:

PR #2343, opened 2024-02-27
Remove unused consts

michi-covalent:


michi-covalent:
why doesn't golangci-lint detect these unused consts? 💭

PR #2344, opened 2024-02-28
uninstall: when --wait is specified, use foreground deletion.

tommyp1ckles:
By default, the helm libraries will use background cascading delete which means the call to do helm uninstall returns following the deployment being removed.

This means that running workloads, such as hubble-relay, may continue to be in the terminating state following `cilium uninstall --wait` exiting.

We depend on this behavior in CI E2E to clean up and reuse clusters for testing Cilium in different configurations.

In flakes such as: https://github.com/cilium/cilium/issues/30993 it seems like the old Hubble Pods are bleeding into the "fresh" install. These should be harmless, however this is triggering failures of the [no-error-logs] assertion in the following connectivity tests.

This change will provide a more thorough uninstall procedure in this case.


michi-covalent (approved):
looks innocent enough

christarazi (approved):

derailed (approved):
@tommyp1ckles Nice work!

michi-covalent (approved):

PR #2360, opened 2024-03-04
vendor: Pick up cilium/cilium v1.16.0-pre.0

michi-covalent:


go.mod
```diff
@@ -14,7 +14,7 @@ replace (
 require (
```
rolinh, line 14: I can't comment there but the replace directive for controller-tools is now `v0.8.0-2` and should be updated accordingly.

rolinh (changes requested):

go.mod
michi-covalent: oh yeah nice catch, updating the PR now 🚀 🙏

rolinh (approved):

PR #2361, opened 2024-03-05
sysdump: Export Hooks and InitSysdumpFlags

michi-covalent:
Move SysdumpHooks and InitSysdumpFlags to the sysdump package so that they can be imported from outside the cilium-cli repo.

Ref: cilium/design-cfps#9


sayboras (approved):
Looks 😇

PR #2363, opened 2024-03-05
Add error log check exception for leader election errors.

tommyp1ckles:
Add error log check exception for leader election errors.

    The operator uses the client-go leader election sdk, which reattempts acquiring the lock, logging failures via klog [1].

    [1] https://github.com/cilium/cilium/blob/6a70af21c3b6985096fca3f8240139b548ac4760/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go#L332

    In general, such errors should be transient, if it is persistent then it wouldn't be isolated to just this api so it should be ok to ignore all such errors.

    Signed-off-by: Tom Hadlaw <tom.hadlaw@isovalent.com>


meyskens (approved):
LGTM for me

michi-covalent (approved):

PR #2372, opened 2024-03-06
Pr/tp/fqdn datapath error exception

tommyp1ckles:


joestringer:
IMO this is most likely triggered by some underlying bug or instability in the environment that the testsuites are not accounting for.

Any time we extend the set of "errors to ignore", I get nervous because those are problems that our users may face in real environments, and we may be ignoring legitimate signal of a bug in CI and just passing it on to users. The background to the "X must be investigated" check is to try to motivate investigation of underlying bugs rather than ignore them until users find & report them.

squeed:
I've done a lot of investigation as to why this error message pops up. Most of the time, it's Envoy just taking a nap. I've not seen it take much longer than 150ms, which I suppose is the result of the current CPU environment we have. However, it can also indicate a deadlock or bug that is being papered over by a circuit breaker.

So, Joe is right, we should not ignore this error message if possible.

I'd like to change the default timeout in our test environments to 200ms, but I haven't had the time to look in to it.

tommyp1ckles:
@squeed Anything specific about 200ms vs 150?

Looks like this can be configured via: `tofqdns-proxy-response-max-delay`

PR #2375, opened 2024-03-07
Store namespace and k8sClient in the command context

michi-covalent:
Store namespace and k8sClient as command context values so that the connectivity package can access them from outside the cli package once it moves to cilium/cilium repo.

Ref: cilium/design-cfps#9

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package api
+
+import (
+	"context"
+
+	"github.com/cilium/cilium-cli/k8s"
+)
+
+// key is a context.Context value key type. It's unexported to avoid key collisions.
+type key int
+
+const (
+	namespaceKey key = iota
+	k8sClientKey
+)
+
+// SetNamespaceContextValue stores a namespace string as a context value and
+// returns the resulting context. You can access the namespace by calling
+// GetNamespaceContextValue.
+func SetNamespaceContextValue(ctx context.Context, namespace string) context.Context {
+	return context.WithValue(ctx, namespaceKey, namespace)
+}
+
+// SetK8sClientContextValue stores a namespace string as a context value and
+// returns the resulting context. You can access the namespace by calling
+// GetK8sClientContextValue.
+func SetK8sClientContextValue(ctx context.Context, client *k8s.Client) context.Context {
+	return context.WithValue(ctx, k8sClientKey, client)
+}
+
+// GetNamespaceContextValue retrieves the namespace from a context that was
+// stored by SetNamespaceContextValue.
+func GetNamespaceContextValue(ctx context.Context) (string, bool) {
+	namespace, ok := ctx.Value(namespaceKey).(string)
+	return namespace, ok
+}
+
+// GetK8sClientContextValue retrieves the k8s.Client from a context that was
+// stored by SetK8sClientContextValue.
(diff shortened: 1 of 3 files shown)


michi-covalent:
@tklauser what do you think about this change? simply exporting namespace and k8sClient fields doesn't work because it creates a circular dependency between https://github.com/cilium/cilium-cli/blob/pr/michi/majestic-ketchup/cli/cmd.go and https://github.com/cilium/cilium/blob/pr/michi/majestic-ketchup/pkg/cilium-cli/cmd/connectivity.go 💭

squeed:
If this is the direction we do decide to go, I'd like to see typed getters for these. (I'm missing a lot of context though).

defaults/defaults.go
```diff
@@ -162,3 +162,9 @@ var (
 	// CLIVersion is the software version of the Cilium CLI.
 	CLIVersion string
 )
+
+// NamespaceKey is the command context key for namespace.
+type NamespaceKey struct{}
+
+// K8sClientKey is the command context key for k8sClient.
+type K8sClientKey struct{}
```
tklauser, line 170: Like Casey already mentioned, I also wonder why these aren't typed. Why not:

```suggestion
// NamespaceKey is the command context key for namespace.
type NamespaceKey string

// K8sClientKey is the command context key for k8sClient.
type K8sClientKey string
```

But I might be missing something obvious.

michi-covalent: golangci-lint complains about using string as the context key 💭 

```
context-keys-type: should not use basic type string as key in context.WithValue (revive)
```

tklauser: Hm :thinking: interesting that the linter doesn't explain why we shouldn't use basic types :disappointed:

tklauser: Also interesting that the official Go doc example for `context.Context.WithValue` uses a type with underlying type `string`: https://pkg.go.dev/context#example-WithValue

michi-covalent: yeah https://go.dev/blog/context has some explanation, it's to avoid key collisions. i can do something similar 💭

michi-covalent: maybe something like this? https://gist.github.com/michi-covalent/a1ef2fa24a2b416670a5b5e97b436392

tobias how do you feel about adding a new `api` package in cilium-cli? i can't put this in `cli` package because it creates a circular dependency between https://github.com/cilium/cilium-cli/blob/pr/michi/majestic-ketchup/cli/cmd.go and https://github.com/cilium/cilium/blob/pr/michi/majestic-ketchup/pkg/cilium-cli/cmd/connectivity.go

tklauser: > maybe something like this? https://gist.github.com/michi-covalent/a1ef2fa24a2b416670a5b5e97b436392

LGTM. I guess https://gist.github.com/michi-covalent/a1ef2fa24a2b416670a5b5e97b436392#file-gistfile1-txt-L24 should usbe `k8sClientKey` instead of `namespaceKey` though.

> tobias how do you feel about adding a new `api` package in cilium-cli? i can't put this in `cli` package because it creates a circular dependency between https://github.com/cilium/cilium-cli/blob/pr/michi/majestic-ketchup/cli/cmd.go and https://github.com/cilium/cilium/blob/pr/michi/majestic-ketchup/pkg/cilium-cli/cmd/connectivity.go

Let's do it :rocket:

michi-covalent: updated the PR 🚀 🙏

PR #2384, opened 2024-03-11
BPF Masquerade test

tommyp1ckles:
Adds test coverage for bpf-masquerade (as well as any other IP masquerading mode) via connectivity tests. Currently this is done via Ginkgo runtime tests in cilium/cilium.

To accommodate this, and to improve overall test debugging this also replaces the normal json-mock based echo server implementation, used by connectivity tests, with a Go based implementation.

Currently, the node based json-mock server is used for creating test servers for running various tests.
Unfortunately this has some shortcomings, namely:

* Mock image is built outside of CLI repo thus making it difficult to make changes to.
* JSON Mock is designed to provide a server that returns some static schema of data. Cilium connectivity tests don't really this functionality, (we only really need to define some extra http routes to test route based policy). Instead a more flexible implementation would make it easier to adapt to future use cases.
* JSON Mock doesn't log on the index server as it's not really part of the schema. Better logging from the servers would be useful for diagosing test l7 failures/flakes (ex. https://github.com/cilium/cilium/issues/27762) as it would let us correlate timestamped events on the server side.

This adds a new standalone server called "echoserver" which is a small Go http server with the following routes:

GET / -> Index route, used for general l4/l7 connectivity testing.
GET /private -> Private route, used for l7 network policy.
GET /public -> Private route, used for l7 network policy.
GET /echo -> Responds with a json payload the remote address (i.e. <ip>:<client_port>) and timestamp.

These routes will cover existing test cases, with the latter being used to implement a bpf masquerade connectivity test.

### DIFF
--- a/Dockerfile.echo
+++ b/Dockerfile.echo
@@ -1,0 +1,5 @@
+FROM alpine:latest
+
+ADD ./echoserver /echoserver
+
+CMD /echoserver
--- a/Makefile
+++ b/Makefile
@@ -99,4 +99,11 @@ check:
 	docker run --rm -v `pwd`:/app -w /app docker.io/golangci/golangci-lint:$(GOLANGCILINT_WANT_VERSION) golangci-lint run
 endif
 
-.PHONY: $(TARGET) release local-release install clean test bench check clean-tags tags
+echoserver:
+	GOOS=linux $(GO_BUILD) -o echoserver ./echoserver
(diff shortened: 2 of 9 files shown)


brb:
Neat! We could extend this test to cover not only BPF masq, but also the connectivity to outside. Currently, by default, we use 1.1.1.1 as the endpoint to outside, which is flaky. We could use that echo server as the endpoint.

PR #2365, opened 2024-03-05
Improve the helm values autodetection logic

giorio94:
Please review commit by commit, and refer to the individual messages for additional details.

* Remove unused fields/parameters from the `getHelmValues` logic;
* Drop support for EOL Cilium v1.11 and earlier in helm values generation;  
* Drop broken `bpf.masquerade` autodetection logic;
* Fix retrieval of the user-configured routing mode, and uniform retrieval of nested helm value;
* Improve KPR autodetection logic to not override user configuration, and not configure deprecated values;
* Fix KPR probing in connectivity tests to account for both legacy and current values.
* Replace the remaining references to deprecated KPR settings in GHA workflows.

PR #2385, opened 2024-03-11
cleanup: Avoid using deprecated KPR values

sayboras:
This commit is to make sure that we don't use KPR deprecated values (e.g. partial, probe, strict, disabled) in favour of just 'true' and 'false'.

Relates: https://github.com/cilium/cilium/pull/31286


sayboras:
Closed in favour of https://github.com/cilium/cilium-cli/pull/2365

PR #2397, opened 2024-03-13
Add InitializeCommand hook

michi-covalent:
Call InitializeCommand hook with the root command before returning from cli.NewCiliumCommand.


asauber (approved):

derailed (approved):
@michi-covalent LGTM

PR #2400, opened 2024-03-13
refactor: Move Hooks to the api package

michi-covalent:
Move Hooks to the api package so that the connectivity package can use the Hooks without creating a circular dependency once it moves to cilium/cilium repo.

Ref: https://github.com/cilium/design-cfps/pull/9


asauber (approved):
LGTM

PR #2408, opened 2024-03-14
renovate: Don't pick up new Go patch versions in go.mod

michi-covalent:
Let's not pick up new Go patch versions in go.mod to be consistent with cilium/cilium. https://github.com/cilium/cilium/pull/28686 has more context about the rationale.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -81,6 +81,25 @@
       ],
     },
     {
+      // Avoid updating patch releases of golang in go.mod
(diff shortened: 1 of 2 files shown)


michi-covalent:
cc @joestringer @tklauser just an fyi

sayboras (approved):
LGTM, yeah, we have done the same in cilium/proxy (e.g. only 1.x but not 1.x.y)

tklauser (approved):

PR #2412, opened 2024-03-14
gha: increase Multicluster timeout

aspsk:
Recently the Multicluster workflow started to timeout pretty often.
Try to increase the timeout and see if the problem re-appears.


michi-covalent (approved):

michi-covalent:
ok i'll re-enable the multicluster workflow. let's see what happens.

michi-covalent:
let me close and re-open to trigger multicluster

PR #2418, opened 2024-03-16
refactor: Move logging outside internal

michi-covalent:
Move the logging package out of internal/ so that it can be imported from outside cilium/cilium-cli repo. This is useful for building the cilium CLI binary from cilium/cilium repo [^1].

[^1]: cilium/design-cfps#9


asauber (changes requested):
Any reason to not move this into /pkg ?

michi-covalent:
> Any reason to not move this into /pkg ?

there is no `pkg/` directory right now. not sure if it's worth adding it at this point 💭 i can go either way though.

asauber (approved):

PR #2427, opened 2024-03-18
Explicitly configure the "nodes without cilium" affinity on both initial installation and subsequent upgrades 

giorio94:
Explicitly configure the appropriate affinities to not schedule Cilium components on the nodes labeled as `cilium.io/no-schedule` during the preinstall phase, which is executed both during the initial installation and on subsequent upgrades. This change removes the side-effect from the getHelmValues method, which is no longer executed on upgrade after a recent refactoring. Additionally, let's convert the `--nodes-without-cilium` flag to boolean, as the values are always ignored.

### DIFF
--- a/cli/install.go
+++ b/cli/install.go
@@ -26,7 +26,7 @@ func addCommonInstallFlags(cmd *cobra.Command, params *install.Parameters) {
 	cmd.Flags().StringVar(&params.Version, "version", defaults.Version, "Cilium version to install")
 	cmd.Flags().StringVar(&params.DatapathMode, "datapath-mode", "", "Datapath mode to use { tunnel | native | aws-eni | gke | azure | aks-byocni } (default: autodetected).")
 	cmd.Flags().BoolVar(&params.ListVersions, "list-versions", false, "List all the available versions without actually installing")
-	cmd.Flags().StringSliceVar(&params.NodesWithoutCilium, "nodes-without-cilium", []string{}, "List of node names on which Cilium will not be installed. In Helm installation mode, it's assumed that the no-schedule node labels are present and that the infrastructure has set up routing on these nodes to provide connectivity within the Cilium cluster.")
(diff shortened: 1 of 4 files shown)


giorio94:
> Additionally testing in [cilium/cilium#30604](https://github.com/cilium/cilium/pull/30604)

CI is green there, the only ci-ipsec-upgrade failure is unrelated. Marking ready for review.

gandro (approved):

cli/install.go
```diff
@@ -26,7 +26,7 @@ func addCommonInstallFlags(cmd *cobra.Command, params *install.Parameters) {
 	cmd.Flags().StringVar(&params.Version, "version", defaults.Version, "Cilium version to install")
 	cmd.Flags().StringVar(&params.DatapathMode, "datapath-mode", "", "Datapath mode to use { tunnel | native | aws-eni | gke | azure | aks-byocni } (default: autodetected).")
 	cmd.Flags().BoolVar(&params.ListVersions, "list-versions", false, "List all the available versions without actually installing")
-	cmd.Flags().StringSliceVar(&params.NodesWithoutCilium, "nodes-without-cilium", []string{}, "List of node names on which Cilium will not be installed. In Helm installation mode, it's assumed that the no-schedule node labels are present and that the infrastructure has set up routing on these nodes to provide connectivity within the Cilium cluster.")
+	cmd.Flags().BoolVar(&params.NodesWithoutCilium, "nodes-without-cilium", false, "Configure the affinities to avoid scheduling Cilium components on nodes labeled with cilium.io/no-schedule. It is assumed that the infrastructure has set up routing on these nodes to provide connectivity within the Cilium cluster.")
```
michi-covalent, line 29: do we even need this flag at this point? 💭

michi-covalent (approved):
🚀 🙏

cli/install.go
giorio94: Yes, because we configure additional affinity rules which are not required outside of CI.

PR #2431, opened 2024-03-19
connectivity: enable IPv6 ns-lb-with-l7-policy on recent 1.13 release

julianwiedmann:
https://github.com/cilium/cilium-cli/pull/1629 initially limited the IPv6 test to v1.14, as v1.13 was missing some functionality.

But now that v1.13 has https://github.com/cilium/cilium/pull/31161, the IPv6 path should actually work. So let's extend the test coverage.


julianwiedmann:
Leaving this as draft, as Cilium doesn't pull the latest CLI into its v1.13 CI.

PR #2437, opened 2024-03-20
Include Makefile.override if present

michi-covalent:
Include Makefile.override if the file exists to allow overriding Make variables without modifying Makefile. Copied from cilium/cilium#26159.


michi-covalent:
moving this over to https://github.com/cilium/cilium-cli/pull/2441. closing 🚀 🙏

PR #2438, opened 2024-03-20
errors: add [no-error-logs] exception for issue caused by fqdn bug.

tommyp1ckles:
FQDN Proxy together running with Wireguard has an issue where proxied dns requests can cause a port collision. See: https://github.com/cilium/cilium/issues/31535

Despite this, the actual fqdn tests seem to be resilient to this, likely due to retries.

While we work on a fix for this issue, let's add a temporary exception here.

### DIFF
--- a/connectivity/tests/errors.go
+++ b/connectivity/tests/errors.go
@@ -6,6 +6,7 @@ package tests
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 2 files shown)


connectivity/tests/errors.go
```diff
@@ -251,4 +251,10 @@ const (
 	failedToReleaseLock    = "Failed to release lock:"
 	previouslyUsedCIDR     = "Unable to find identity of previously used CIDR"                           // from https://github.com/cilium/cilium/issues/26881
 	klogLeaderElectionFail = "error retrieving resource lock kube-system/cilium-operator-resource-lock:" // from: https://github.com/cilium/cilium/issues/31050
+
+	// knownBugWireguardCollision is for a known issue: https://github.com/cilium/cilium/issues/31535.
+	// In spite of this occurrence, fqdn connectivity tests still pass thus it should be ok to ignore these for a while
+	// while we fix this issue.
+	// TODO: Remove this after: #31535 has been fixed.
+	knownBugWireguardCollision = "bind: address already in use" // from: https://github.com/cilium/cilium/issues/30901
```
gandro, line 259: nit: Assuming this is not just a WireGuard problem, can we rename to `knownBugTransparentProxyPortCollision`?

I'd also prefer if we somehow checked for `Cannot forward proxied DNS lookup` too. `"address already in use" might be an error issues by other subsystems where it is probably more fatal

tommyp1ckles: Good point, I was working with the limitation that this has to be a single string.  We may want to support regex based matches for this.

tommyp1ckles: Kind of tangential, we should probably be able to override these exceptions with a flag rather than needing to open cli PRs...

tommyp1ckles: > nit: Assuming this is not just a WireGuard problem, can we rename to knownBugTransparentProxyPortCollision?

Left this as as wireguard as I reworked this to use a regex to capture a more specific exception.
If we start to see similar collision errors, but not related to the WG binding I think we should expose that for the time being.

gandro (approved):
Nice, thanks!

connectivity/tests/errors.go
```diff
@@ -20,29 +21,47 @@ import (
...
+func (s stringMatcher) IsMatch(log string) bool {
+	return strings.Contains(log, string(s))
+}
+
+type regexMatcher struct {
+	*regexp.Regexp
+}
+
+func (r regexMatcher) IsMatch(log string) bool {
+	return r.Regexp.Match([]byte(log))
```
tklauser, line 39: nit: could use `regexp.(*Regexp).MatchString` here:

```suggestion
	return r.Regexp.MatchString(log)
```

tklauser (approved):

PR #2439, opened 2024-03-21
defaults: Move Version and HelmRepository from const to var

michi-covalent:
Make Version and HelmRepository variables instead of consts so that they can be overridden using build tags.

### DIFF
--- a/defaults/defaults.go
+++ b/defaults/defaults.go
@@ -8,9 +8,6 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


michi-covalent:
chatted a bit with @tklauser about this offline. we decided to use `go build -X` flag to override consts as needed. closing this issue, i'll do some cleanup in https://github.com/cilium/cilium-cli/pull/2437 to make it easier to override `-X` flag 🎏

PR #2445, opened 2024-03-22
Explicitly wait for ingress service creation

giorio94:
Rather than just listing the services, let's explicitly wait until we successfully retrieve all expected services, so that we can provide a clear error in case something went wrong and they were not created.

### DIFF
--- a/connectivity/check/wait.go
+++ b/connectivity/check/wait.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/cilium/cilium/api/v1/models"
 	"golang.org/x/exp/slices"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/cilium-cli/defaults"
 	"github.com/cilium/cilium-cli/k8s"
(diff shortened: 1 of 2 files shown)


connectivity/check/wait.go
```diff
@@ -134,6 +135,29 @@ func WaitForCoreDNS(ctx context.Context, log Logger, client Pod) error {
...
+	for {
+		svc, err := client.GetService(ctx, namespace, name, metav1.GetOptions{})
+		if err == nil {
+			return Service{Service: svc.DeepCopy()}, nil
+		}
+
+		log.Debugf("[%s] Failed to retrieve Service %s/%s: %s", client.ClusterName(), namespace, name, err)
+
+		select {
+		case <-time.After(PollInterval):
```
tommyp1ckles, line 153: Might be better to use inctimer here

tommyp1ckles (approved):

connectivity/check/wait.go
giorio94: Hmm, I'd personally keep `time.After` for consistency with all other usages (`inctimer` doesn't seem to be used at all in Cilium CLI), and a global change would be out of scope here. It shouldn't matter much in terms of performance, as the CLI is not a long-running application as the Cilium agent, and it is far from being optimized in terms of memory consumption.

christarazi (approved):

PR #2448, opened 2024-03-23
Lock the connectivity command

michi-covalent:


tklauser (commented):
Might be worth to also add a package-level godoc comment stating the lock.

tklauser (approved):
Let's do it!

aanm (approved):

PR #2450, opened 2024-03-25
Fix flow validation with fully-qualified DNS names

giorio94:
Make sure that we don't append a trailing dot to the DNS name to be matched if already fully-qualified, and remove it from HTTP validation to match the behavior of the curl action.


giorio94:
CI is green. Dropping the temporary commit and marking ready for review

PR #2454, opened 2024-03-28
Encryption status output enriched with IPsec details

viktor-kurchenko:
Expected IPsec key count, IPsec key type and IPsec key rotation in progress status added to the encrypt output.

### DIFF
--- a/encrypt/model.go
+++ b/encrypt/model.go
@@ -3,14 +3,26 @@
 
 package encrypt
 
+import "github.com/cilium/cilium/api/v1/models"
+
 type clusterStatus struct {
-	TotalNodeCount          int              `json:"total-node-count,omitempty"`
-	EncDisabledNodeCount    int              `json:"enc-disabled-node-count,omitempty"`
-	EncIPsecNodeCount       int              `json:"enc-ipsec-node-count,omitempty"`
-	EncWireguardNodeCount   int              `json:"enc-wireguard-node-count,omitempty"`
-	IPsecKeysInUseNodeCount map[int64]int64  `json:"ipsec-keys-in-use-node-count,omitempty"`
-	IPsecMaxSeqNum          string           `json:"ipsec-max-seq-num,omitempty"`
-	IPsecErrCount           int64            `json:"ipsec-err-count,omitempty"`
-	XfrmErrors              map[string]int64 `json:"xfrm-errors,omitempty"`
-	XfrmErrorNodeCount      map[string]int64 `json:"xfrm-error-node-count,omitempty"`
+	TotalNodeCount             int              `json:"total-node-count,omitempty"`
+	EncDisabledNodeCount       int              `json:"enc-disabled-node-count,omitempty"`
+	EncIPsecNodeCount          int              `json:"enc-ipsec-node-count,omitempty"`
+	EncWireguardNodeCount      int              `json:"enc-wireguard-node-count,omitempty"`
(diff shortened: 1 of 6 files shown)


jschwinger233 (approved):
Looks good. Then we should consider use cilium-cli to perform key rotation in ci-ipsec-e2e, (and even in the ipsec doc).

encrypt/model.go
```diff
@@ -3,14 +3,26 @@
...
+	IPsecMaxSeqNum             string           `json:"ipsec-max-seq-num,omitempty"`
+	IPsecErrCount              int64            `json:"ipsec-err-count,omitempty"`
+	IPsecPerNodeKey            bool             `json:"ipsec-per-node-key,omitempty"`
+	IPsecKeyRotationInProgress bool             `json:"ipsec-key-rotation-in-progress,omitempty"`
+	IPsecExpectedKeyCount      int              `json:"ipsec-expected-key-count,omitempty"`
+	XfrmErrors                 map[string]int64 `json:"xfrm-errors,omitempty"`
+	XfrmErrorNodeCount         map[string]int64 `json:"xfrm-error-node-count,omitempty"`
+}
+
+type encryptionStatus struct {
```
asauber, line 23: Rather than calculate these additional fields per node, they should be calculated once for the entire cluster, and thus shouldn't appear in this type (this new type is not needed).

encrypt/status.go
```diff
@@ -184,6 +249,9 @@ func clusterNodeStatus(res map[string]models.EncryptionStatus) (clusterStatus, e
 		}
 		if v.Mode == "IPsec" {
 			cs.EncIPsecNodeCount++
+			cs.IPsecExpectedKeyCount = v.IPsecExpectedKeyCount
```
asauber, line 252: This silently uses the value of these fields from the last `encryptionStatus`.

Rather than store these fields on each `encryptionStatus`, they should be calculated once for the entire cluster (as mentioned above), by using the values from the first `models.EncryptionStatus` with mode "IPSec".

This eliminates some anti-patterns:

* Copying fields into each element of a collection before they are presented
* Storing overlapping data in a structure that could diverge over time

asauber (changes requested):

viktor-kurchenko:
> Looks good. Then we should consider using cilium-cli to perform key rotation in ci-ipsec-e2e, (and even in the ipsec doc).

@jschwinger233 I've refactored `expectedIPsecKeyCount` function, could you please double check?
Thanks!

encrypt/status.go
jschwinger233: Viktor: how about making `clusterStatus`  as a return value of `enrichWithIPsecDetails()`, where we can store `expectedKeyCount` in `clusterStatus`. In this way, we don't have to store `expectedKeyCount` in nodes' `encryptionStatus` and read from there later.

jschwinger233 (commented):
The revised `expectedKeyCount` looks good, thank you!

encrypt/status.go
viktor-kurchenko: @asauber @jschwinger233 I've refactored IPsec status but still have no idea how to get rid of `nodeStatus` struct.
Could you please review one more time?
Thanks!

jschwinger233 (approved):

viktor-kurchenko:
@brb kind ping )

brb (approved):

PR #2472, opened 2024-04-07
ci: Use zip output for windows releases

jbpaux:
Windows users nearly never use tar.gz files but it's common to use zip files.

I suggest this PR to change in the releases the output format for windows binaries.


jbpaux:
I've added the zip utility as it's not by default in the docker image. You can retrigger the workflow :)

tklauser (approved):
Thanks

PR #2414, opened 2024-03-15
connectivity: Improve coverage of Ingress service

giorio94:
Please refer to the individual commits for additional details.

Blocked by https://github.com/cilium/cilium/issues/31653

PR #2476, opened 2024-04-10
ingress: Use same node service

sayboras:
### Description

This commit is to use same node service for Ingress, so that we don't
really need to have multi-node cluster for Ingress Controller, also this
will help with hairpining case.

### Testing

Testing was done in https://github.com/cilium/cilium/pull/31280


giorio94:
Could we go for https://github.com/cilium/cilium-cli/pull/2414 instead? Otherwise we would lose coverage in the clustermesh case.

Issue #2478, opened 2024-04-11
mark AKS (BYOCNI) as required


michi-covalent:
i will mark https://github.com/cilium/cilium-cli/actions/workflows/aks-byocni.yaml as not required. it is consistently failing at `Clean up AKS` step, most likely due to https://github.com/Azure/azure-cli/issues/28708. this is a tracking issue to keep an eye on the upstream issue, and mark the `AKS (BYOCNI)` as required again once it's fixed.

michi-covalent:
the aks issue got fixed in #2546. added `AKS BYOCNI Installation and Connectivity Test` back to the required status check list. closing.

PR #2479, opened 2024-04-11
ci: refresh login before aks cleanup

MrFreezeex:
Since somewhat recently aks credential seems to be only valid for 5min. For instance here is an error fetched on an action failure when trying to cleanup a cluster.

```
ERROR: AADSTS700024: Client assertion is not within its valid time range. Current time: 2024-04-11T18:28:00.5441148Z, assertion valid from 2024-04-11T18:01:56.0000000Z, expiry time of assertion 2024-04-11T18:06:56.0000000Z.
```

This commit solves that by login in again right before we issue the cleanup task.


MrFreezeex:
nevermind I didn't see https://github.com/cilium/cilium-cli/issues/2478

Issue #1930, opened 2023-08-23
cilium execute command


liyihuang:
## Proposal / RFE

I personally have run these scenario a lot.

- I have a pod nginx running and I would like to get the cilium pod with this nginx. I will need to perform the following steps
    1. I run `kubectl get pods -o wide` to find out the node it's running for this nginx pod
    2. I run `kubectl get pods -l k8s-app=cilium -o name -n kube-system -o wid` to find out which cilium is on the same node with nginx
    3. I run `kubectl exec -it -n kube-system cilium-zmglg -- cilium xxx` or `kubectl exec -it -n kube-system cilium-zmglg -- bash` to get it cilium
 - I would like to collect all cilium commands for something, and I have to run these shell command
     1. `kubectl get pods -l k8s-app=cilium -o name -n kube-system | xargs -I {} sh -c 'echo Pod: {}; kubectl exec -n kube-system {} -- cilium status; echo'`

**Is your feature request related to a problem?**

**Describe the solution you'd like**

I hope we can have `cilium execute podname` where I will can run the cilium command to the pod that I want so I dont have to run step 1, 2, and 3 and `cilium execute all` where I can run the cilium command to all cilium pods

liyihuang:
This is what it should look like. Let's say I want to run some command on cilium pod
```
cilium exec --pod nginx-xxx-yyy cilium-health status
Defaulted container "cilium-agent" out of: cilium-agent, config (init), mount-cgroup (init), apply-sysctl-overwrites (init), mount-bpf-fs (init), wait-for-node-init (init), clean-cilium-state (init), install-cni-binaries (init)
Probe time:   2023-08-23T14:32:36Z
Nodes:
  ip-10-2-6-163.us-east-2.compute.internal (localhost):
    Host connectivity to 10.2.6.163:
      ICMP to stack:   OK, RTT=404.483µs
      HTTP to agent:   OK, RTT=1.602324ms
    Endpoint connectivity to 10.2.6.107:
      ICMP to stack:   OK, RTT=412.573µs
      HTTP to agent:   OK, RTT=1.135271ms
  ip-10-2-5-115.us-east-2.compute.internal:
    Host connectivity to 10.2.5.115:
      ICMP to stack:   Connection timed out
      HTTP to agent:   OK, RTT=2.150797ms
    Endpoint connectivity to 10.2.5.239:
      ICMP to stack:   Connection timed out
      HTTP to agent:   OK, RTT=2.062174ms
[...]

PR #2490, opened 2024-04-18
add cilium status -o text for cilium CLI

liyihuang:
add cilium status -o text to get the cilium status --verbose from agent, so people can use cilium CLI to troubleshoot the issue without get into the agent container

add the -p --podname to select which pod they want to choose for cilium status --verbose.

By default -p is empty and will randomly pick one.

### DIFF
--- a/cli/status.go
+++ b/cli/status.go
@@ -23,6 +23,13 @@ func newCmdStatus() *cobra.Command {
 		Use:   "status",
 		Short: "Display status",
 		Long:  ``,
+		PreRun: func(cmd *cobra.Command, _ []string) {
+			pod, _ := cmd.Flags().GetString("pod")
+			output, _ := cmd.Flags().GetString("output")
+			if pod != "" && output != "text" {
+				fatalf("specifying a --pod requires output option -o text")
+			}
+		},
 		RunE: func(_ *cobra.Command, _ []string) error {
 			params.Namespace = namespace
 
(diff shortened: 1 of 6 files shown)


YutaroHayakawa:
Please fix the build error pointed out by CI. Let me set this to draft for now. Please push the "Ready for Review` button once you're done.

liyihuang:
I think the current issue is on CI than the code

YutaroHayakawa (commented):
It looks like this PR extends the `cilium status` command but introduces new functionality. The current `cilium status` focuses on the global view of the Cilium as a system, including the status of the Cilium Operator, Hubble, or ClusterMesh. However, this extension focuses on the single agent's status. Switching those two behaviors using output format specifier (text) feels awkward for me. I'd suggest making a new command like `cilium agent status` instead.

Also, I'm wondering how much value this command adds over `kubectl exec <cilium-pod> -- cilium-dbg status --verbose`. Seems like the difference is random pod selection and reducing some types. If we just want to reduce the type, we can do it with a bit of shell scripting.

```
function cs() {
	pod_name=$1
	kubectl -n kube-system exec -it $pod_name -- cilium-dbg status --verbose
}
```

liyihuang:
If the cilium status is focusing on overall status. I have no problem to implement a new sub command since I brought https://github.com/cilium/cilium-cli/issues/1930.

I totally understand this could be implemented by putting some scripts somewhere. If a lot of users need to do that, then I think we could improve the UX so people don't have to write their own script

soggiest:
> Also, I'm wondering how much value this command adds over kubectl exec <cilium-pod> -- cilium-dbg status --verbose. Seems like the difference is random pod selection and reducing some types. Then, we can do it with a bit of shell scripting.

While it's true we can get similar functionality by shell scripting I do believe there is value in being able to get this level of information for one or more agents as part of the Cilium-CLI. There have been multiple instances (this is anecdotal I know, but bear with me) of customers requesting this level of functionality from the CLI, it was confusing at times to use the Cilium-CLI for high level status and connectivity testing and then having to use cilium-dbg via kubectl exec for more verbose status and troubleshooting. If we can it would be convenient to have as much troubleshooting functionality in the Cilium-CLI as possible.

liyihuang:
Another way of implementing this is to have subcommand like cilium agent or cilium exec. then we have more flexibility of what we can cover here(https://github.com/cilium/cilium-cli/issues/1930).

let's say we have `cilium agent` subcommand to run the command against the agent. I can think of the following use case

1. `cilium agent exec cilium status --verbose` randomly run `cilium status --verbose` on one of the agents
2. `cilium agent --podname all exec cilium status --verbose`, run `cilium status --verbose` on all of the agents
3. `cilium agent --podname nginx-1 exec cilium status --verbose` run `cilium status --verbose` on the cilium on the same node with nginx-1 

Then cilium status --verbose is be one of examples, where we can get whatever command we want here. Another example I can think of we tried to run some bash to gather the output of the ipsec status for the ipsec issue last year.

asauber:
Implementing a general `exec` facility opens the `cilium-cli` up to a class of security vulnerabilities that are not worth the trouble. As a general illustration, [many CVEs](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=%22exec%22) are the result of poorly sanitized input to an `exec` function. In the implementation proposed, `cilium agent exec` would allow commands to be run in the _host_ network namespace with _root_ level privileges on any Node in a Kubernetes cluster or other system running Cilium. I think this is best left as a problem for kubectl to solve. I have the following alias set in my bashrc to accomplish a similar result.

```
alias cexec="kubectl exec -ti -n kube-system daemonset/cilium -c cilium-agent --"
```

This allows usage like

```
cexec cilium status
```

That being said, I do think that having a first-class subcommand to print agent status as implemented here would be useful. To help distinguish it from the current `status` subcommand, I like the idea of `cilium agent status`, with usage similar to:

```
cilium agent status
cilium agent status --pod all
cilium agent status --pod cilium-7gx48 cilium-8gb31 # --pod matches cilium pod names only
cilium agent status --withpod nginx-1               # --withpod matches sibling pods
```

liyihuang:
closing this PR after the discussion and will have the seperate cilium agent subcommand PR to address this type of ask

PR #2496, opened 2024-04-22
Connectivity test concurrent run

viktor-kurchenko:
This PR introduces connectivity test concurrent run.

The new `test-concurrency` input param can be used to specify a count of K8S test namespaces.
So, most connectivity tests will run concurrently across test namespaces.

The new parameter is marked `hidden` because of the following reasons:
1. test and stabilisation internally (e.g.: in CI)
2. report output must be improved to avoid mess (especially for the `debug` and `verbose` modes). I want to address this in a separate PR.

Command run example: `cilium connectivity test --test-concurrency=5`

Local test in a 4 node kind cluster result:
- ~17 minutes with [`--test-concurrency=1`](https://github.com/cilium/cilium-cli/actions/runs/9057864520/job/24882553506?pr=2496)
- ~5 minutes with [`--test-concurrency=5`](https://github.com/cilium/cilium-cli/actions/runs/9057864520/job/24882553700?pr=2496)

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -24,6 +24,14 @@ jobs:
   installation-and-connectivity:
     name: Kind Installation and Connectivity Test
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # run connectivity tests explicitly without concurrency
+          - test-concurrency: 1
+          # run connectivity tests concurrently
+          - test-concurrency: 5
     timeout-minutes: 50
     steps:
       - name: Checkout
@@ -106,13 +114,15 @@ jobs:
           #
           # Dispatch interval is set to 100ms, b/c otherwise (default is 0), the flow validation might time out.
           cilium connectivity test --debug --test-namespace test-namespace \
+            --test-concurrency=${{ matrix.test-concurrency }} \
             --conn-disrupt-dispatch-interval 100ms \
             --include-conn-disrupt-test --conn-disrupt-test-setup
 
           # Run the connectivity test in non-default namespace (i.e. not cilium-test)
           cilium connectivity test --debug --all-flows --test-namespace test-namespace \
+            --test-concurrency=${{ matrix.test-concurrency }} \
             --include-unsafe-tests --include-conn-disrupt-test \
-            --collect-sysdump-on-failure --junit-file cilium-junit-1.xml \
+            --collect-sysdump-on-failure --junit-file cilium-junit-1-concurrency-${{ matrix.test-concurrency }}.xml \
             --junit-property type=no-tunnel \
             --curl-insecure \
             --external-target chart-testing-worker2 \
@@ -152,8 +162,9 @@ jobs:
       - name: Connectivity test
         run: |
           cilium connectivity test --debug --force-deploy --all-flows --test-namespace test-namespace \
+            --test-concurrency=${{ matrix.test-concurrency }} \
             --include-unsafe-tests \
-            --collect-sysdump-on-failure --junit-file cilium-junit-2.xml \
+            --collect-sysdump-on-failure --junit-file cilium-junit-2-concurrency-${{ matrix.test-concurrency }}.xml \
             --junit-property type=ipsec \
             --curl-insecure \
             --external-target chart-testing-worker2 \
@@ -167,7 +178,7 @@ jobs:
         if: ${{ always() }}
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
-          name: cilium-junits
+          name: cilium-junits-concurrency-${{ matrix.test-concurrency }}
           path: cilium-junit*.xml
           retention-days: 2
 
@@ -176,14 +187,14 @@ jobs:
         run: |
           cilium status
           kubectl get pods --all-namespaces -o wide
-          cilium sysdump --output-filename cilium-sysdump-out --hubble-flows-count 10000
+          cilium sysdump --output-filename cilium-sysdump-out-concurrency-${{ matrix.test-concurrency }} --hubble-flows-count 10000
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
       - name: Upload sysdump
         if: ${{ !success() }}
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
-          name: cilium-sysdumps
+          name: cilium-sysdumps-concurrency-${{ matrix.test-concurrency }}
           path: cilium-sysdump-*.zip
           retention-days: 5
 
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -25,7 +25,6 @@ jobs:
     name: Kind Installation and Connectivity Test
     runs-on: ubuntu-22.04
     strategy:
-      fail-fast: false
       matrix:
         include:
           # run connectivity tests explicitly without concurrency
@@ -201,6 +200,13 @@ jobs:
   helm-upgrade-clustermesh:
     name: Kind Helm Upgrade Clustermesh
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+          # run connectivity tests explicitly without concurrency
+          - test-concurrency: 1
+          # run connectivity tests concurrently
+          - test-concurrency: 5
     timeout-minutes: 50
 
     env:
@@ -326,19 +332,21 @@ jobs:
           #
           # Dispatch interval is set to 100ms, b/c otherwise (default is 0), the flow validation might time out.
           cilium connectivity test --context $CLUSTER1 --multi-cluster $CLUSTER2 --debug \
+            --test-concurrency=${{ matrix.test-concurrency }} \
             --conn-disrupt-dispatch-interval 100ms \
             --include-conn-disrupt-test --conn-disrupt-test-setup
 
           cilium connectivity test --context $CLUSTER1 --multi-cluster $CLUSTER2 --debug \
+            --test-concurrency=${{ matrix.test-concurrency }} \
             --include-unsafe-tests --include-conn-disrupt-test \
-            --collect-sysdump-on-failure --junit-file cilium-junit-clustermesh-1.xml \
+            --collect-sysdump-on-failure --junit-file cilium-junit-clustermesh-1-concurrency-${{ matrix.test-concurrency }}.xml \
             --junit-property mode=clustermesh --junit-property type=ipsec
 
       - name: Upload JUnit
         if: ${{ always() }}
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
-          name: cilium-junits-helm-upgrade-clustermesh
+          name: cilium-junits-helm-upgrade-clustermesh-concurrency-${{ matrix.test-concurrency }}
           path: cilium-junit*.xml
           retention-days: 2
 
@@ -347,24 +355,24 @@ jobs:
         run: |
           cilium --context $CLUSTER1 status
           kubectl --context $CLUSTER1 get pods --all-namespaces -o wide
-          cilium --context $CLUSTER1 sysdump --output-filename cilium-sysdump-out-c1
+          cilium --context $CLUSTER1 sysdump --output-filename cilium-sysdump-out-c1-concurrency-${{ matrix.test-concurrency }}
           cilium --context $CLUSTER2 status
           kubectl --context $CLUSTER2 get pods --all-namespaces -o wide
-          cilium --context $CLUSTER2 sysdump --output-filename cilium-sysdump-out-c2
+          cilium --context $CLUSTER2 sysdump --output-filename cilium-sysdump-out-c2-concurrency-${{ matrix.test-concurrency }}
         shell: bash {0} # Disable default fail-fast behaviour so that all commands run independently
 
       - name: Upload sysdump from cluster 1
         if: ${{ !success() }}
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
-          name: cilium-sysdump-out-c1.zip
-          path: cilium-sysdump-out-c1.zip
+          name: cilium-sysdump-out-c1-concurrency-${{ matrix.test-concurrency }}.zip
+          path: cilium-sysdump-out-c1-concurrency-${{ matrix.test-concurrency }}.zip
           retention-days: 5
 
       - name: Upload sysdump from cluster 2
         if: ${{ !success() }}
         uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
-          name: cilium-sysdump-out-c2.zip
-          path: cilium-sysdump-out-c2.zip
+          name: cilium-sysdump-out-c2-concurrency-${{ matrix.test-concurrency }}.zip
+          path: cilium-sysdump-out-c2-concurrency-${{ matrix.test-concurrency }}.zip
           retention-days: 5
--- a/cli/connectivity.go
+++ b/cli/connectivity.go
@@ -38,6 +38,7 @@ func newCmdConnectivity(hooks api.Hooks) *cobra.Command {
 
 var params = check.Parameters{
 	ExternalDeploymentPort: 8080,
+	EchoServerHostPort:     4000,
 	Writer:                 os.Stdout,
 	SysdumpOptions: sysdump.Options{
 		LargeSysdumpAbortTimeout: sysdump.DefaultLargeSysdumpAbortTimeout,
@@ -234,6 +235,7 @@ func newConnectivityTests(params check.Parameters) ([]*check.ConnectivityTest, e
 		params := params
 		params.TestNamespace = fmt.Sprintf("%s-%d", params.TestNamespace, i+1)
 		params.ExternalDeploymentPort += i
+		params.EchoServerHostPort += i
 		cc, err := check.NewConnectivityTest(k8sClient, params, defaults.CLIVersion)
 		if err != nil {
 			return nil, err
--- a/cli/connectivity.go
+++ b/cli/connectivity.go
@@ -86,7 +86,7 @@ func RunE(hooks api.Hooks) func(cmd *cobra.Command, args []string) error {
 			connTests[0].Logf("Cancellation request (%s) received, cancelling tests...", context.Cause(ctx))
 		}()
 
-		return connectivity.Run(ctx, params, connTests, hooks)
+		return connectivity.Run(ctx, connTests, hooks)
 	}
 }
 
--- a/cli/connectivity.go
+++ b/cli/connectivity.go
@@ -5,7 +5,6 @@ package cli
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -24,8 +23,6 @@ import (
 	"github.com/cilium/cilium-cli/sysdump"
 )
 
-var errInternal = errors.New("encountered internal error, exiting")
-
 func newCmdConnectivity(hooks api.Hooks) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "connectivity",
@@ -40,7 +37,8 @@ func newCmdConnectivity(hooks api.Hooks) *cobra.Command {
 }
 
 var params = check.Parameters{
-	Writer: os.Stdout,
+	ExternalDeploymentPort: 8080,
+	Writer:                 os.Stdout,
 	SysdumpOptions: sysdump.Options{
 		LargeSysdumpAbortTimeout: sysdump.DefaultLargeSysdumpAbortTimeout,
 		LargeSysdumpThreshold:    sysdump.DefaultLargeSysdumpThreshold,
@@ -70,8 +68,7 @@ func RunE(hooks api.Hooks) func(cmd *cobra.Command, args []string) error {
 			}
 		}
 
-		// Instantiate the test harness.
-		cc, err := check.NewConnectivityTest(k8sClient, params, defaults.CLIVersion)
+		connTests, err := newConnectivityTests(params)
 		if err != nil {
 			return err
 		}
@@ -86,33 +83,10 @@ func RunE(hooks api.Hooks) func(cmd *cobra.Command, args []string) error {
 
 		go func() {
 			<-ctx.Done()
-			cc.Logf("Cancellation request (%s) received, cancelling tests...", context.Cause(ctx))
+			connTests[0].Logf("Cancellation request (%s) received, cancelling tests...", context.Cause(ctx))
 		}()
 
-		done := make(chan struct{})
-		var finished bool
-
-		// Execute connectivity.Run() in its own goroutine, it might call Fatal()
-		// and end the goroutine without returning.
-		go func() {
-			defer func() { done <- struct{}{} }()
-			err = connectivity.Run(ctx, cc, hooks)
-
-			// If Fatal() was called in the test suite, the statement below won't fire.
-			finished = true
-		}()
-		<-done
-
-		if !finished {
-			// Exit with a non-zero return code.
-			return errInternal
-		}
-
-		if err != nil {
-			return fmt.Errorf("connectivity test failed: %w", err)
-		}
-
-		return nil
+		return connectivity.Run(ctx, params, connTests, hooks)
 	}
 }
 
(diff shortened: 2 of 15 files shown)


michi-covalent:
@viktor-kurchenko could you take a look at kind test failures? `from-cidr-host-netns/from-cidr-to-pod/host-netns-to-pod` seems to be failing consistently.

connectivity/suite.go
```diff
@@ -5,6 +5,9 @@ package connectivity
 
 import (
 	"context"
+	"errors"
+
+	"github.com/hashicorp/go-multierror"
```
michi-covalent, line 10: maybe use https://pkg.go.dev/errors#Unwrap instead? you just need to accumulate errors and unwrap right? @tklauser what do you think?

connectivity/check/deployment.go
```diff
@@ -875,18 +875,19 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
...
-				containerPort := 8080
+				// in case if test concurrency is > 1 port must be unique for each test namespace
+				port := 8080 + ct.id
 				echoExternalDeployment := newDeployment(deploymentParameters{
 					Name:           echoExternalNodeDeploymentName,
 					Kind:           kindEchoExternalNodeName,
-					Port:           containerPort,
-					NamedPort:      "http-8080",
-					HostPort:       8080,
+					Port:           port,
```
michi-covalent, line 883: i probably missed it, but where is the corresponding change to adjust cilium network policies?

michi-covalent (commented):
let's add a matrix for the kind job and run with both `--test-concurrency=1` and `--test-concurrency=5`: https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml#L24

cli/install.go
```diff
@@ -105,7 +105,7 @@ func newCmdUninstallWithHelm() *cobra.Command {
 			params.HelmReleaseName = helmReleaseName
 			ctx := context.Background()
 
-			cc, err := check.NewConnectivityTest(k8sClient, check.Parameters{
+			cc, err := check.NewConnectivityTest(0, k8sClient, check.Parameters{
```
michi-covalent, line 108: this won't work if `--test-concurrency` was set to a value greater than 1.

i don't like the fact `install` depends on `connectivity`. let me think about how to decouple these 2 more cleanly 💭

michi-covalent: let's start with this refactoring: https://github.com/cilium/cilium-cli/pull/2498

after this i'll make the change to delete test namespaces by prefix.

connectivity/suite.go
tklauser: Ack, using `%w` and/or `errors.Unwrap` would be preferable.

marseel (approved):

connectivity/check/policy.go
```diff
@@ -454,12 +455,19 @@ func sumMap(m map[string]int) int {
...
+// can apply or delete policies in case of connectivity test concurrency > 1
+var policyApplyDeleteLock = sync.Mutex{}
+
 // applyPolicies applies all the Test's registered network policies.
 func (t *Test) applyPolicies(ctx context.Context) error {
 	if len(t.cnps) == 0 && len(t.ccnps) == 0 && len(t.knps) == 0 && len(t.cegps) == 0 {
 		return nil
 	}
 
+	policyApplyDeleteLock.Lock()
```
nebril, line 468: I guess nothing stops a test running with CCNP from deleting it and causing other test to fail if there is a name collision between CCNPs.

Why do we need this lock in the first place? Every test is only managing it's own policies isn't it?

Unless it's about policy revision numbers, which I guess are racey.

nebril (changes requested):
Awesome change overall, very happy to see it! I think the only blocking comment from me is showing the user which test Fatald.

connectivity/check/policy.go
michi-covalent: yeah it is about policy revision numbers unfortunately 😥 

so right now the connectivity test waits for the revision number to get incremented after applying / deleting policies. the idea is to wait until cilium reconciles the policy before generating traffic.

i vaguely remember chatting with @joestringer about this, i wanted a more reliable way to check whether cilium-agent implemented a particular CNP revision. not sure if anything concrete came out of it 💭

joestringer: Right, there's no great mechanism to create a CNP and then find out whether it's been implemented or not. The policy revision number stuff helps in a sense, because each time you create a policy, it changes this number. But there's no way to correlate the revision number with the specific policy, so we mainly rely on being the only one to create policy during a window, then if the number increases, we know it was implemented.

Typically this is something you might expect to show up in a status field for a CR. Unfortunately since policy implementation is per-node, this ends up with `N` policies each reporting the implementation status for `M` nodes, which scales poorly. IIRC we even removed the option to even report CNP status recently (https://github.com/cilium/cilium/pull/29395).

This would be a good topic to bring up with @cilium/sig-policy (#sig-policy on Slack) to discuss what reasonable options there may be to support this workflow. I think it'd be quite useful to users as well if we came up with a good solution to this problem.

cli/install.go
michi-covalent: https://github.com/cilium/cilium-cli/pull/2503 to delete test namespaces by prefix

michi-covalent: forget https://github.com/cilium/cilium-cli/pull/2503, you can rebase now 🚀

connectivity/suite.go
viktor-kurchenko: Not sure here... `go-multierror` allows to [run many goroutines and collect all the errors](https://github.com/cilium/cilium-cli/pull/2496/files#diff-ab78417b50a516eda19c907a86d67689ea4deb1708e96e6efab60bcce67765b4R49-R56).
Am I missing something?

cli/install.go
viktor-kurchenko: Thank you, Michi!

connectivity/suite.go
michi-covalent: no you are not missing anything, it's just a tradeoff between writing our own code with standard libraries vs using third party libraries. i think in this case it's reasonable to write a bit more code to collect errors from goroutines and avoid depending on another library.

cli/connectivity.go
```diff
@@ -238,3 +214,25 @@ func registerCommonFlags(flags *pflag.FlagSet) {
...
 }
+
+func newConnectivityTests(params check.Parameters) ([]*check.ConnectivityTest, error) {
+	if params.TestConcurrency < 2 {
+		cc, err := check.NewConnectivityTest(0, k8sClient, params, defaults.CLIVersion)
+		if err != nil {
+			return nil, err
+		}
+		return []*check.ConnectivityTest{cc}, nil
+	}
```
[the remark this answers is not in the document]
viktor-kurchenko: Fixed, thanks!

connectivity/suite.go
viktor-kurchenko: I've implemented [MultiError runner](https://github.com/cilium/cilium-cli/pull/2496/files#diff-9f58834ede9ee2a7744d0a7fb594cd2b764629ce312f4898979924bd74cecf81R14) component to replace HashiCorp lib.

connectivity/suite.go
```diff
@@ -17,47 +20,64 @@ type Hooks interface {
...
+	if err := meg.Wait().ErrorOrNil(); err != nil {
 		return err
 	}
-
-	if err := builder.FinalTests(ct); err != nil {
-		return err
+	for i := 0; i < len(connTests); i++ {
+		if !finish[i] {
+			// Exit with a non-zero return code.
+			return errors.New("encountered internal error, exiting")
```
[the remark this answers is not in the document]
viktor-kurchenko: Good question!
I don't think it's possible to do it in the suite because it doesn't know which test crashed a goroutine.
We can try to improve it [here](https://github.com/cilium/cilium-cli/blob/main/connectivity/check/context.go#L374).

WDYT?

viktor-kurchenko:
> let's add a matrix for the kind job and run with both `--test-concurrency=1` and `--test-concurrency=5`: https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml#L24

I've done some testing in CI and if `--test-concurrency=3` kind cluster needs 5 Cilium workers (more resources needed).
Maybe let's start with `3`?

michi-covalent:
>  Maybe let's start with 3?

yeah any number greater than 1 is fine 🚀🙏

viktor-kurchenko:
> @viktor-kurchenko could you take a look at kind test failures? `from-cidr-host-netns/from-cidr-to-pod/host-netns-to-pod` seems to be failing consistently.

Fixed!

viktor-kurchenko:
> yeah any number greater than 1 is fine 🚀🙏

Done.
Also, I've added [this ugly thing](https://github.com/cilium/cilium-cli/pull/2496/files#diff-e9c24daafcd1e27b49ed8503aee69248c36d7fc1a47ecbafffe36ba946c61f27R102-R103) due to the [annoying error](https://github.com/cilium/cilium-cli/actions/runs/8974239249/job/24648943491?pr=2496#step:11:78)
Have you seen it before?

connectivity/check/policy.go
viktor-kurchenko: Yeah, the lock has been added due to the policy revision number.
Is it ok to resolve this conversation?

connectivity/check/deployment.go
viktor-kurchenko: Good question!
I've re-worker it a bit and replaced ConnectivityTest.id with [Parameters.ExternalDeploymentPort](https://github.com/cilium/cilium-cli/pull/2496/files#diff-6a57efc490ffd3a1c4a9274abc080ec97e305f89af38214d38f09736b445a2e7R68).
But I don't see any network policy that needs to be adjusted...

michi-covalent:
> Also, I've added [this ugly thing](https://github.com/cilium/cilium-cli/pull/2496/files#diff-e9c24daafcd1e27b49ed8503aee69248c36d7fc1a47ecbafffe36ba946c61f27R102-R103) due to the [annoying error](https://github.com/cilium/cilium-cli/actions/runs/8974239249/job/24648943491?pr=2496#step:11:78)
Have you seen it before?

no i haven't seen it before, we probably shouldn't do `|| true` to ignore the error. it might be better to increase the number of workers in a separate pull request so that this pull request doesn't keep growing 💭

viktor-kurchenko:
> no i haven't seen it before, we probably shouldn't do `|| true` to ignore the error. it might be better to increase the number of workers in a separate pull request so that this pull request doesn't keep growing 💭

So, should I revert the CI changes for now?

nebril (approved):
LGTM, thanks!

michi-covalent:
> So, should I revert the CI changes for now?

you can keep this pr as is for now. just open a separate pr to increase the number of workers, and we can figure out what's going on there.

.github/workflows/kind.yaml
```diff
@@ -85,14 +91,15 @@ jobs:
...
-          echo "worker3_ip=${worker3_ip}" >> $GITHUB_OUTPUT
+          export worker6_ip=$(kubectl get nodes chart-testing-worker6 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
+          export worker7_ip=$(kubectl get nodes chart-testing-worker7 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
+          echo "worker6_ip=${worker6_ip}" >> $GITHUB_OUTPUT
+          echo "worker7_ip=${worker7_ip}" >> $GITHUB_OUTPUT
           kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml
           kubectl rollout status -n cert-manager deployment.apps/cert-manager
           kubectl rollout status -n cert-manager deployment.apps/cert-manager-webhook
           kubectl create ns external-targets
+          cat .github/external-targets/certs.yaml | envsubst | kubectl apply -n external-targets -f - || true
```
joestringer, line 102: Why is this necessary?

.github/workflows/kind.yaml
```diff
@@ -85,14 +91,15 @@ jobs:
...
         id: external_targets
         run: |
-          export worker2_ip=$(kubectl get nodes chart-testing-worker2 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
-          export worker3_ip=$(kubectl get nodes chart-testing-worker3 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
-          echo "worker2_ip=${worker2_ip}" >> $GITHUB_OUTPUT
-          echo "worker3_ip=${worker3_ip}" >> $GITHUB_OUTPUT
+          export worker6_ip=$(kubectl get nodes chart-testing-worker6 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
+          export worker7_ip=$(kubectl get nodes chart-testing-worker7 -o=jsonpath='{.status.addresses[?(@.type=="InternalIP")].address}')
+          echo "worker6_ip=${worker6_ip}" >> $GITHUB_OUTPUT
+          echo "worker7_ip=${worker7_ip}" >> $GITHUB_OUTPUT
```
joestringer, line 97: I don't understand this change, there are several new nodes but then the test is changed to only use the last two nodes?

joestringer (dismissed):

viktor-kurchenko:
> you can keep this pr as is for now. just open a separate pr to increase the number of workers, and we can figure out what's going on there.

But without node increase it won't be possible to add CI matrix for concurrent run.
Even for the `--test-concurrency=2` it requires at least 3 cilium nodes.

.github/workflows/kind.yaml
viktor-kurchenko: It was an ugly attempt to fix this [error](https://github.com/cilium/cilium-cli/actions/runs/8974239249/job/24648943491?pr=2496#step:11:78).
I'll address it in a separate PR, removed.

.github/workflows/kind.yaml
viktor-kurchenko: [Two last nodes](https://github.com/cilium/cilium-cli/pull/2496/files#diff-94e083e5b848fc71129be9f77c0f2587aff5c3018835ada978a4fb75770e49f5R25-R33) have a special purpose: to host external (non-managed by Cilium) workloads.
So, I just increased the index for the external nodes after adding the new workers.

joestringer: Oh I see, I didn't look closely at the name of the step. Makes sense :+1:

.github/kind-config.yaml
```diff
@@ -14,6 +14,14 @@ nodes:
...
   - role: worker
     image: kindest/node:v1.19.11
+  - role: worker
+    image: kindest/node:v1.19.11
+  - role: worker
+    image: kindest/node:v1.19.11
+  - role: worker
+    image: kindest/node:v1.19.11
+  - role: worker
+    image: kindest/node:v1.19.11
```
joestringer, line 24: What's the motivation for more nodes in the kind environment?

I'm wondering overall how many of these do we need, and how they trade off the availability of resources like CPU etc. ? I presume that all of these nodes are effectively just local processes on the underlying multi-core environment, so whether we densely pack a couple of nodes or create many loosely packed nodes, the overall usage is presumably around the same. Downside with more nodes is that we would have more Cilium instances to run, sync resources between, etc., though I don't know if that's meaningful at the scale of the sorts of tests we run here.

On the other hand if for instance we think this makes the tests easier to debug when they go wrong, I could understand that rationale.

joestringer: ^ Would be good to add this "why" information to the corresponding commit message too.

viktor-kurchenko: Good point!
If we want to run conn tests with `--test-concurrency=3` it will spin up 3 test namespaces with 3 times more deployments. So, more resources are needed.
Maybe I'm missing something but without node increase, some conn test pods are stuck in the `Pending` status forever.

joestringer: Broadly the resources argument makes sense, but it's not obvious to me why additional nodes would make more CPU/memory available. I could also be missing something about the way that kind provisions the containers.

I think it'd be interesting to know why we hit that `Pending` status, and whether we are hitting some other known limit, whether it's configurable, or whether there's a hard requirement for more nodes. This knowledge could help inform future optimizations around how much concurrency to run, too.

.github/workflows/kind.yaml
```diff
@@ -24,6 +24,12 @@ jobs:
   installation-and-connectivity:
     name: Kind Installation and Connectivity Test
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test-concurrency: 1
+          - test-concurrency: 5
```
asauber, line 32: Is the intention with this matrix to run a suite of test that is not concurrent, implicit based on the concurrency parameter, and also run a _different_ suite of tests that is current, based upon the other parameter? If so, there should be a comment here explaining if this is the case (or not).

asauber (changes requested):

cli/connectivity.go
```diff
@@ -238,3 +216,31 @@ func registerCommonFlags(flags *pflag.FlagSet) {
...
+		fmt.Printf("--test-concurrency parameter value is invalid [%d], using 1 instead\n", params.TestConcurrency)
+		params.TestConcurrency = 1
+	}
+	if params.TestConcurrency < 2 {
+		cc, err := check.NewConnectivityTest(k8sClient, params, defaults.CLIVersion)
+		if err != nil {
+			return nil, err
+		}
+		return []*check.ConnectivityTest{cc}, nil
+	}
```
asauber, line 231: Rather than have a special case path for `TestConcurrency == 1` which is identical to the loop body, why not remove this block entirely?

connectivity/builder/builder.go
```diff
@@ -79,23 +79,86 @@ type testBuilder interface {
 	build(ct *check.ConnectivityTest, templates map[string]string)
 }
 
-// NetworkPerformanceTests builds the network performance connectivity tests.
-func NetworkPerformanceTests(ct *check.ConnectivityTest) error {
-	return getTests(ct, []testBuilder{networkPerf{}})
+// GetTestSuites returns a slice of connectivity test builder functions.
+// Each function contains set of tests according to the provided parameters.
```
asauber, line 83: This godoc doesn't really describe what this function is doing. 

This function returns a slice of functions, which when those functions are invoked, append `testBuilder` structs to a slice which are each invoked by an IoC pattern.

The key distinction here _is that this slice of functions are not the test builders_. They are two levels removed from the invocation of the test builders.

A better docstring would be:

```
// GetTestSuites returns a slice of functions, that when invoked, result in a
// collection of tests being built. These functions use helper methods to add various
// collections of test builders depending upon the provided parameters.
```

This function is in the `builder` package, and it's return value is called `builder` below, but we already have a new type named `testBuilder`, which is a struct that represents a test which can be built.

The type of these functions would therefore be better named `suiteBuilder` or similar.

connectivity/builder/builder.go
```diff
@@ -79,23 +79,86 @@ type testBuilder interface {
...
+func GetTestSuites(params check.Parameters) ([]func(connTests []*check.ConnectivityTest, extra func(ct *check.ConnectivityTest) error) error, error) {
+	switch {
+	case params.Perf:
+		return []func(connTests []*check.ConnectivityTest, _ func(ct *check.ConnectivityTest) error) error{
```
asauber, line 87: Giving `func(connTests []*check.ConnectivityTest, _ func(ct *check.ConnectivityTest) error) error` a type name would make this function more readable.

connectivity/suite.go
```diff
@@ -17,47 +19,64 @@ type Hooks interface {
...
+	if err := setupConnectivityTests(ctx, connTests, extra); err != nil {
 		return err
 	}
 
-	ct.Infof("Cilium version: %v", ct.CiliumVersion)
+	connTests[0].Infof("Cilium version: %v", connTests[0].CiliumVersion)
 
-	if ct.Params().Perf {
-		if err := builder.NetworkPerformanceTests(ct); err != nil {
+	builders, err := builder.GetTestSuites(connTests[0].Params())
```
asauber, line 29: This is where `builder` could instead be `suiteBuilder` to make the overall logic much more readable.

e.g.

```
suiteBuilders, err := builder.GetTestSuites(connTests[0].Params())
```

connectivity/builder/builder.go
```diff
@@ -79,23 +79,86 @@ type testBuilder interface {
...
+	case params.IncludeConnDisruptTest && params.ConnDisruptTestSetup:
+		// Exit early, as --conn-disrupt-test-setup is only needed to deploy pods which
+		// will be used by another invocation of "cli connectivity test"
+		// with include --include-conn-disrupt-test"
+		return []func(connTests []*check.ConnectivityTest, _ func(ct *check.ConnectivityTest) error) error{
+			func(connTests []*check.ConnectivityTest, _ func(ct *check.ConnectivityTest) error) error {
+				return connDisruptTests(connTests[0])
+			},
+		}, nil
+	case params.TestConcurrency > 1:
```
asauber, line 101: This value of `TestConcurrency` is never used to limit the concurrency to a specific value. The only purpose that it is serving with this setup is to run the `concurrentTests` in a separate collection of goroutines with unlimited concurrency, block, and then run the `sequentialTests` with unlimited concurrency.

I would expect this value of `TestConcurrency` to control the number of concurrent tests, which it does not appear to. Unless I'm missing something.

Based on the way that this `builder` is invoked below, even the `sequentialTests` will be invoked with unlimited concurrency by the `runner.MultiError`. Any collection of tests called `sequentialTests` should be run sequentially, with no concurrency at all.

Otherwise it would be better to call these `concurrentSuite1` and `concurrentSuite2`.

I may be missing something so let me know if this is not the behavior that you are seeing.

asauber (changes requested):

.github/kind-config.yaml
viktor-kurchenko: I found the root cause `:facepalm:`.
It's related to the `echoExternalNodePods` that use `Host Port`.
This commit contains the fix: https://github.com/cilium/cilium-cli/pull/2496/commits/b3f84699e268475aeb01b414c9b7e721c5c4bbe7

Thanks!

cli/connectivity.go
viktor-kurchenko: In case of `params.TestConcurrency < 2` it doesn't change `params.TestNamespace`.

.github/workflows/kind.yaml
viktor-kurchenko: Added.

connectivity/suite.go
viktor-kurchenko: Renamed.

connectivity/builder/builder.go
viktor-kurchenko: Fixed.

connectivity/builder/builder.go
viktor-kurchenko: Nice, thank you very much!
Fixed.

connectivity/builder/builder.go
viktor-kurchenko: `TestConcurrency` defines the max number of test namespaces that will be provisioned in a cluster.
To run two or > tests can be run concurrently only in separate namespaces to avoid policy collision.

Unfortunately, there are a couple of exceptions that lead to sequential runs:
1. Tests that use cluster-wide network policy
2. Check log test that must be always run at the end

.github/workflows/kind.yaml
```diff
@@ -24,6 +24,14 @@ jobs:
   installation-and-connectivity:
     name: Kind Installation and Connectivity Test
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # run connectivity tests explicitly without concurrency
+          - test-concurrency: 1
```
joestringer, line 32: Would it make sense to retain fail-fast for test-concurrency: 1 for quicker iteration and keep it off for higher concurrency?

joestringer (approved):
LGTM for my codeowners. I think the new policy lock retains the same assumptions we have today where only one piece of software should be manipulating policies at the time that you do the connectivity test (ie, only the CLI does it). And that should be good enough for now. It would be cool to explore better mechanisms for ensuring that datapath updates have occurred for a specific policy, but for now we don't really have a solution to that problem and it's not worth holding back this effort until we do that.

.github/workflows/kind.yaml
viktor-kurchenko: Good catch! I've disabled `fail-fast` for the debugging purpose.
I think it's better to remove it at all for the workflow.

cli/connectivity.go
asauber: Does anything rely on the pre-existing namespace name? I'm fine with this as is, but it look like an opportunity to remove a few chunks of code.

asauber (approved):

cli/connectivity.go
asauber: The namespace changing from `cilium-test` to `cilium-test-1`, doesn't seem too confusing. This change can be left for the future, up to you.

michi-covalent (approved):
let's do it

PR #2501, opened 2024-04-24
k8s: remove unused filter functionality from k8sclient.CiliumLogs

mhofstetter:
Currently, the logic to retrieve all logs from Cilium Pods is implemented using `bufio.Scanner`. This might result in errors for long  loglines - hiding the actual error.

```
Error reading Cilium logs: error reading cilium-agent logs for kube-system/cilium-c2sx7: bufio.Scanner: token too long
```

AFAIU, the only reason to use the `bufio.Scanner` is the support for filtering the logs by a given regular expression. But this functionality is not used.

Therefore, this commit removes the filter functionality and the usage of `bufio.Scanner`.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -15,7 +15,6 @@ import (
 	"io"
 	"net"
 	"os"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


nebril (approved):
Nice!

thorn3r (approved):

asauber (approved):
more red than green, very good

PR #2503, opened 2024-04-25
uninstall: Delete test namespaces by prefix

michi-covalent:
This commit changes the behavior of --test-namespace flag. Instead of only deleting the namespace that matches --test-namespace flag exactly, delete all the namespaces with the prefix specified in --test-namespace flag. This is in preparation to support running connectivity tests in parallel using multiple namespaces (#2496).

### DIFF
--- a/cli/install.go
+++ b/cli/install.go
@@ -30,7 +30,7 @@ func addCommonInstallFlags(cmd *cobra.Command, params *install.Parameters) {
 
 // addCommonUninstallFlags adds uninstall command flags that are shared between classic and helm mode.
 func addCommonUninstallFlags(cmd *cobra.Command, params *install.UninstallParameters) {
-	cmd.Flags().StringVar(&params.TestNamespace, "test-namespace", defaults.ConnectivityCheckNamespace, "Namespace to uninstall Cilium tests from")
+	cmd.Flags().StringVar(&params.TestNamespacePrefix, "test-namespace", defaults.ConnectivityCheckNamespace, "Prefix of namespaces to uninstall Cilium tests from")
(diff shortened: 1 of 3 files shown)


michi-covalent:
✅ tested manually
```
% ./cilium install
% kubectl create ns cilium-test
% kubectl create ns cilium-test-1
% kubectl create ns cilium-test-2
% kubectl create ns cilium-test-3
% ./cilium uninstall
🔥 Deleting pods in cilium-test namespace...
🔥 Deleting cilium-test namespace...
🔥 Deleting pods in cilium-test-1 namespace...
🔥 Deleting cilium-test-1 namespace...
🔥 Deleting pods in cilium-test-2 namespace...
🔥 Deleting cilium-test-2 namespace...
🔥 Deleting pods in cilium-test-3 namespace...
🔥 Deleting cilium-test-3 namespace...
⌛ Uninstalling Cilium
```

nebril (changes requested):
The change looks good overall, but I think we should not change the behaviour of this flag and should add a new one.

I am very afraid that somebody can run these tests in `x` namespaces and then they can also have `x-production` namespace in their cluster, which will be deleted AFAIU.

If we are deleting people's namespaces, I feel like it should be as explicit on their side as possible.

mhofstetter (changes requested):
i second the opinion of @nebril . I'm against changing the behaviour of the existing flag.

michi-covalent:
amazing code review @nebril @mhofstetter thank you for keeping me in check 🥰 let's add a separate opt-in --test-namespace-prefix flag 🎏

michi-covalent:
let me close this one 🥷

PR #2511, opened 2024-04-29
connectivity: Delete test-conn-disrupt pods immediately after test run

brb:
See commit msgs.

Test runs: https://github.com/cilium/cilium/pull/32234

### DIFF
--- a/connectivity/builder/no_interrupted_connections.go
+++ b/connectivity/builder/no_interrupted_connections.go
@@ -4,6 +4,8 @@
 package builder
 
 import (
+	"context"
+
 	"github.com/cilium/cilium-cli/connectivity/check"
 	"github.com/cilium/cilium-cli/connectivity/tests"
 )
@@ -13,5 +15,19 @@ type noInterruptedConnections struct{}
 func (t noInterruptedConnections) build(ct *check.ConnectivityTest, _ map[string]string) {
 	newTest("no-interrupted-connections", ct).
 		WithCondition(func() bool { return ct.Params().IncludeConnDisruptTest }).
-		WithScenarios(tests.NoInterruptedConnections())
+		WithScenarios(tests.NoInterruptedConnections()).
+		WithFinalizer(func(ctx context.Context) error {
+			// Delete the test-conn-disrupt pods immediately after the test run
+			// has finished to reduce CPU consumption (the pods generate a lot
+			// of traffic, and in most CI jobs Cilium runs with monitor
+			// aggregation disabled).
+			if !ct.Params().ConnDisruptTestSetup {
+				for _, client := range ct.Clients() {
+					if err := ct.DeleteConnDisruptTestDeployment(ctx, client); err != nil {
+						return err
+					}
+				}
+			}
+			return nil
+		})
 }
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -420,17 +420,6 @@ func (ct *ConnectivityTest) Run(ctx context.Context) error {
 		wg.Wait()
 	}
 
-	// Delete conn disrupt pods after running tests. Otherwise, after calling
-	// --flush-ct they might get into crashloop state from which it takes time
-	// to recover, which makes subsequent tests to fail.
-	if ct.Params().IncludeConnDisruptTest && !ct.Params().ConnDisruptTestSetup {
-		for _, client := range ct.Clients() {
-			if err := ct.DeleteConnDisruptTestDeployment(ctx, client); err != nil {
-				return err
-			}
-		}
-	}
-
 	// Report the test results.
 	return ct.report()
 }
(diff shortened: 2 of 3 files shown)


connectivity/check/test.go
```diff
@@ -311,7 +311,7 @@ func (t *Test) Run(ctx context.Context, index int) error {
 	// in which case this function executes as normal.
 	defer func() {
 		// Run all the Test's registered finalizers.
-		t.finalize()
+		t.finalize(ctx)
```
giorio94, line 314: We may not want to propagate the same test context to finalizers, as it may already have been canceled at that point. Consider for instance a long running test, which gets interrupted by the user via ^C. At that point, the context will be canceled, and possible finalizers would not be executed, possibly leaving dangling resources. It is probably better to use a separate context, associated with a tight timeout and/or wired again to the interrupts or. WDYT?

brb: Hmm, good call. Maybe for now we could pass `context.TODO()` (as it's done in https://github.com/cilium/cilium-cli/blob/main/connectivity/check/policy.go#L531-L534)?

brb: Please see the 2nd commit.

connectivity/builder/no_interrupted_connections.go
```diff
@@ -13,5 +15,15 @@ type noInterruptedConnections struct{}
...
-		WithScenarios(tests.NoInterruptedConnections())
+		WithScenarios(tests.NoInterruptedConnections()).
+		WithFinalizer(func(ctx context.Context) error {
+			// Delete the test-conn-disrupt pods immediately after the test run
+			// has finished to reduce CPU consumption (the pods generate a lot
+			// of traffic, and in most CI jobs Cilium runs with monitor
+			// aggregation disabled).
+			if !ct.Params().ConnDisruptTestSetup {
+				return ct.DeleteConnDisruptTestDeployment(ctx, ct.K8sClient())
+			}
```
giorio94, line 26: The [old logic](https://github.com/cilium/cilium-cli/blob/pr%2Fbrb%2Fearly-terminate-conn-disrupt/connectivity/check/context.go#L423-L432) could be likely removed at this point. `DeleteConnDisruptTestDeployment` currently also prints a log line which could probably converted to debug level for consistency.

connectivity/builder/no_interrupted_connections.go
```diff
@@ -13,5 +15,15 @@ type noInterruptedConnections struct{}
...
 		WithCondition(func() bool { return ct.Params().IncludeConnDisruptTest }).
-		WithScenarios(tests.NoInterruptedConnections())
+		WithScenarios(tests.NoInterruptedConnections()).
+		WithFinalizer(func(ctx context.Context) error {
+			// Delete the test-conn-disrupt pods immediately after the test run
+			// has finished to reduce CPU consumption (the pods generate a lot
+			// of traffic, and in most CI jobs Cilium runs with monitor
+			// aggregation disabled).
+			if !ct.Params().ConnDisruptTestSetup {
+				return ct.DeleteConnDisruptTestDeployment(ctx, ct.K8sClient())
```
giorio94, line 25: Should iterate over all clients, to ensure that the clean-up happens correctly also in a clustermesh environment.

connectivity/check/test.go
```diff
@@ -294,7 +294,10 @@ func (t *Test) finalize() {
 	t.Debug("Finalizing Test", t.Name())
 
 	for _, f := range t.finalizers {
-		if err := f(); err != nil {
+		// Use a detached context to make sure this call is not affected by
+		// context cancellation. Usually, finalization (e.g., netpol removal)
+		// needs to happen even when the user interrupted the program.
+		if err := f(context.TODO()); err != nil {
```
giorio94, line 300: I wonder if it could make sense to add a timeout to prevent finalizers from hanging forever in case of e.g., network connectivity problems.

connectivity/check/test.go
giorio94: Looks good to me, thanks!

connectivity/builder/no_interrupted_connections.go
brb: Oops, I keep forgetting about these :+1:

connectivity/builder/no_interrupted_connections.go
brb: Yep, I think it's safe to remove the old logic.

connectivity/check/test.go
brb: What could be a reasonable timeout? 5min?

giorio94: Hmm, I was actually thinking about a way lower value (something between 1 and 5 seconds), but it may be indeed to low if we are waiting for any operation to actually happen (like in case of policies removal). Maybe 1 minute?

giorio94 (approved):
Thanks!

christarazi (approved):

PR #2520, opened 2024-04-30
something like this

michi-covalent:


michi-covalent:
apparently you can't issue oidc tokens for pull requests from forks. back to the drawing board.

PR #2532, opened 2024-05-06
Update stable release to v0.16.7

michi-covalent:


derailed (approved):
@michi-covalent Nice work!

PR #2536, opened 2024-05-07
golangci: remove deprecated go version properties

tklauser:
Currently, the Go version for some linters is configured via separate properties. This has been deprecated in favor of the global config property `run.go` and results in the following warning.

    % golangci-lint
    WARN [config_reader] The configuration option `linters.staticcheck.go` is deprecated, please use global `run.go`.
    WARN [config_reader] The configuration option `linters.gosimple.go` is deprecated, please use global `run.go`.
    ...

In addition, the new property `run.go` defaults to the Go version defined in the `go.mod` file.

Therefore, remove the deprecated properties completely and depend on the fallback mechanism (`go.mod`).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,19 +36,13 @@ linters-settings:
       Copyright Authors of {{ PROJECT }}
   goimports:
     local-prefixes: github.com/cilium/cilium-cli
-  gosimple:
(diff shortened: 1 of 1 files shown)


derailed (approved):
@tklauser LGTM

PR #2545, opened 2024-05-11
Add support for IPv6 in connectivity tests

lel-amri:
As of today, values specified in options `--external-ip` and `--external-other-ip` are assumed to be IPv4 addresses in several places in the code. This cause some tests to fail even when a proper IPv6 configuration is specified with options `--external-cidr`, `--external-ip`, `--external-other-ip` and `--external-target`. This pull request remove the assumptions that these options holds IPv4 addresses.

### DIFF
--- a/connectivity/tests/common.go
+++ b/connectivity/tests/common.go
@@ -27,7 +27,7 @@ func FormattedAsIPv6(s string) bool {
 	return net.ParseIP(s) != nil && strings.Contains(s, ":")
 }
 
-func EnsureIPv6InBackets(s string) string {
+func EnsureIPv6InBrackets(s string) string {
 	if s[0] != '[' && FormattedAsIPv6(s) {
 		return fmt.Sprintf("[%s]", s)
 	}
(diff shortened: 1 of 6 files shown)


christarazi (approved):

connectivity/tests/common.go
```diff
@@ -4,12 +4,36 @@
...
+}
+
+func FormattedAsIPv6(s string) bool {
+	// Test with ParseIP whether this is a valid IP address. Then check that the
+	// string contains `:`. If that's the case, it must be a string formatted as
+	// and IPv6 address.
+	return net.ParseIP(s) != nil && strings.Contains(s, ":")
+}
+
+func EnsureIPv6InBackets(s string) string {
```
doniacld, line 30:
```suggestion
func EnsureIPv6InBrackets(s string) string {
```

doniacld (changes requested):
One nit comment, otherwise LGTM 🚀

lel-amri:
Hello @christarazi and @doniacld, thanks for your review and sorry for the delay.

I'd like to write some tests to cover the new code paths but I'm unsure how to proceed. Do you have some pointers to put me on the right track?

PR #2562, opened 2024-05-25
Connectivity tests concurrent logger.

viktor-kurchenko:
This PR introduces connectivity tests concurrent logger component.
The concurrency logger used for output aggregation and ordering in case of `--test-concurrency` param is > than 1.

Also, the PR contains fixes for JUnit repot name and some cleanup.

### DIFF
--- a/connectivity/check/logger.go
+++ b/connectivity/check/logger.go
@@ -1,0 +1,159 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package check
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// NewConcurrentLogger factory function that returns ConcurrentLogger.
+func NewConcurrentLogger(writer io.Writer, concurrency int) *ConcurrentLogger {
+	return &ConcurrentLogger{
+		messageCh: make(chan message),
+		writer:    writer,
+		// The concurrency parameter is used for nsTestsCh buffer size calculation.
+		// The buffer will be able to accept 10 times more unique connectivity tests
+		// than concurrency value. Write to the channel implemented in a separate
+		// goroutine to avoid deadlock in case if buffer is full.
+		nsTestsCh:        make(chan string, concurrency*10),
+		nsTestMsgs:       make(map[string][]message),
+		nsTestMsgsLock:   sync.Mutex{},
+		collectorStarted: atomic.Bool{},
+		printerDoneCh:    make(chan bool),
+	}
+}
+
+type ConcurrentLogger struct {
+	messageCh        chan message
+	writer           io.Writer
+	nsTestsCh        chan string
+	nsTestMsgs       map[string][]message
+	nsTestMsgsLock   sync.Mutex
+	collectorStarted atomic.Bool
+	printerDoneCh    chan bool
+	nsTestCount      int
+}
+
+// Start starts ConcurrentLogger internals in separate goroutines:
+// - collector: collects incoming test messages.
+// - printer: sends messages to the writer in corresponding order.
+func (c *ConcurrentLogger) Start(ctx context.Context) {
+	c.collectorStarted.Store(true)
+	go c.collector(ctx)
+	go c.printer()
+}
+
+// Stop closes incoming message channel and waits while all messages are printed.
+func (c *ConcurrentLogger) Stop() {
+	close(c.messageCh)
+	<-c.printerDoneCh
+}
+
+type message struct {
+	namespace string
+	testName  string
+	data      string
+	finish    bool
+}
+
+func (m message) nsTest() string {
+	return fmt.Sprintf("%s:%s", m.namespace, m.testName)
+}
+
+// Printf schedules message for the test to be printed.
+func (c *ConcurrentLogger) Printf(test *Test, format string, args ...interface{}) {
+	buf := &bytes.Buffer{}
+	if test.ctx.timestamp() {
+		mustFprintf(buf, timestamp())
+	}
+	mustFprintf(buf, format, args...)
+	c.messageCh <- message{
+		namespace: test.ctx.params.TestNamespace,
+		testName:  test.name,
+		data:      buf.String(),
+	}
+}
+
+// FinishTest schedules the final message for the test to be printed.
+// The message will be populated with the test log buffer if the test failed.
+func (c *ConcurrentLogger) FinishTest(test *Test) {
+	buf := &bytes.Buffer{}
+	if test.failed && test.logBuf != nil {
+		if _, err := io.Copy(buf, test.logBuf); err != nil {
+			panic(fmt.Errorf("failed to read from test log buffer: %w", err))
+		}
+	}
+	c.messageCh <- message{
+		namespace: test.Context().Params().TestNamespace,
+		testName:  test.Name(),
+		data:      buf.String(),
+		finish:    true,
+	}
+}
+
+func (c *ConcurrentLogger) collector(ctx context.Context) {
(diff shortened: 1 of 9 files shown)


connectivity/check/logger.go
```diff
@@ -0,0 +1,159 @@
...
+	c.printerDoneCh <- true
+	close(c.nsTestsCh)
+}
+
+func (c *ConcurrentLogger) printTestMessages(nsTest string) {
+	for i := 0; ; {
+		c.nsTestMsgsLock.Lock()
+		messages := c.nsTestMsgs[nsTest]
+		c.nsTestMsgsLock.Unlock()
+		if len(messages) == i {
```
nebril, line 140: I am pretty sure this is a data race. `messages` is a reference to `c.nsTestMsgs[nsTest]`, which can be appended to while checking its length or while printing data. Appends can also reallocate underlying arrays which may lead to reading stale or wrong data. I think we need to hold the lock for printing these messages if there is a possibility of an `append` happening.

nebril (changes requested):
Awesome to see this change! I think I found one issue, left feedback inline.

connectivity/check/logger.go
viktor-kurchenko: Good catch!
But I don't think we have a data race because it reads `messages` from the map on each loop iteration.
So, if the collector goroutine appends new elements into the slice (and even the slice will be replaced with the larger one) it will be read during the next loop iteration.

The loop will continue till it receives a message with the `finish` flag set to `true`.

Of course, we can extend the lock section but I want to keep it as small as possible.
Makes sense?

viktor-kurchenko: @nebril I've been inspired by your comment and pushed two additional fixes:
- fix potential deadlock [commit](https://github.com/cilium/cilium-cli/pull/2562/commits/5072f6920d39421b3eb4193b04f6da88a0dd4466)
- improve message wait logic [commit](https://github.com/cilium/cilium-cli/pull/2562/commits/0cfbabe18dd820148024843a3ce1a5a4a1a7b5f9)

Please let me know what do you think?
Thank you!

CC: @asauber

connectivity/check/logger.go
```diff
@@ -0,0 +1,187 @@
...
+			c.nsTestMsgsLock.Unlock()
+			// wait for new test messages
+			<-c.printerWaitCh
+			continue
+		}
+		c.nsTestMsgsLock.Unlock()
+		for ; i < len(messages); i++ {
+			mustFprintf(c.writer, messages[i].data)
+			if messages[i].finish {
+				c.nsTestCount++
```
asauber, line 176: `nsTestCount` would be better named `nsTestFinishCount`

connectivity/check/logger.go
```diff
@@ -0,0 +1,187 @@
...
+	// read messages while the collector is working
+	for c.collectorStarted.Load() {
+		// double-check if there are new messages to avoid
+		// deadlock reading from the `nsTestsCh` channel
+		if c.collectedTestCount() > c.nsTestCount {
+			c.printTestMessages(<-c.nsTestsCh)
+		}
+	}
+	// collector stopped but there still might be messages to print
+	for c.collectedTestCount() > c.nsTestCount {
```
asauber, line 147: I think this condition (and the same one above), would be more readable if reversed. What do you think?

`for c.nsTestFinishedCount < c.collectedTestCount()`

connectivity/check/logger.go
asauber: The `collectedTestCount()` change looks good to me, there is definitely the possibility of `<-c.nsTestsCh` being written to after `collectorStarted` is set to false.

However, the `printerWaitCh` change does _not_ look necessary to me. @nebril this loop is `for i := 0; ;`, where `i` is the number of currently seen messages for the given test name. Until the test sends a message with `finish` set, we loop here while looking at `if len(messages) == i` where `i` is the "currenty seen" seen message. Thus, it doesn't matter if `messages` was appended to before we reached this point, the next iteration will pick it up (e.g. `i == 42` and the 44th message is still being written. That's fine, this printer goroutine will busy loop for a bit). @viktor-kurchenko given this, I would suggest renaming `i` to `printedMessageIndex`. It's not exactly Go style, but I would say dynamic indices always need some extra information to be readable.

Therefore, I would suggest removing this `<-c.printerWaitCh` concept. Although the logic looks correct (the given test name will always attempt to unlock if we are blocked here), it adds some unnecessary complexity.

asauber (approved):

connectivity/check/logger.go
viktor-kurchenko: Renamed, thanks!

connectivity/check/logger.go
viktor-kurchenko: Makes sense! Done, thanks!

connectivity/check/logger.go
viktor-kurchenko: Hmm, that's very interesting!!!
I think it works because write to slice and slice read from the map protected by lock:
- [Write protection](https://github.com/cilium/cilium-cli/pull/2562/files#diff-1935b7c291fa20da9f0f1085798b51cf6c45b1735e75c32b4ea048d1f5d0c214R111-R120)
- [Slice read protection](https://github.com/cilium/cilium-cli/pull/2562/files#diff-1935b7c291fa20da9f0f1085798b51cf6c45b1735e75c32b4ea048d1f5d0c214R154-R156)

Also, I think the below code snippet is more similar to the logger approach:
```
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	k := "1"
	m := make(map[string][]int, 0)
	m[k] = make([]int, 0, 1_000_000)
	l := sync.Mutex{}

	go func() {
		for i := 0; i < 1_000_000; i++ {
			l.Lock()
			m[k] = append(m[k], i)
			l.Unlock()
		}
	}()
	go func() {
		for {
			l.Lock()
			a := m[k]
			l.Unlock()
			for i := range a {
				fmt.Print(a[i])
			}
		}
	}()

	time.Sleep(time.Second)
}
```

If you remove the lock, you will see a data race.
I've made the slice capacity big enough to guarantee that the slice pointer won't change.

nebril: Slice pointer should be allowed to change, but I cannot get this code to run with a race even with capacity set to 0. Maybe `a := m[k]` creates a new slice header, which is unaffected by appends in another goroutine? Seems like you are right though, thanks a lot for taking the time to write this example.

nebril (approved):

connectivity/check/logger.go
viktor-kurchenko: Yeah, if you change capacity to 0 it becomes more obvious: reader routine will work with a slice copy (eventually, not always).
But when capacity is set to the max allowed value then the writer and reader always work with the same slice.
However, the reader works with a slice shallow copy which doesn't change length even when the writer appends a new element.

Try this:
```
package main

import "fmt"

func main() {
	s := make([]int, 0, 10)
	for i := 0; i < 5; i++ {
		s = append(s, i)
	}
	m := map[string][]int{
		"1": s,
	}

	s1 := m["1"]
	fmt.Printf("\nno changes\n")
	display(m, s1)

	m["1"] = append(m["1"], 6)
	fmt.Printf("\nafter append to map\n")
	display(m, s1)

	s1 = append(s1, 7)
	fmt.Printf("\nafter append to shallow copy\n")
	display(m, s1)
}

func display(m map[string][]int, s []int) {
	fmt.Printf("in map len: %d, addr: %p -> ", len(m["1"]), m["1"])
	for i := range m["1"] {
		fmt.Printf("%d ", m["1"][i])
	}
	fmt.Printf("\nshallow copy len: %d, addr: %p -> ", len(s), s)
	for i := range s {
		fmt.Printf("%d ", s[i])
	}
	println()
}
```

PR #2568, opened 2024-05-29
connectivity: Add local redirect policy tests

aditighag:
This PR introduces local redirect policy tests to the connectivity suite.

In order to break the circular dependency, there is a separate PR on cilium/cilium repo that enables these tests using a custom version of cilium-cli obtained from this PR changes.  Here are successful test runs from the cilium/cilium PR CI runs - 

Runs after addressing review comments - 
https://github.com/cilium/cilium/actions/runs/9409627836/job/25920752091
https://github.com/cilium/cilium/actions/runs/9409627836/job/25919911738

https://github.com/cilium/cilium/actions/runs/9387516652/job/25850621644
https://github.com/cilium/cilium/actions/runs/9377202073/job/25819021276
https://github.com/cilium/cilium/actions/runs/9375976798/job/25815109405 
https://github.com/cilium/cilium/actions/runs/9377202073/job/25818391762

### DIFF
--- a/connectivity/check/peer.go
+++ b/connectivity/check/peer.go
@@ -9,9 +9,10 @@ import (
 	"net/url"
 	"strconv"
 
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/cilium/cilium/api/v1/flow"
 	ciliumv2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
-	corev1 "k8s.io/api/core/v1"
 
 	"github.com/cilium/cilium-cli/k8s"
 	"github.com/cilium/cilium-cli/utils/features"
@@ -494,6 +495,61 @@ func (he httpEndpoint) Labels() map[string]string {
 }
 
 func (he httpEndpoint) FlowFilters() []*flow.FlowFilter {
+	return nil
+}
+
+type LRPFrontend struct {
+	name string
+	ip   string
+	port string
+}
+
+func NewLRPFrontend(frontend ciliumv2.RedirectFrontend) *LRPFrontend {
+	var lf LRPFrontend
+	if f := frontend.AddressMatcher; f != nil {
+		lf.ip = f.IP
+		lf.port = f.ToPorts[0].Port
+		lf.name = fmt.Sprintf("%s:%s", lf.ip, lf.port)
+
+		return &lf
+	}
+
+	return nil
+}
+
+func (l LRPFrontend) Name() string {
+	return l.name
+}
+
+func (l LRPFrontend) Scheme() string {
+	return "http"
+}
+
+func (l LRPFrontend) Path() string {
+	return ""
+}
+
+func (l LRPFrontend) Address(features.IPFamily) string {
+	return l.ip
+}
+
+func (l LRPFrontend) Port() uint32 {
+	p, err := strconv.Atoi(l.port)
+	if err != nil {
+		return 0
+	}
+	return uint32(p)
+}
+
+func (l LRPFrontend) HasLabel(string, string) bool {
+	return false
+}
+
+func (l LRPFrontend) Labels() map[string]string {
+	return nil
+}
+
+func (l LRPFrontend) FlowFilters() []*flow.FlowFilter {
 	return nil
 }
 
--- a/connectivity/tests/lrp.go
+++ b/connectivity/tests/lrp.go
@@ -1,0 +1,190 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package tests
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	v2 "github.com/cilium/cilium/pkg/k8s/apis/cilium.io/v2"
+
+	"github.com/cilium/cilium-cli/connectivity/check"
+	"github.com/cilium/cilium-cli/defaults"
+	"github.com/cilium/cilium-cli/utils/features"
+	"github.com/cilium/cilium-cli/utils/wait"
+)
+
+// LRP runs test scenarios for local redirect policy. It tests local redirection
+// connectivity from test source pods to LRP frontend.
+//
+// It tests connectivity with the configured skipRedirectFromBackend flag for:
+// - client pods to LRP frontend
+// - LRP backend pods to LRP frontend
+func LRP(skipRedirectFromBackend bool) check.Scenario {
+	return lrp{skipRedirectFromBackend: skipRedirectFromBackend}
+}
+
+type lrp struct {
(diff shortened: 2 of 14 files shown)


connectivity/tests/lrp.go
```diff
@@ -0,0 +1,190 @@
...
+			backends, ok := backendsMap[node]
+			if !ok {
+				// No LRP backend pods deployed on this node.
+				continue
+			}
+			cmd := strings.Split("cilium bpf lb list -o json", " ")
+			stdout, err := ciliumPod.K8sClient.ExecInPod(ctx, ciliumPod.Pod.Namespace, ciliumPod.Pod.Name, defaults.AgentContainerName, cmd)
+			if err != nil {
+				t.Fatal("Failed to run cilium bpf lb list -o json command:", err)
+			}
```
christarazi, line 138: This pattern is not ideal as it is relies on finicky timeouts and mechanisms to parse output in an error-prone way which we have seen cause many flakes in the Ginkgo suite over time. Ofttimes with timeouts, it's inevitable that we will have a PR to bump it because the initial value was too low. Then we get to a very bloated duration which leads to tests taking longer and longer over time. With parsing command-line outputs, it's inevitable that we get flakes where the command doesn't execute properly, i.e. exiting with a strange error code, empty output, etc, that we presumed are ultimately underlying infrastructure problems. 

Now we are porting this sort of logic over to the Cilium CLI, which goes against the goals of the Cilium CLI that's meant to be a less error-prone style of testing than the Ginkgo suite.

On a more practical note, it seems that this logic is trying to detect when the LRP entries have been created. Are there alternative mechanisms to detecting when LRP is "ready" besides dumping the BPF map? For example, with CNPs we can detect the policy revision number has incremented. If an alternative doesn't exist today, it might be worth discussing what it would look like and temporarily moving ahead with this PR as-is (pending other feedback of course).

aditighag: >For example, with CNPs we can detect the policy revision number has incremented.

Is that done in userspace or datapath (BPF maps)? Does that correspond to individual policy entries? We don't have such equivalent metric on LB side of things. 

That's a fair question. 

Some more context on the datapath side of things in general - 

Checking userspace state is not enough, as it can take some time to translate plumb policies into datapath. You wouldn't want to have a busy loop checking if a test curl command eventually succeeded (or fails for that matter) within a timeout. So the more definite signals are checking for BPF entries (LB, egress gateway, etc tests follow similar logic). Another useful aspect with these kind of checks is that they are the next logical step in debugging when connectivity tests fail. For example, if I'm debugging an LB connectivity issue, I would go about checking BPF LB map entries, and whether they were plumbed correctly.

aditighag: >In simpler terms, if you change/apply a policy, all endpoints wait until they have been regenerated for the policy change (i.e. wait for the revision to increment). Once they have all finished regenerating, the policy revision exposed via the API is incremented which is what the CLI watches.

Thanks for the context. I don't see how the same logic could translate to other BPF maps (e.g., LB). Does that mean we would have to serialize service LB plumbing? The point around checking BPF entries as a logical step for debugging LB still stands though. 

>Ofttimes with timeouts, it's inevitable that we will have a PR to bump it because the initial value was too low. 

Going back to your concerns around timeouts, won't you have similar timeouts while watching policy revision bumps?

christarazi: > Going back to your concerns around timeouts, won't you have similar timeouts while watching policy revision bumps?

Technically yes, but the waiting mechanism is built into `cilium-dbg policy wait ...`, so there's a lot less testing code that needs to be written which means a lot less surface area for potential problems/errors.

christarazi: > So the more definite signals are checking for BPF entries (LB, egress gateway, etc tests follow similar logic)

It sounds like some sort of first-class 'ready' signal would benefit LB, LRP, egress gateway features all the same. Anyway, it's something to think about and for the immediate PR, it doesn't sound like we have other alternatives so we are left with the current implementation for now.

aditighag: > > Going back to your concerns around timeouts, won't you have similar timeouts while watching policy revision bumps?
> 
> Technically yes, but the waiting mechanism is built into `cilium-dbg policy wait ...`, so there's a lot less testing code that needs to be written which means a lot less surface area for potential problems/errors.

Ah, TIL!! Thanks for highlighting this.

christarazi (approved):

connectivity/tests/lrp.go
```diff
@@ -0,0 +1,190 @@
...
+
+	ensureBPFLBEntries := func() error {
+		for _, ciliumPod := range ct.CiliumPods() {
+			node := ciliumPod.Pod.Spec.NodeName
+			backends, ok := backendsMap[node]
+			if !ok {
+				// No LRP backend pods deployed on this node.
+				continue
+			}
+			cmd := strings.Split("cilium bpf lb list -o json", " ")
```
derailed, line 134: nit: The command is invariant might as well move it out of the loop and split once

connectivity/tests/lrp.go
```diff
@@ -0,0 +1,190 @@
...
+			for _, backend := range backends {
+				found := false
+				for fe, beStr := range lbMap {
+					if fe != frontend {
+						continue
+					}
+					for _, be := range beStr {
+						if strings.Contains(be, backend) {
+							found = true
+							break
```
derailed, line 167: nit; not sure if this is a deal or not but we would continue looping thru the map when be is found. 
Couldn't we just index and loop into the map ie lbMap[frontend]  ok? -> loop?

connectivity/check/peer.go
```diff
@@ -491,6 +492,60 @@ func (he httpEndpoint) FlowFilters() []*flow.FlowFilter {
...
+		lf.ip = f.IP
+		lf.port = f.ToPorts[0].Port
+
+		return &lf
+	}
+
+	return nil
+}
+
+func (l LRPFrontend) Name() string {
```
derailed, line 513: nit: Is the name ever set?

derailed (approved):
@aditighag Nice work!

connectivity/tests/lrp.go
aditighag: Good catch! Simplified the logic.

aditighag (commented):
Thanks for the review. @derailed Addressed the review comments.

connectivity/tests/lrp.go
christarazi: @aditighag I remembered that I didn't ask for a tracking issue regarding this, but it would be nice to track the work to migrating this sort of testing style to something less error-prone, so we don't forget about it.

aditighag: I've filed - https://github.com/cilium/cilium/issues/33372.

PR #2569, opened 2024-05-30
Fix clusteremesh status retrieval on Cilium v1.14 when kvstoremesh is enabled

giorio94:
The blamed commit broke clustermesh status retrieval on Cilium v1.14 when kvstoremesh is enabled. Indeed, although it already included a check to try to figure out whether the kvstoremesh status was not yet exposed by a given Cilium version, it did not account for v1.14 using a completely separate image (and binary name) for kvstoremesh. Let's fix this by adding the appropriate exception.

Fixes: fa7e30e1a5db ("clustermesh: output kvstoremesh status as part of clustermesh status")


michi-covalent (approved):
looks innocent enough

squeed (approved):

PR #2591, opened 2024-06-11
Sanitize the autodetected cluster name

giorio94:
Replace underscores and dots with dashes, as they are not allowed characters for cluster names. 

~~Additionally perform the same validation already implemented by the latest Cilium version, to exit early and provide a helpful message in case of errors.~~

Edit. I've removed these changes as bumping the Cilium dependency seems to cause a couple of tests to fail quite consistently.

~~(The large diff count is caused by bumping the Cilium dependency to the current main, to reuse the already available validation function. Please review commit by commit)~~

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -214,10 +214,10 @@ jobs:
       kind_config_2: .github/kind-config-2.yaml
       # helm/kind-action will override the "name:" provided in the kind config with "chart-testing" unless these are
       # specified as inputs. These must also match the suffix here for CLUSTER1 and CLUSTER2.
-      CLUSTER_NAME_1: c.1
-      CLUSTER_NAME_2: c.2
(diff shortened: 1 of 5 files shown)


sayboras (approved):
Just a note that cluster name in kubeconfig can be manipulated anytime, but I guess it's ok as part of installation.

viktor-kurchenko (approved):
LGTM

derailed (approved):
@giorio94 LGTM

giorio94:
Let me try dropping the commit which bumps the cilium dependency, to see if the CI failures resolve (as they cannot be possibly caused by the other changes). That also leads to dropping the extra validation, which I'll add back in a separate PR.

giorio94:
@doniacld Gentle ping :pray:

giorio94:
@doniacld Gentle ping 🙏

doniacld (approved):

Issue #1907, opened 2023-08-14
Improve error messages for "release not found"


asauber:
In codepaths where the CLI might fail with `release not found`, such as 

```
Error: Unable to enable ClusterMesh: release: not found
```

we should include more information in this message to allow for a resolution.

In most cases, the user is working with a cluster without a Helm-based Cilium install, or they haven't specified the `--namespace`.

Instructing a retry with `CILIUM_CLI_MODE=classic` can also be included until full removal of classic mode.

dhanusaputra:
hi, can I help with this please?

network-charles:
I’m currently having this issue while attempting to enable clustermesh. What namespace are you referring to, “kube-system”?

cnvergence:
I would like to help with that, as I encountered the issue already fixed in https://github.com/cilium/cilium-cli/issues/2274
I will work on this PR: https://github.com/cilium/cilium-cli/pull/2597

PR #2597, opened 2024-06-13
chore: improve error messages for Helm commands

cnvergence:
Fixes: https://github.com/cilium/cilium-cli/issues/1907

### DIFF
--- a/cli/clustermesh.go
+++ b/cli/clustermesh.go
@@ -56,7 +56,7 @@ func newCmdClusterMeshStatus() *cobra.Command {
 
 			cm := clustermesh.NewK8sClusterMesh(k8sClient, params)
 			if _, err := cm.Status(context.Background()); err != nil {
-				fatalf("failed to determine status:  %s", err)
+				fatalf("Failed to determine status:  %s", err)
 			}
 			return nil
 		},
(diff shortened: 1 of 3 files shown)


cli/clustermesh.go
```diff
@@ -243,7 +243,7 @@ func newCmdClusterMeshEnableWithHelm() *cobra.Command {
 			params.HelmReleaseName = helmReleaseName
 			ctx := context.Background()
 			if err := clustermesh.EnableWithHelm(ctx, k8sClient, params); err != nil {
-				fatalf("Unable to enable ClusterMesh: %s", err)
+				fatalf("Unable to enable ClusterMesh: %s, please ensure that you have specified the --namespace and --helm-release-name if you are working with a cluster with customized Cilium installation.", err)
```
derailed, line 246: Might be an easier to grok if these read something like `failed to enable/disable xxx`

derailed (commented):
@cnvergence Thank you for these updates!

cli/clustermesh.go
cnvergence: good point, also I did mistake by starting it with uppercase

cnvergence: I will also include other error messages

derailed: fatalf is just a wrapper to print a message to stderr and exit. Thus messages can start with upper case.

cnvergence: correct, I restored it

cnvergence:
@derailed PTAL again :)

derailed (approved):
@cnvergence Thank you for the updates!

cnvergence:
hi @derailed, shall I fix something in this PR?

PR #2605, opened 2024-06-18
connectivity: Add Port Range Tests 

nathanjsweet:
- Add port range tests for all port-based connectivity tests.

### DIFF
--- a/connectivity/builder/client_egress_expression.go
+++ b/connectivity/builder/client_egress_expression.go
@@ -8,16 +8,33 @@ import (
 
 	"github.com/cilium/cilium-cli/connectivity/check"
 	"github.com/cilium/cilium-cli/connectivity/tests"
+	"github.com/cilium/cilium-cli/utils/features"
 )
 
 //go:embed manifests/client-egress-to-echo-expression.yaml
 var clientEgressToEchoExpressionPolicyYAML string
 
+//go:embed manifests/client-egress-to-echo-expression-port-range.yaml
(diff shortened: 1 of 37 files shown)


connectivity/builder/client_egress_expression.go
```diff
@@ -8,16 +8,33 @@ import (
...
+//go:embed manifests/client-egress-to-echo-expression-port-range.yaml
+var clientEgressToEchoExpressionPolicyPortRangeYAML string
+
 type clientEgressExpression struct{}
 
 func (t clientEgressExpression) build(ct *check.ConnectivityTest, _ map[string]string) {
+	clientEgressExpressionTest(ct, false)
+	if ct.Features[features.PortRanges].Enabled {
+		clientEgressExpressionTest(ct, true)
+	}
```
christarazi, line 26: Is this running the same test twice, one with and without port ranges? I see that this pattern is applied to all the other tests that were modified. Wouldn't this in theory double the amount of time to run tests? If so, is this really necessary? Why not just run the test with port ranges enabled if it's available?

nathanjsweet: Running these tests in production CI doesn't consume much additional time. I am concerned about removing simple port tests from the connectivity tests in the long term as it could lead us to miss a bug.

christarazi (approved):

derailed (approved):
@nathanjsweet Nice work!!

youngnick (approved):

PR #2610, opened 2024-06-20
connectivity: Add pod-to-pod-no-frag to check MTU misconfigurations

brb:
CI run - https://github.com/cilium/cilium/pull/33286. Only ci-ipsec and ci-eks failures are relevant. I will fix them in a PR for cilium/cilium, which bumps CLI vsn.

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -1029,7 +1029,7 @@ func (ct *ConnectivityTest) CurlCommandParallelWithOutput(peer TestPeer, ipFam f
 	return cmd
 }
 
-func (ct *ConnectivityTest) PingCommand(peer TestPeer, ipFam features.IPFamily) []string {
+func (ct *ConnectivityTest) PingCommand(peer TestPeer, ipFam features.IPFamily, extraArgs ...string) []string {
 	cmd := []string{"ping", "-c", "1"}
 
 	if ipFam == features.IPFamilyV6 {
@@ -1040,7 +1040,10 @@ func (ct *ConnectivityTest) PingCommand(peer TestPeer, ipFam features.IPFamily) 
 		cmd = append(cmd, "-W", strconv.FormatFloat(connectTimeout, 'f', -1, 64))
 	}
 
+	cmd = append(cmd, extraArgs...)
+
 	cmd = append(cmd, peer.Address(ipFam))
+
 	return cmd
 }
 
--- a/connectivity/tests/pod.go
+++ b/connectivity/tests/pod.go
@@ -6,9 +6,17 @@ package tests
 import (
 	"context"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/cilium/cilium-cli/connectivity/check"
 	"github.com/cilium/cilium-cli/utils/features"
+)
+
+const (
+	HdrSizeICMPEcho = 8
+	HdrSizeIPv4     = 20
+	HdrSizeIPv6     = 40
 )
 
 // PodToPod generates one HTTP request from each client pod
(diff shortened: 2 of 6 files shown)


connectivity/check/context.go
```diff
@@ -1030,6 +1030,12 @@ func (ct *ConnectivityTest) CurlCommandParallelWithOutput(peer TestPeer, ipFam f
 }
 
 func (ct *ConnectivityTest) PingCommand(peer TestPeer, ipFam features.IPFamily) []string {
```
bimmlerd, line 1032: nit: why not just `PingCommand(peer TestPeer, ipFam features.IPFamily, extraArgs ...string)`?

connectivity/tests/pod.go
```diff
@@ -165,3 +167,81 @@ func (s *podToPodWithEndpoints) curlEndpoints(ctx context.Context, t *check.Test
...
+	t.Debugf("Running %s", strings.Join(cmd, " "))
+	mtuBytes, err := client.K8sClient.ExecInPod(ctx, client.Pod.Namespace,
+		client.Pod.Name, "", cmd)
+	if err != nil {
+		t.Fatalf("Failed to get route MTU in pod %s: %s", client, err)
+	}
+
+	// Derive MTU from pod iface instead
+	if mtuBytes.Len() == 0 {
+		cmd := []string{
```
bimmlerd, line 205: maybe easier to directly `cat /sys/class/net/eth0/mtu` instead of shelling out to `ip`?

jschwinger233 (approved):

connectivity/tests/pod.go
```diff
@@ -165,3 +167,81 @@ func (s *podToPodWithEndpoints) curlEndpoints(ctx context.Context, t *check.Test
...
+		}
+		t.Debugf("Running %s", strings.Join(cmd, " "))
+		mtuBytes, err = client.K8sClient.ExecInPod(ctx, client.Pod.Namespace,
+			client.Pod.Name, "", cmd)
+		if err != nil {
+			t.Fatalf("Failed to get eth0 MTU in pod %s: %s", client, err)
+		}
+	}
+
+	mtuStr := strings.TrimRight(mtuBytes.String(), "\n\r")
```
bimmlerd, line 217: nit: `strings.TrimSpace`?

connectivity/tests/pod.go
```diff
@@ -165,3 +167,81 @@ func (s *podToPodWithEndpoints) curlEndpoints(ctx context.Context, t *check.Test
...
+	}
+	t.Debugf("Running %s", strings.Join(cmd, " "))
+	mtuBytes, err := client.K8sClient.ExecInPod(ctx, client.Pod.Namespace,
+		client.Pod.Name, "", cmd)
+	if err != nil {
+		t.Fatalf("Failed to get route MTU in pod %s: %s", client, err)
+	}
+
+	// Derive MTU from pod iface instead
+	if mtuBytes.Len() == 0 {
```
bimmlerd, line 204: should probably strip whitespace from `mtuBytes` first?

bimmlerd (commented):
couple of nits

connectivity/tests/pod.go
brb: I tested that in the case of no MTU in a route, the `mtuBytes.Len()` is zero. But to be on the same side, let me clear trim spaces first.

christarazi:
I'm going to repost a previous review comment from another PR because I am seeing the same pattern being repeated here: https://github.com/cilium/cilium-cli/pull/2568#discussion_r1628210710

Is this really the best place to have this kind of test? I don't think I understand the point of moving away from Ginkgo-style tests if we are going to port the same logic and testing pattern into the CLI.

christarazi (commented):
My ci-structure review:

I'm going to repost a previous review comment from another PR because I am seeing the same pattern being repeated here: https://github.com/cilium/cilium-cli/pull/2568#discussion_r1628210710

Is this really the best place to have this kind of test? I don't think I understand the point of moving away from Ginkgo-style tests if we are going to port the same logic and testing pattern into the CLI.

connectivity/tests/pod.go
```diff
@@ -165,3 +167,85 @@ func (s *podToPodWithEndpoints) curlEndpoints(ctx context.Context, t *check.Test
...
+		// Make sure that the server pod is on another node than client
+		if pod.Pod.Status.HostIP != client.Pod.Status.HostIP {
+			server = pod
+			break
+		}
+	}
+
+	t.ForEachIPFamily(func(ipFam features.IPFamily) {
+		t.NewAction(s, fmt.Sprintf("ping-%s", ipFam), client, server, ipFam).Run(func(a *check.Action) {
+			payloadSize := mtu - 8
```
tommyp1ckles, line 237: nit: For readability, it would be helpful to have these as constants describing what header they're the size of (i.e. this and the ip4/ip6 hdr size below) - to avoid having to scroll up for the comments.

tommyp1ckles (approved):

brb:
> Is this really the best place to have this kind of test? I don't think I understand the point of moving away from Ginkgo-style tests if we are going to port the same logic and testing pattern into the CLI.

@christarazi I get your concerns regarding flakiness. Unfortunately, with Ginkgo deprecation we don't have any other place to add such e2e tests. There was an effort to create a E2E framework during the hyperjump, but it turned to be a massive investment. Maybe this topic can be picked again by some working group.

At the same time, IMO such tests are helpful for end users to diagnose non-obvious Cilium connectivity issues (which misconfigured MTU can lead to). And what I hear from users that they extensively use `cilium-cli connectivity test` to rely on infra deployment correctness.

(Ginkgo/gomega as a testing framework is / was a terrible foundation for Cilium's E2E tests. The argumentation https://www.reddit.com/r/golang/comments/1azj63h/comment/ks1srp2/ resonates well with us).

christarazi (approved):
@brb 

> I get your concerns regarding flakiness. Unfortunately, with Ginkgo deprecation we don't have any other place to add such e2e tests. There was an effort to create a E2E framework during the hyperjump, but it turned to be a massive investment. Maybe this topic can be picked again by some working group.

I am mostly worried about lines 198-230 in `connectivity/tests/pod.go` because it seems to me that checking for the MTU on the node itself for the pod doesn't necessarily need to be done at this level. Couldn't it be validated at endpoint creation time that the appropriate MTU value was passed to the container?

The lines beyond the above lines mentioned is fine because IMO that is testing potential fragmentation issues with MTU. Ideally, that should be done in a BPF program unit test IIUC, but this is better than issuing ad-hoc commands to validate a parameter value (i.e. MTU).

It would be nice to have an issue to track the work to migrate to a better testing style, otherwise this code will just sit here until it causes us so many flakes that we then have to look at it again. (I realize I didn't mention this in the previous PR I posted a review comment like this to; will do that there as well).

For now I will approve to unblock because it seems the tradeoff of not having this sort of coverage is worse than not having it at all while we build better tests. Ideally we aren't adding more tests with this pattern going forward.

bimmlerd (approved):

brb:
> Couldn't it be validated at endpoint creation time that the appropriate MTU value was passed to the container?

Unfortunately, noup. In-between pod-to-pod there are multiple network devices involved (some managed by Cilium), which MTU misconfiguration might not be reflected in an endpoint's MTU.

> Ideally, that should be done in a BPF program unit test IIUC

The BPF unit tests do not involve netdevs. Perhaps, we should start thinking about introducing integration tests, in which cilium-agent loads BPF progs via pkg/datapath and then we inject a packet (e.g., via libpcap). That could be a replacement for the test cases from this PR. Maybe a topic for the next hive time (if yes, which WG?)?

PR #2626, opened 2024-06-25
gke: Set --test-concurrency to 5

michi-covalent:
We've been baking in this flag with the Kind workflow. Let's start using it more widely.


michi-covalent:
let's use #2639 instead

Issue #2553, opened 2024-05-16
mark multicluster.yaml as required


michi-covalent:
i removed multicluster.yaml from the required status check list since it's been failing a lot.
```
multicluster.yaml https://github.com/cilium/cilium-cli/actions/workflows/multicluster.yaml?query=branch%3Amain+event%3Aschedule++
from                to                  duration success rate
2024-05-15 21:04:04 2024-05-16 15:03:42 22m27s   🙀 25% 1/4
2024-05-14 21:03:54 2024-05-16 15:03:42 22m26s   🙀 25% 2/8
2024-05-12 21:03:41 2024-05-16 15:03:42 24m6s    🙀 38% 6/16
2024-05-08 21:03:51 2024-05-16 15:03:42 24m25s   🙀 41% 13/32
2024-04-30 21:04:32 2024-05-16 15:03:42 24m45s   🙀 42% 27/64
```
i chatted a bit with @giorio94 offline about how to mitigate the issue. here are some options:

- disable flow validation.
- remove multicluster.yaml altogether (assuming https://github.com/cilium/cilium/blob/main/.github/workflows/conformance-clustermesh.yaml gives us enough coverage).
- leave it as not-required until either we find somebody to look into the actual issue or https://github.com/cilium/design-cfps/pull/9 gets implemented and cilium-cli gets merged into cilium repo.

i'll leave it up to @cilium/sig-clustermesh to decide how to handle this 🚀🙏

Issue #2518, opened 2024-04-30
CI: multicluster no-policies flaky


michi-covalent:
## CI failure

- sample failure: https://github.com/cilium/cilium-cli/actions/runs/8885391193/job/24396640159

- Upload the zip file generated from that test failure.

[cilium-sysdump-out.zip.zip](https://github.com/cilium/cilium-cli/files/15167205/cilium-sysdump-out.zip.zip)

- Include any output from logs that you think may be relevant (to ease GitHub searches).

```
% ./ci-dashboard show cilium cilium-cli -w multicluster.yaml
multicluster.yaml https://github.com/cilium/cilium-cli/actions/workflows/multicluster.yaml?query=branch%3Amain+event%3Aschedule++
from                to                  duration success rate
2024-04-29 15:04:02 2024-04-30 09:03:56 22m44s   🤨 50% 2/4
2024-04-28 15:04:35 2024-04-30 09:03:56 23m35s   🤨 50% 4/8
2024-04-26 15:04:11 2024-04-30 09:03:56 23m37s   🙀 44% 7/16
2024-04-22 15:03:48 2024-04-30 09:03:56 24m39s   🙀 38% 12/32
2024-04-14 15:24:54 2024-04-30 09:03:56 25m22s   🙀 42% 27/64

failed jobs
job name                                        failure count
Multicluster Installation and Connectivity Test 37

failed steps
step name                                   failure count
Wait for test job                           36
Create gcloud-free kubeconfig for cluster 2 1
Post-test information gathering             1
Clean up GKE                                1

cancelled steps
step name failure count

failed tests
test name   failure count
no-policies 34

error logs
error message count
```

PR #2633, opened 2024-06-26
connectivity: Ensure Hubble Relay is connected to all the nodes

michi-covalent:
Sometimes it takes a while for all the nodes to get connected to Hubble Relay in the multicluster workflow [^1]. If Hubble is enabled, don't start running connectivity tests until all the nodes are connected to avoid flow validation failures.

[^1]: https://github.com/cilium/cilium-cli/actions/runs/9664927814/job/26660784590#step:21:1456

Fixes: #2518
Fixes: #2553

### DIFF
--- a/connectivity/check/context.go
+++ b/connectivity/check/context.go
@@ -653,10 +653,21 @@ func (ct *ConnectivityTest) enableHubbleClient(ctx context.Context) error {
 			ct.Fail("In --flow-validation=strict mode, Hubble must be available to validate flows")
 			return fmt.Errorf("hubble is not available: %w", err)
 		}
-	} else {
-		ct.Infof("Hubble is OK, flows: %d/%d", status.NumFlows, status.MaxFlows)
+		return nil
 	}
 
+	for status.GetNumUnavailableNodes().GetValue() > 0 {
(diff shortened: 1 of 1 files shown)


connectivity/check/context.go
```diff
@@ -653,10 +653,21 @@ func (ct *ConnectivityTest) enableHubbleClient(ctx context.Context) error {
...
 		}
-	} else {
-		ct.Infof("Hubble is OK, flows: %d/%d", status.NumFlows, status.MaxFlows)
+		return nil
 	}
 
+	for status.GetNumUnavailableNodes().GetValue() > 0 {
+		ct.Infof("Waiting for %d nodes to become available: %s",
+			status.GetNumUnavailableNodes().GetValue(), status.GetUnavailableNodes())
+		time.Sleep(5 * time.Second)
```
christarazi, line 662: Is there something special about 5s?

christarazi (approved):

connectivity/check/context.go
michi-covalent: no not really, i just picked a number and hardcoded it. i thought about adding a flag to make the retry interval configurable, but it felt like an overkill. what do you think? 💭

PR #2635, opened 2024-06-26
sysdump: fix no Cilium output if operator was not detected

marseel:
Fixes regression from #1776

### DIFF
--- a/cli/sysdump.go
+++ b/cli/sysdump.go
@@ -34,8 +34,13 @@ func newCmdSysdump(hooks sysdump.Hooks) *cobra.Command {
 			if sysdumpOptions.CiliumNamespace == "" && cmd.Flags().Changed("namespace") {
 				sysdumpOptions.CiliumNamespace = namespace
 			}
-			if sysdumpOptions.CiliumOperatorNamespace == "" && cmd.Flags().Changed("namespace") {
-				sysdumpOptions.CiliumOperatorNamespace = namespace
+			if sysdumpOptions.CiliumOperatorNamespace == "" {
+				if cmd.Flags().Changed("namespace") {
+					sysdumpOptions.CiliumOperatorNamespace = namespace
+				} else {
+					// Assume the same namespace for operator as for agent if not specified
+					sysdumpOptions.CiliumOperatorNamespace = sysdumpOptions.CiliumNamespace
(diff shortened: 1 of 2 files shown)


michi-covalent (commented):
we chatted about this a bit offline. ideally `--cilium-namespace` should set namespace for both `cilium-agent` and `cilium-operator` unless `--cilium-operator-namespace` is also specified. could we do something like this:
```
% git diff
diff --git a/cli/sysdump.go b/cli/sysdump.go
index 2069e548..de6fc8d3 100644
--- a/cli/sysdump.go
+++ b/cli/sysdump.go
@@ -34,8 +34,12 @@ func newCmdSysdump(hooks sysdump.Hooks) *cobra.Command {
                        if sysdumpOptions.CiliumNamespace == "" && cmd.Flags().Changed("namespace") {
                                sysdumpOptions.CiliumNamespace = namespace
                        }
-                       if sysdumpOptions.CiliumOperatorNamespace == "" && cmd.Flags().Changed("namespace") {
-                               sysdumpOptions.CiliumOperatorNamespace = namespace
+                       if sysdumpOptions.CiliumOperatorNamespace == "" {
+                               if cmd.Flags().Changed("namespace") {
+                                       sysdumpOptions.CiliumOperatorNamespace = namespace
+                               } else {
+                                       sysdumpOptions.CiliumOperatorNamespace = sysdumpOptions.CiliumNamespace
+                               }
                        }
                        // Honor --helm-release-name global flag in case it is set and --cilium-helm-release-name is not set
                        if sysdumpOptions.CiliumHelmReleaseName == "" && cmd.Flags().Changed("helm-release-name") {
```

marseel:
Hey @michi-covalent 
I thought about it a bit more with a fresh brain. 
I've added your snippet as a separate commit, but I still believe that having "&& -> ||" change makes sense.
If user does not specify a namespace at all and we fail to detect only one of agent/operator namespaces (for example if someone deletes operator deployment) we might still want to run agent tasks.

I've also opened https://github.com/cilium/cilium-cli/pull/2645 which seems like a reasonable next-step to improve the sysdump experience for users. Maybe good-first-time issue :) ?

michi-covalent (approved):
🚀

PR #2636, opened 2024-06-26
Add support for passing annotations to `clustermesh-apiserver` service in the Helm CLI mode

mantoine96:
In a previous PR (https://github.com/cilium/cilium-cli/pull/1938), support was added for passing annotations to the `clustermesh-apiserver` Service definition using the classic CLI.

However the current mode for the CLI, which relies on Helm, lacks this functionality. This PR therefore introduces this feature by relying on the `clustermesh.apiserver.service.annotations` Helm value and setting it.

I have made sure that passing annotations will also work with the automatic detection of Kubernetes flavor + allow overriding the preset annotations for these flavors.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -93,6 +93,7 @@ type K8sClusterMesh struct {
 
 type Parameters struct {
 	Namespace            string
+	ServiceAnnotations   []string
 	ServiceType          string
 	DestinationContext   string
 	Wait                 bool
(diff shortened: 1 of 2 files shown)


clustermesh/clustermesh.go
```diff
@@ -1425,48 +1426,75 @@ func log(format string, a ...interface{}) {
...
+		for k, v := range targetAnnotationsMap {
+			sourceAnnotationsMap[k] = v
+		}
+	}
+}
+
 func generateEnableHelmValues(params Parameters, flavor k8s.Flavor) (map[string]interface{}, error) {
 	helmVals := map[string]interface{}{
 		"clustermesh": map[string]interface{}{
+			"clusters":     []string{},
```
thorn3r, line 1449: Did you intend to include this line? There's a `clustermesh.config.clusters` value, but doesn't seem related to this change.

thorn3r (changes requested):
Hey @mantoine96 thanks for the PR! I left 1 comment, otherwise looks good

clustermesh/clustermesh.go
mantoine96: Good catch! My bad, this wasn't meant to be here at all. 🙈

mantoine96:
Thanks for your review @thorn3r. The unintended change has been removed.

asauber (changes requested):
For this type of customization, our recommendation is to use layered helm values. For example:

```
cilium install \   
    --helm-values=${HOME}/go/src/github.com/cilium/cilium/contrib/testing/kind-common.yaml \   
    --helm-values=${HOME}/src/mystack/cilium-values.yaml \ 
    --helm-values=${HOME}/src/mylayeredstack/cilium-values.yaml
```

The contribution is appreciated, but I am inclined to close this PR.

One of the reasons for implementing helm mode was to remove the many flags that had been built up for various customizations available via the Helm chart.

mantoine96:
Hey! Thanks for the feedback. Totally understand it, although I feel annotations are harmless enough, but I can see the bigger picture. 

I'll close this PR then!

Issue #2623, opened 2024-06-23
CI: run cilium-cli inside docker instead of using cilium-cli-test-job-chart helm chart


michi-covalent:
a minor quality of life improvement to make it easier to use cilium-cli in github workflows.

PR #2641, opened 2024-06-26
Remove cilium-cli-test-job-chart

michi-covalent:
Now all the cloud workflows run cilium-cli inside a container instead of using cilium-cli-test-job-chart.

Also update eks-tunnel.yaml and remove the helm install step. I forgot to remove it in 5146f46d.

Fixes: #2623


viktor-kurchenko (approved):
Like!

PR #2648, opened 2024-06-27
Simplify Kind workflow

michi-covalent:
Simplify the Kind workflow now that we use --test-concurrency flag in all the other workflows.

- Run "Installation and Connectivity Test job" with --test-concurrency=5. This takes around 20 minutes.
- Run "Helm Upgrade Clustermesh" job without --test-concurrency flag so that we run cilium connectivity test without --test-concurrency flag at least once. This takes around 15 minutes.

Also remove --debug flag from all the invocations of cilium connectivity test. It's a bit too noisy, and it makes it difficult to find relevant logs when tests fail.


viktor-kurchenko (approved):
Let it be!

PR #2653, opened 2024-07-01
Allow changing Kube Service type for connectivity tests

antonipp:
# Description
The goal of this PR is to allow users to change the type of Services created by automated connectivity tests. 

The default type is hardcoded to `NodePort` which is not necessarily the best choice for everyone. For example, we are currently setting up a [GateKeeper rule](https://open-policy-agent.github.io/gatekeeper-library/website/validation/block-nodeport-services) to block `NodePort` service creation in all of our clusters, mainly for security, performance and practicality reasons (for more context, see for instance https://oteemo.com/think-nodeport-kubernetes/).

The newly added parameter looks like this (and still has `NodePort` as the default):
```
      --service-type string             Type of Kubernetes Services created for connectivity tests (default "NodePort")
```

# Testing
- [x] Ran the tests without setting `--service-type`, verified that `NodePort` services were still being created
- [x] Ran the tests with `--service-type=ClusterIP`, verified that `ClusterIP` services were properly created:
```
$ kubectl -n cilium-test get svc echo-other-node
NAME              TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
echo-other-node   ClusterIP   10.x.x.x        <none>        8080/TCP   4m23s
```
- [x] Ran the tests with an invalid value `--service-type=FAIL`, verified that the tests failed with a clear error:
```
connectivity test failed: Service "echo-same-node" is invalid: spec.type: Unsupported value: "FAIL": supported values: "ClusterIP", "ExternalName", "LoadBalancer", "NodePort"
```

### DIFF
--- a/connectivity/check/deployment.go
+++ b/connectivity/check/deployment.go
@@ -303,7 +303,7 @@ var serviceLabels = map[string]string{
 	"kind": kindEchoName,
 }
 
-func newService(name string, selector map[string]string, labels map[string]string, portName string, port int) *corev1.Service {
+func newService(name string, selector map[string]string, labels map[string]string, portName string, port int, serviceType string) *corev1.Service {
 	ipFamPol := corev1.IPFamilyPolicyPreferDualStack
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
@@ -311,7 +311,7 @@ func newService(name string, selector map[string]string, labels map[string]strin
 			Labels: labels,
 		},
 		Spec: corev1.ServiceSpec{
-			Type: corev1.ServiceTypeNodePort,
+			Type: corev1.ServiceType(serviceType),
 			Ports: []corev1.ServicePort{
 				{Name: portName, Port: int32(port)},
 			},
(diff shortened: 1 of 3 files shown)


connectivity/check/deployment.go
```diff
@@ -911,9 +911,8 @@ func (ct *ConnectivityTest) deploy(ctx context.Context) error {
 
 				svc := newService(echoExternalNodeDeploymentName,
 					map[string]string{"name": echoExternalNodeDeploymentName, "kind": kindEchoExternalNodeName},
-					map[string]string{"kind": kindEchoExternalNodeName}, "http", 8080)
+					map[string]string{"kind": kindEchoExternalNodeName}, "http", 8080, "ClusterIP")
 				svc.Spec.ClusterIP = corev1.ClusterIPNone
-				svc.Spec.Type = corev1.ServiceTypeClusterIP
```
asauber, line 916: With this PR, we lose the type information from corev1, but I think that's reasonable in this case since these strings are so well-known and would fail to provision if misconfigured.

asauber (approved):

nebril (approved):

antonipp:
Thank you for the reviews @asauber @nebril 
Do you know how I can merge the PR?

asauber:
Hi @antonipp, apologies for the shuffling here.

This code has been moved into https://github.com/cilium/cilium/tree/main/cilium-cli.

Would you mind re-opening this change against that tree?

antonipp:
Sure, I re-opened it: https://github.com/cilium/cilium/pull/34498
Closing this one.

PR #2657, opened 2024-07-04
connectivity: add local-redirect-policy-with-node-dns test

ysksuzuki:
This PR introduces local-redirect-policy-with-node-dns connectivity test that checks if the LRP with node-local-dns setup is properly working.
https://docs.cilium.io/en/stable/network/kubernetes/local-redirect-policy/#node-local-dns-cache

ref: https://github.com/cilium/cilium/issues/33415

### DIFF
--- a/connectivity/deploy/manifests/node-local-dns-ds.yaml
+++ b/connectivity/deploy/manifests/node-local-dns-ds.yaml
@@ -1,0 +1,82 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: node-local-dns
+  namespace: kube-system
+  labels:
+    k8s-app: node-local-dns
+spec:
+  updateStrategy:
+    rollingUpdate:
+      maxUnavailable: 10%
+  selector:
+    matchLabels:
+      k8s-app: node-local-dns
+  template:
+    metadata:
+      labels:
+        k8s-app: node-local-dns
+      annotations:
+        policy.cilium.io/no-track-port: "53"
+        prometheus.io/port: "9253"
+        prometheus.io/scrape: "true"
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: cilium.io/no-schedule
+                operator: NotIn
+                values:
+                - "true"
+      priorityClassName: system-node-critical
+      serviceAccountName: node-local-dns
+      dnsPolicy: Default  # Don't use cluster DNS.
+      tolerations:
+      - key: "CriticalAddonsOnly"
+        operator: "Exists"
+      - effect: "NoExecute"
+        operator: "Exists"
+      - effect: "NoSchedule"
+        operator: "Exists"
+      containers:
+      - name: node-cache
+        image: registry.k8s.io/dns/k8s-dns-node-cache:1.15.16
+        resources:
+          requests:
+            cpu: 25m
+            memory: 5Mi
+        args: [ "-localip", "169.254.20.10,{{.PILLAR_DNS_SERVER}}", "-conf", "/etc/Corefile", "-upstreamsvc", "kube-dns-upstream", "-skipteardown=true", "-setupinterface=false", "-setupiptables=false" ]
+        ports:
+        - containerPort: 53
+          name: dns
+          protocol: UDP
+        - containerPort: 53
+          name: dns-tcp
+          protocol: TCP
+        - containerPort: 9253
+          name: metrics
+          protocol: TCP
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8080
+          initialDelaySeconds: 60
+          timeoutSeconds: 5
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/coredns
+        - name: kube-dns-config
+          mountPath: /etc/kube-dns
+      volumes:
+      - name: kube-dns-config
+        configMap:
+          name: kube-dns
+          optional: true
+      - name: config-volume
+        configMap:
+          name: node-local-dns
+          items:
+            - key: Corefile
+              path: Corefile.base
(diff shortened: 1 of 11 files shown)


connectivity/deploy/manifests/node-local-dns-ds.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
joestringer, line 2: In which cases will this test be run?

The background of this question is that users can run the CLI connectivity tests in production environments, and I wouldn't want to deploy another DNS DaemonSet in a live environment by default as this could potentially cause at least confusion, if not active disruption to the cluster. Even if this doesn't cause disruption, deploying a DaemonSet to thousands of nodes in a production environment could take significant resources.

This might be a case where we should move the node-local DNS setup pieces into a dedicated GitHub workflow so that this test more closely matches the way that our users will configure and test this feature.

joestringer: (I see code that should make this test only run when LRP is enabled, but still, deploying a node-local DNS DaemonSet as part of the test is breaking the assumption "treat the cluster like a live production environment".)

joestringer (changes requested):
I suggest to move the node-local DNS deployment into a GitHub workflow instead.

aditighag (commented):
Thanks for adding the the tests. (Full review is in progress.) 

Somewhat related to Joe's comment: We should run this as a separate test (and not as a part of the concurrent test suite). The reason being, node-local dns cache takes over DNS resolution once deployed with LRP.

ysksuzuki:
> We should run this as a separate test (and not as a part of the concurrent test suite). The reason being, node-local dns cache takes over DNS resolution once deployed with LRP.

Yes, it's the reason why I added `local-redirect-policy-with-node-dns` in `sequentialTests`

>  deploying a DaemonSet to thousands of nodes in a production environment could take significant resources.

Given that users can run the connectivity test in production as Joe mentioned, I agree that the node local DNS should not be deployed as part of the connectivity test. 

How about the following approach? WDYT?

- add an option  `contrib/scripts/kind.sh` to setup a cluster with the node-local DNS with LRP
- add node-local DNS config in the conformance e2e matrix. (set up with the node-local DNS with LRP if `node-local-dns: true `)
- run `local-redirect-policy-with-node-dns` if the cluster has the node-local DNS

aditighag:
> > We should run this as a separate test (and not as a part of the concurrent test suite). The reason being, node-local dns cache takes over DNS resolution once deployed with LRP.
> 
> Yes, it's the reason why I added `local-redirect-policy-with-node-dns` in `sequentialTests`

Ah, sorry I missed that. 🚀 

> Given that users can run the connectivity test in production as Joe mentioned, I agree that the node local DNS should not be deployed as part of the connectivity test.
> 

How about we gate this test behind a flag (e.g., `run-node-local-dns-test`) similar to other perf/disruptive etc tests? IMO, it's useful for users to be able to run the test as part of the connectivity test suite before they deploy node-local DNS using LRP.

joestringer:
@ysksuzuki the `kind.sh` proposal is OK for cilium/cilium, but this repository doesn't have those scripts. I think the test should also be run and pass from this repository? cc @michi-covalent

ysksuzuki:
> How about we gate this test behind a flag (e.g., run-node-local-dns-test) similar to other perf/disruptive etc tests? IMO, it's useful for users to be able to run the test as part of the connectivity test suite before they deploy node-local DNS using LRP.

I think it could be another option, and agree that it's useful for users who planning to deploy that feature. 
@joestringer WDYT?

michi-covalent:
we'll merge cilium-cli into cilium repo soon-ish, so for now maybe you can just curl https://raw.githubusercontent.com/cilium/cilium/main/contrib/scripts/kind.sh if you need to use the script.

put the test behind --include-unsafe-tests flag https://github.com/cilium/cilium-cli/blob/50aa7f1610d650aa83e7ba14ac1c47867992a2c5/cli/connectivity.go#L138, and also add a feature detection logic so that this test only runs when node-local-dns is present. 

then move any setup logic that mutates the cluster state outside of cilium-test namespace into the workflow file like joe suggested. something similar to https://github.com/cilium/cilium-cli/blob/50aa7f1610d650aa83e7ba14ac1c47867992a2c5/.github/workflows/kind.yaml#L85-L99

joestringer:
I think it's reasonable to gate the test behind a flag if we are worried about its impact on production environments.

I don't think it's the responsibility of cilium-cli to deploy node-local DNS; cilium-cli should just run a test in a target environment. If the target environment has node-local DNS, then it should ideally detect that and run tests that will validate that LRP is working correctly in that environment. If the target environment doesn't have node-local DNS, then I think it makes sense to either do the test another way, or to skip the test.

joestringer:
> maybe you can just curl https://raw.githubusercontent.com/cilium/cilium/main/contrib/scripts/kind.sh

Generally it's discouraged to point directly to a moving target like `main`, but this would be expedient for the short term. We can always pin the version after the feature lands in kind.sh in cilium/cilium.

ysksuzuki:
@michi-covalent Thank you for your comment.

> then move any setup logic that mutates the cluster state outside of cilium-test namespace into the workflow file like joe suggested. 

Do I need to push my branch on `cilium/cilium-cli` not a fork to modify the workflow file?

ysksuzuki:
> I don't think it's the responsibility of cilium-cli to deploy node-local DNS; cilium-cli should just run a test in a target environment.

Make sense. Changing the cluster DNS setup as part of the test would be too invasive.

michi-covalent:
> Do I need to push my branch on cilium/cilium-cli not a fork to modify the workflow file?

yeah to test workflow file changes you need to push a branch to cilium/cilium-cli repo and open a pull request from there 🚀🙏

ysksuzuki:
close in favor of #2669

PR #2660, opened 2024-07-04
kvstoremesh: don't disable by default

marseel:
In 1.16 we enabled kvstoremesh by default in helm. Currently, cilium-cli was still setting kvstoremesh.enabled to false. Now, if user does not explicitly specify if kvstoremesh should be enabled or disabled, we will rely on default helm value.


giorio94 (approved):
Thanks!

squeed (approved):

PR #2664, opened 2024-07-05
clustermesh: validate service type, and fix deprecated LB annotations

giorio94:
Currently, the "clustermesh enable" command lists ClusterIP as one of the supported service types. However, exposing the clustermesh-apiserver via a ClusterIP service is hardly ever a sensible idea, as this component must be reachable from other clusters. Still, this possibility has proved to be sometimes confusing for newbie users, who ended up with a broken setup.

In an effort to prevent these problems, let's explicitly validate the specified service type, and allow only NodePort and LoadBalancer services. In any case, it is still possible to configure a service of type ClusterIP via the helm chart in the handful of very advanced use-cases (if any) which can actually work and benefit from it.

While being there, let's also update the deprecated annotations used to configure internal LoadBalancers.


mhofstetter (approved):
`cli` :heavy_check_mark:

giorio94:
@jrajahalme Gentle ping :pray:

jrajahalme (approved):

PR #2665, opened 2024-07-05
clustermesh: fix remote clusters configuration reset upon deactivation

giorio94:
The blamed commit explicitly reset the list of remote clusters upon clustermesh deactivation. However, it did so setting the corresponding setting to nil, which in Cilium v1.16 triggers an error due to not matching the schema configured. Let's address this by configuring an empty array instead.

Fixes: b148800cf7a2 ("clustermesh: reset remote clusters configuration upon disconnection")


giorio94:
@jrajahalme Gentle ping :pray:

jrajahalme (approved):

PR #2670, opened 2024-07-10
connectivity: Add echo-ingress-l7-via-hostport-with-encryption testcase

jschwinger233:
Add testcase to cover https://github.com/cilium/cilium/issues/32899.

The new testcase has been tested by https://github.com/cilium/cilium/pull/33691, two failing checks are ci-e2e-upgrade and mergeability.

### DIFF
--- a/connectivity/builder/echo_ingress_l7.go
+++ b/connectivity/builder/echo_ingress_l7.go
@@ -40,4 +40,20 @@ func (t echoIngressL7) build(ct *check.ConnectivityTest, _ map[string]string) {
 		WithCiliumPolicy(echoIngressL7HTTPPolicyYAML). // L7 allow policy with HTTP introspection
 		WithScenarios(tests.PodToPodWithEndpoints()).
 		WithExpectations(expectation)
+
+	newTest("echo-ingress-l7-via-hostport-with-encryption", ct).
+		WithCondition(func() bool { return !ct.Params().SingleNode }).
+		WithFeatureRequirements(
+			features.RequireEnabled(features.L7Proxy),
+			// Once https://github.com/cilium/cilium/issues/33168 is fixed, we
+			// can enable for IPsec too.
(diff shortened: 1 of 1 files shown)


mhofstetter (approved):
Thanks!

connectivity/builder/echo_ingress_l7.go
```diff
@@ -16,26 +16,44 @@ var echoIngressL7HTTPPolicyYAML string
...
+		WithFeatureRequirements(
+			features.RequireEnabled(features.L7Proxy),
+			// Once https://github.com/cilium/cilium/issues/33168 is fixed, we
+			// can enable for IPsec too.
+			features.RequireMode(features.EncryptionPod, "wireguard"),
+			// Otherwise pod->hostport traffic will be policy
+			// denied on the ingress of dest node when
+			// routing=vxlan + kpr=1 + bpf_masq=1
+			features.RequireEnabled(features.EncryptionNode),
+		).
```
julianwiedmann, line 55: Thank you! Looking forward - I think we should also run this test without encryption enabled :). And hopefully find a way to remove the requirement on NodeEncryption ...

julianwiedmann (approved):

connectivity/builder/echo_ingress_l7.go
julianwiedmann: @jschwinger233 ah but one thing we *do* need (and this is arguably a CLI release blocker) - this test should only run with cilium 1.16.0 or newer.

jschwinger233: Yeah, on it.

julianwiedmann: > Thank you! Looking forward - I think we should also run this test without encryption enabled :). And hopefully find a way to remove the requirement on NodeEncryption ...

Moved this to https://github.com/cilium/cilium-cli/issues/2758.

PR #2679, opened 2024-07-12
connectivity: disrupt deployment improved

viktor-kurchenko:
Deploy conn-disrupt test actors only in the first test namespace in case of tests concurrent run to avoid resource wasting. Conn-disrupt tests always run at the beginning, sequentially and in the first test namespace.

### DIFF
--- a/.github/workflows/kind.yaml
+++ b/.github/workflows/kind.yaml
@@ -105,8 +105,7 @@ jobs:
           # disruption), but we want to make sure that the command works as expected.
           #
           # Dispatch interval is set to 100ms, b/c otherwise (default is 0), the flow validation might time out.
-          cilium connectivity test --test-namespace test-namespace \
-            --test-concurrency=5 \
+          cilium connectivity test --test-namespace test-namespace-1 \
             --conn-disrupt-dispatch-interval 100ms \
             --include-conn-disrupt-test --conn-disrupt-test-setup
 
--- a/cli/connectivity.go
+++ b/cli/connectivity.go
@@ -243,6 +243,7 @@ func newConnectivityTests(params check.Parameters, logger *check.ConcurrentLogge
 	for i := 0; i < params.TestConcurrency; i++ {
 		params := params
 		params.TestNamespace = fmt.Sprintf("%s-%d", params.TestNamespace, i+1)
+		params.TestNamespaceIndex = i
 		if params.ExternalTargetCANamespace == "" {
 			params.ExternalTargetCANamespace = params.TestNamespace
 		}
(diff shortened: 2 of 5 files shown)


michi-covalent (commented):
let's add a test for this in one of the workflows. https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml is already using --test-concurrency flag, so maybe do this conn disrupt thingy there.

edit: btw kind workflow doesn't seem too happy 👀 https://github.com/cilium/cilium-cli/actions/runs/9911162365/job/27383203458?pr=2679

viktor-kurchenko:
> let's add a test for this in one of the workflows. https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml is already using --test-concurrency flag, so maybe do this conn disrupt thingy there.
> 
> edit: btw kind workflow doesn't seem too happy 👀 https://github.com/cilium/cilium-cli/actions/runs/9911162365/job/27383203458?pr=2679

I've fixed the error.
Also, kind workflow already contains disrupt tests: [here](https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml#L111) and [here](https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml#L116).

I think we can do the same for performance tests.

michi-covalent (commented):
i guess we still need to specify --test-concurrency flag here? 💭 

https://github.com/cilium/cilium-cli/blob/6d90b8c7146eaabf5a09f1f5a541f2f53dbf28f9/.github/workflows/kind.yaml#L109

viktor-kurchenko:
> i guess we still need to specify --test-concurrency flag here? 💭
> 
> https://github.com/cilium/cilium-cli/blob/6d90b8c7146eaabf5a09f1f5a541f2f53dbf28f9/.github/workflows/kind.yaml#L109

I've removed `--test-concurrency=5` from the first run conn tests run.

Also, please note that `--test-namespace test-namespace` is renamed to `--test-namespace test-namespace-1` otherwise we need to start disrupt tests again for the second run.

michi-covalent (approved):
for future improvement, maybe it's better to create a namespace `cilium-test-1` instead of `cilium-test` when test-concurrency is set to 1 to get rid of the special case where there is no `-$index` suffix in the namespace name.

aanm (approved):

christarazi (approved):

michi-covalent:
ok i'm merging this. i tried this cilium-cli against the cilium repo, and there are some failures https://github.com/cilium/cilium/actions/runs/9914828717 let's investigate before enabling --test-concurrency flag in cilium repo 🧑‍🔬

PR #2684, opened 2024-07-15
flags: Use option.NewNamedMapOptions

michi-covalent:
Use option.NewNamedMapOptions instead of flags.StringToStringVar to avoid getting lint errors when merging cilium-cli repo to cilium repo [^1].

[^1]: https://github.com/cilium/cilium/blob/main/contrib/scripts/check-viper.sh

### DIFF
--- a/cli/connectivity.go
+++ b/cli/connectivity.go
@@ -13,6 +13,7 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


derailed (approved):
@michi-covalent Thank you for this update!

PR #2686, opened 2024-07-15
connectivity: echo-ingress-l7-via-hostport-with-encryption only runs with cilium 1.16.0 or newer

jschwinger233:
…with cilium 1.16.0 or newer.

Fixes: https://github.com/cilium/cilium-cli/pull/2670


sayboras (approved):
LGTM :heavy_check_mark:

PR #2695, opened 2024-07-18
Improve network perf test

marseel:
This PR contains various improvements to net perf testing - more details in commit messages.

Most notable improvement is that we no longer use fixed message size for TCP_STREAM test, which results in way more reasonable results as compared to the previous tests. 
First of all, we can finally saturate network throughput and also throughput between pods on the same node is way higher than throughput between pods running on different nodes. Example test results for native routing:
```
-------------------------------------------------------------------------------------
📋 Scenario        | Node       | Test            | Duration        | Throughput Mb/s 
-------------------------------------------------------------------------------------
📋 pod-to-pod      | same-node  | TCP_STREAM      | 10s             | 14257.07     
📋 host-to-host    | same-node  | TCP_STREAM      | 10s             | 25766.59     
📋 pod-to-pod      | other-node | TCP_STREAM      | 10s             | 9541.04      
📋 host-to-host    | other-node | TCP_STREAM      | 10s             | 9579.99      
-------------------------------------------------------------------------------------
```
Previously native throughput was showing sad 1500 Mb/s for the same Cilium configuration. 

How the new output looks like in perfdash:
image
image
Makes it much easier to compare host vs pod throughput / latency etc.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,6 +8,7 @@
 # @cilium/sig-encryption         Encryption management
 # @cilium/sig-bgp                BGP integration
 # @cilium/sig-clustermesh        Clustermesh and external workloads
+# @cilium/sig-scalability        Scalability & performance of Cilium
 # @cilium/sig-hubble             Hubble integration
 # @cilium/sig-k8s                K8s integration, K8s CNI plugin
 # @cilium/vendor                 Vendoring, dependency management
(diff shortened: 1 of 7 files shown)


christarazi (approved):

CODEOWNERS
```diff
@@ -59,7 +61,6 @@
 /connectivity/tests/health.go @cilium/sig-agent
 /connectivity/tests/host.go @cilium/sig-agent
 /connectivity/tests/ipsec_xfrm.go @cilium/ipsec
-/connectivity/tests/perfpod.go @cilium/sig-datapath
```
qmonnet, line 62: This is because the file has moved to `connectivity/perf/benchmarks/netperf/`, right?

qmonnet (approved):
Looks good to me, thanks!

I spotted a typo (`explicilty`) in the description in the third commit, if you need to update your PR for any other reason [EDIT: For example, to address a merge conflict :slightly_smiling_face:].

CODEOWNERS
marseel: yes, I've moved that long time ago, just didn't update CODEOWNERS at that time: https://github.com/cilium/cilium-cli/pull/2168
I'm trying to decouple perf command from connectivity test stuff, but making small steps towards that :)

pchaigno (approved):

marseel:
"Lock cilium-cli repo / Lock cilium-cli repo (pull_request)"
:sob:

PR #2698, opened 2024-07-18
Update CONTRIBUTING.md 

michi-covalent:
Document that we are planning to merge Cilium CLI code into cilium/cilium repository.

Ref: https://github.com/cilium/design-cfps/pull/9

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,13 @@
 # Contributing to Cilium CLI
 
+## IMPORTANT Cilium CLI is moving 🚚📦📦📦
+
+We are planning to merge Cilium CLI code into cilium/cilium repository to
+simplify the overall development process after Cilium v1.16.0 gets released.
+See [CFP-25694](https://github.com/cilium/design-cfps/pull/9) for details.
+If you have questions, post a message in
+[#development Cilium Slack channel](https://cilium.slack.com/archives/C2B917YHE).
+
 ## Contribution workflow
 
 Cilium CLI uses GitHub for collaborative development. Please use GitHub issues


michi-covalent:
cc @joestringer @aanm (i can't seem to @ cilium-maintainers team)

joestringer:
@cilium/cilium-maintainers

README.md
joestringer, line 1: Is the intent to change the user experience for this too? For instance, expect users to download from cilium/cilium, release on a different cadence?

michi-covalent:
> Is the intent to change the user experience for this too? For instance, expect users to download from cilium/cilium, release on a different cadence?

the current proposal is to *not* change the user experience, at least in the short term. we'll continue to publish release artifacts in cilium/cilium-cli repo at its own release cadence independent of cilium/cilium release cycle:

https://github.com/cilium/design-cfps/blame/3b5b719b9fac007d2ea683b18aafe4b3cb62e1eb/cilium/CFP-25694-move-cilium-cli-to-cilium-repo.md#L77-L80

joestringer:
OK, the reason I ask is that this is placing the warning right at the top as the first thing that users see. This looks like a warning for users to get ready to change the way they pull the tool.

Developer-focused warnings should probably be lower down in the readme in a section that's clearly about how you contribute.

michi-covalent:
yeah makes sense. maybe i can update https://github.com/cilium/cilium-cli/blob/main/CONTRIBUTING.md instead 💭

pchaigno:
Is it expected that cilium/cli is also an owner for the README? That doesn't make much sense for me.

michi-covalent:
> Is it expected that cilium/cli is also an owner for the README? That doesn't make much sense for me.

cilium/cli team is the default owner of this repo, so yeah i guess it is sort of expected https://github.com/cilium/cilium-cli/blob/main/CODEOWNERS#L17 do you have any particular team on your mind who should own README.md?

pchaigno:
cilium-cli-maintainers? 😁 Or we split cli into cilium-cli and cilium's CLIs.

pchaigno:
Yeah, so we at least can agree that the description doesn't match the current role of fallback for all of cilium-cli.

joestringer:
The reviewer group for CLI is currently documented as having a project-wide scope: https://docs.cilium.io/en/stable/contributing/development/reviewers_committers/review_process/#project-wide . We can of course revisit this and figure out if there is a better set of individuals who could take ownership of different components.

pchaigno (approved):

michi-covalent:
https://github.com/cilium/cilium-cli/pull/2710 should reduce the noise to the cli team during the cilium-cli release process 🚀 🙏

joestringer (approved):

PR #2702, opened 2024-07-19
clustermesh: parallelize connections

littlejo:
The aim of the PR is to parallelize helm upgrade during the command "cilium clustermesh connect".

In localhost with Kind, it takes about 15 seconds. With my patch, it takes about 8 seconds.
In remote, it can take about 60 seconds. With my patch, it takes about 30 seconds.

### DIFF
--- a/clustermesh/clustermesh.go
+++ b/clustermesh/clustermesh.go
@@ -19,6 +19,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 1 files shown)


thorn3r (commented):
LGTM! Do you mind adding a bit more context about the change to the commit message?

littlejo:
@thorn3r 
> LGTM! Do you mind adding a bit more context about the change to the commit message?

Is it better?

thorn3r:
> @thorn3r
> 
> > LGTM! Do you mind adding a bit more context about the change to the commit message?
> 
> Is it better?

I think for the future it would be good to include not just the `what` but the `why`, maybe adding some of the details of your testing and the reason for the change. Just a nit, non-blocking

thorn3r (approved):

littlejo:
> I think for the future it would be good to include not just the what but the why, maybe adding some of the details of your testing and the reason for the change. Just a nit, non-blocking

The reason of my change is:
https://medium.com/@littel.jo/cr%C3%A9er-511-clusters-kubernetes-interconnect%C3%A9s-avec-cilium-cluster-mesh-partie-2-03da66a9271d (FR)

If you want to create 511 clusters with cluster mesh at the same time with all connections, save half of time of connections is interesting 🐎

littlejo:
@thorn3r Do you know when it is merged ?

littlejo:
Hello,

afterthought, i think it's not the best idea. `helm upgrade` took a lot of CPU. If you parallelize, you just need more CPUS. So if you parallelize this command a lot of time, you need a super powerful computer unless you don't save time.

PR #2703, opened 2024-07-19
sysdump: fix file separator

viktor-kurchenko:
The commit fixes sysdump collection for Windows based clients.

### DIFF
--- a/sysdump/sysdump.go
+++ b/sysdump/sysdump.go
@@ -2098,7 +2098,8 @@ func (c *Collector) SubmitTetragonBugtoolTasks(pods []*corev1.Pod, tetragonAgent
 // (e.g. "a/b/c" => "b/c"). Don't pass an absolute path. It doesn't do what
 // you think it does.
 func removeTopDirectory(path string) (string, error) {
-	index := strings.IndexByte(path, filepath.Separator)
+	// file separator hardcoded because sysdump always created on Linux OS
+	index := strings.IndexByte(path, '/')
 	if index < 0 {
 		return "", fmt.Errorf("invalid path %q", path)
 	}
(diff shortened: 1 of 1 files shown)


sysdump/sysdump.go
```diff
@@ -2098,7 +2098,11 @@ func (c *Collector) SubmitTetragonBugtoolTasks(pods []*corev1.Pod, tetragonAgent
 // (e.g. "a/b/c" => "b/c"). Don't pass an absolute path. It doesn't do what
 // you think it does.
 func removeTopDirectory(path string) (string, error) {
-	index := strings.IndexByte(path, filepath.Separator)
```
michi-covalent, line 2101: just curious why this doesn't work. `filepath.Separator` is already defined to be `\\` on windows right?

viktor-kurchenko: Yeah, for Windows users `filepath.Separator==\` but cilium dump is from Linux host and inside we have `/`.
So, we can hardcode `filepath.Separator` to `/` but what if in the future Cilium will run on Windows Server? )))

chancez: Yeah so `file path.Separator` is based on the client device's OS. So basically on windows it will be `\` but on Unixes it will be `/`. The bugtool archive will always contain `/` since it runs on Linux, but when we try to remove the separator on a windows host, it will look for `\` instead.

sysdump/sysdump.go
```diff
@@ -2098,7 +2098,11 @@ func (c *Collector) SubmitTetragonBugtoolTasks(pods []*corev1.Pod, tetragonAgent
...
+	index := strings.IndexByte(path, '/')
+	// let's try Windows file separator
+	if index < 0 {
+		index = strings.IndexByte(path, '\\')
+	}
```
chancez, line 2105: Technically I don't think we need to actually try the windows separator since the sysdump archives will be created on Linux only.

viktor-kurchenko: Ok, let's simplify it. Removed.

michi-covalent (approved):

chancez (approved):

viktor-kurchenko:
@pchaigno kind ping)

PR #2704, opened 2024-07-19
connectivity: Assert BGP timer intervals in BGP connectivity tests

rastislavs:
This change asserts that BGP peering connections used in the BGP connectivity tests use the BGP timer intervals as configured via Cilium.

### DIFF
--- a/connectivity/check/frr.go
+++ b/connectivity/check/frr.go
@@ -54,12 +54,16 @@ type FRRBGPPeeringParams struct {
 
 // FRRBGPNeighborInfo holds FRR BGP neighbor information equivalent to "show bgp neighbor json" CLI output entry.
 type FRRBGPNeighborInfo struct {
-	RemoteAS       int    `json:"remoteAs"`
-	LocalAS        int    `json:"localAs"`
-	Hostname       string `json:"hostname"`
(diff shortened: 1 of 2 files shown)


connectivity/check/frr.go
```diff
@@ -250,41 +254,56 @@ func RenderFRRBGPPeeringConfig(t *Test, params FRRBGPPeeringParams) string {
...
-		return
+		return entries
+	}
+}
+
+// AssertFRRBGPNeighborTimers asserts that peering connections of the provided neighbors filtered by checkNeighbors
+// use the provided BGP timer intervals.
+func AssertFRRBGPNeighborTimers(t *Test, neighbors map[string]FRRBGPNeighborInfo, checkNeighbors []netip.Addr, keepAliveSeconds, holdTimeSeconds int32) {
+	for _, addr := range checkNeighbors {
+		neighbor := neighbors[addr.String()]
```
harsimran-pabla, line 300: We should check for existence in the neighbors map. 

This is probably not possible in the way the test is using AssertFRRBGPNeighborTimers.

harsimran-pabla (approved):
One minor check required, otherwise lgtm.

PR #2712, opened 2024-07-24
connectivity: Fix ineffective version check in BGP tests

rastislavs:
Multiple `WithCondition` statements were causing that only the last one was effective, making version checks ineffective. This combines the conditions into a single `WithCondition` statement per test.


harsimran-pabla (approved):

joestringer (approved):
Thanks for the fix.

(I wonder if it's worth putting some footgun check into `WithCondition()` to avoid this override, but we can follow up on that separately cc @cilium/cilium-cli-maintainers )

michi-covalent (approved):
filed https://github.com/cilium/cilium-cli/issues/2713

PR #2714, opened 2024-07-24
tests: lrp: add support for protocol differentiation

jibi:
update the logic used to wait for local redirect bpf entries as from
v1.16 we start supporting protocol differentiation for services, which
means the frontend string description will now include also the protocol


jibi:
this will need to be merged and released in coordination with https://github.com/cilium/cilium/pull/33434

christarazi (approved):

Issue #2720, opened 2024-07-25
pick up the latest kind version


michi-covalent:
set up renovate to keep the kind node image up to date:

- https://github.com/cilium/cilium-cli/blob/main/.github/kind-config.yaml
- https://github.com/cilium/cilium-cli/blob/main/.github/kind-config-1.yaml
- https://github.com/cilium/cilium-cli/blob/main/.github/kind-config-2.yaml

also figure out what's causing this error when updating the kind node image: https://github.com/cilium/cilium-cli/actions/runs/10085247401/job/27886689646?pr=2716
```
      message: 'container runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady
        message:Network plugin returns error: cni plugin not initialized'
      reason: KubeletNotReady 
```
i guess it doesn't like having nodes without any CNI installed 💭

PR #2725, opened 2024-07-25
kind: Pick up the latest kindest/node image

michi-covalent:
Also update kubeadm.k8s.io API version from v1beta2 to v1beta3.

Ref: #2720


michi-covalent:
will update renovate config as a follow up PR 🚀🙏

PR #2620, opened 2024-06-21
Add multicast subcommand

yushoyamaguchi:
## Change
This change adds cilium multicast sub command to manage multicast groups and subscribers throughout the cluster

Following commands are added
```
# show list of multicast group
- cilium multicast list group

# show list of subscribers
- cilium multicast list subscriber --all
- cilium multicast list subscriber --group-ip <group-ip>

# make specified multicast group and add CiliumInternalIP of all cilium nodes to the group in each node
- cilium multicast add --group-ip <group-ip>

# Remove CiliumInternalIP of all cilium nodes from specified multicast group and delete the group in each node
- cilium multicast del --group-ip <group-ip>
```
This is follow up of https://github.com/cilium/cilium-cli/issues/2574

## Display example
cli-result

## Additional
As for testing, we will proceed with the implementation of the connectivity test for multicast as described in this issue.(https://github.com/cilium/cilium-cli/issues/2615)

## Reference
Current way of setup multicast : https://github.com/cilium/cilium/blob/main/Documentation/network/multicast.rst

PR #2726, opened 2024-07-26
CONTRIBUTING.md: Add guidelines regarding cluster state mutation

michi-covalent:


michi-covalent:
i'd like to get some initial feedback before marking it ready for review from folks who have been dealing connectivity test recently, cc @rastislavs @ysksuzuki @yushoyamaguchi @brb @joestringer

CONTRIBUTING.md
```diff
@@ -15,11 +15,36 @@ to discuss proposals and use pull requests to suggest changes. For more
...
+## An important note about mutating cluster resources
+
+Cilium CLI must follow these general rules regarding mutating cluster resources:
+
+- Cilium CLI must not mutate resources that are not managed by the Cilium Helm
+  chart. Cilium CLI must use Cilium Helm values to configure Cilium installations.
+- The only exception to the rule above is the connectivity test command, which
+  may mutate resources within Kubernates namespaces with the name prefix specified
+  by --test-namespace flag. The connectivity test must not mutate resources that
+  affect network connectivity outside the test namespaces.
```
joestringer, line 27: I agree with these as a baseline. I think that these two principles can guard user clusters against dangerous changes. This is important for user trust - we can ask users to just run the CLI tests and the tool will provide safe feedback about the state of the cluster.

These guidelines can help with us thinking about how we build test functionality into the CLI in ways that are useful for both developers and users, rather than just solving problems that developers have during the dev cycle. Ideally we can get this CLI to tell us if the environment is set up correctly to properly handle a particular feature without mutating the cluster in any significant way. Smoke test of sorts.

I also recognize that these are quite restrictive and given that we already have the `--include-unsafe-tests` escape hatch for integration testing, we need some other mechanism to solve those use cases for developers.

CONTRIBUTING.md
```diff
@@ -15,11 +15,36 @@ to discuss proposals and use pull requests to suggest changes. For more
...
+- TODO: I'd like to deprecate the use of --include-unsafe-tests flag for adding
+  tests that mutate resources outside the test namespaces:
+  - These unsafe tests typically can't run in parallel.
+  - It sets a bad precedent for new contributors. They might see test cases in
+    the connectivity test command that mutate ressources outside the test
+    namespaces, and assume that the connectivity test is the right command to
+    add more unsafe tests. It only takes one careless review for these tests
+    that are not guarded by --include-unsafe-tests to slip in.
```
joestringer, line 35: What would be the threshold for how we structure the code that would make you feel more comfortable with protecting user clusters against mutable state?

I could imagine us building into the common libraries for `cilium connectivity test` that the k8s client returns an error if it tries to mutate anything outside the set of resources explicitly declared in the point above this one. We could do this with a simple wrapper around the current k8s client. Then it would be structurally difficult for someone to introduce unsafe logic into the connectivity tests unless they went and modified that common library as well. This would signal the danger more easily I think. We could even restrict the codeowners on that common library to ensure review is more tightly watched, while relaxing codeowners for general test improvements.

Then let's say there's another command like your proposed `cilium test`, that command would get unrestricted access to the k8s client. So it could do whatever it needs to.

CONTRIBUTING.md
```diff
@@ -15,11 +15,36 @@ to discuss proposals and use pull requests to suggest changes. For more
 information see the [Cilium Development
 Guide](https://docs.cilium.io/en/latest/contributing/development/).
 
+## An important note about mutating cluster resources
+
+Cilium CLI must follow these general rules regarding mutating cluster resources:
+
+- Cilium CLI must not mutate resources that are not managed by the Cilium Helm
```
yushoyamaguchi, line 22: If we follow this policy, some of `cilium clustermesh` subcommand and `cilium multicast` subcommands, which I [just added](https://github.com/cilium/cilium-cli/pull/2620) is inappropriate?

michi-covalent: i consider `cilium multicast` to be one of the exceptional cases because there is no CRD to propagate the configuration to all the cilium agents.

i haven't looked at `cilium clustermesh` command for a while but at least clustermesh connection configuration is persisted as helm values: https://github.com/cilium/cilium-cli/blob/9577d6b14fc5ab28ffacc5e37e03df2086bc09a0/clustermesh/clustermesh.go#L1596

yushoyamaguchi: I understand.
Thank you so much for kind description.

CONTRIBUTING.md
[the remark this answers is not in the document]
rastislavs: I think this would make sense if it was just BGP, but we have more features/tests that require "unsafe" setup (e.g. EGW, LRP, policy tests that need a route from the `NodeWithoutCilium`). This means it would require multiple `cilium demo <feature> setup` + `cilium demo <feature> test` (and perhaps `cilium demo <feature> cleanup`) to run them all. 

But we could also have something like `cilium demo all setup` (+ `test` + `cleanup`) that would setup & run them all.

CONTRIBUTING.md
```diff
@@ -15,11 +15,36 @@ to discuss proposals and use pull requests to suggest changes. For more
...
+  chart. Cilium CLI must use Cilium Helm values to configure Cilium installations.
+- The only exception to the rule above is the connectivity test command, which
+  may mutate resources within Kubernates namespaces with the name prefix specified
```
rastislavs, line 25:
```suggestion
  may mutate resources within Kubernetes namespaces with the name prefix specified
```

CONTRIBUTING.md
rastislavs, line 27: Maybe it would be worth explicitly calling out that deploying any cluster-scoped CRD object as part of these tests is forbidden?

brb:
Do I read correctly that the main motivation for advising against mutation cluster state is that such tests cannot be parallelized? The safety aspect is covered by the  `--include-unsafe-tests`.

CONTRIBUTING.md
ysksuzuki: Can we give cilium-cli more restrictive permission when end-users run the connectivity test in their cluster. For example,  prohibit changes to kube-system namespace using k8s RBAC

joestringer: Interesting point. That's possible today, but it's up to the user to provide the restricted permissions to the CLI. The user could always provision and provide superadmin.

joestringer: ^ That said, let's say we provisioned the RBAC in the testsuite with those permissions... it could provide a safeguard, because if the CLI attempted to create resources beyond that scope, it would fail with an RBAC error. Would have to specifically run a testsuite with such RBAC and without the `--include-unsafe-tests` flag.

CONTRIBUTING.md
joestringer: I struggle a bit with the concept of an "unsafe test". I can understand unsafe provisioning then running the test in that environment. Provisioning is inherently mutating important state. I worry that as soon as you frame a _test_ as unsafe, it's fundamentally trying to modify things that could (a) impact other tests or (b) impact users.

CONTRIBUTING.md
yushoyamaguchi: > i consider `cilium multicast` to be one of the exceptional cases

Do you think this does not need to be written in CONTRIBUTING.md ?

PR #2736, opened 2024-07-31
connectivity: add public RenderTemplates function

ysksuzuki:
It provides a way to add external templates and use them with the default rendered templates for additional connectivity tests registered by hooks.

### DIFF
--- a/connectivity/builder/builder.go
+++ b/connectivity/builder/builder.go
@@ -268,6 +268,26 @@ func finalTests(ct *check.ConnectivityTest) error {
 	return injectTests([]testBuilder{checkLogErrors{}}, ct)
 }
 
+// RenderTemplates renders extra templates as well as default templates
+// and returns the rendered templates.
+// This function provide the rendered templates for additional connectivity
+// tests registered by the hooks.
+func RenderTemplates(param check.Parameters, extraTemplates map[string]string) (map[string]string, error) {
+	renderedTemplates, err := renderTemplates(param)
+	if err != nil {
+		return nil, err
+	}
+
+	for key, temp := range extraTemplates {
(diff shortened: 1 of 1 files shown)


connectivity/builder/builder.go
```diff
@@ -268,6 +268,26 @@ func finalTests(ct *check.ConnectivityTest) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for key, temp := range extraTemplates {
+		val, err := template.Render(temp, param)
+		if err != nil {
+			return nil, err
+		}
+		renderedTemplates[key] = val
```
michi-covalent, line 286: maybe add a unit test, and test the case the key collides with the templates map in `renderTemplates` function. might be a bit difficult to debug if we silently overwrite map entries here 💭

michi-covalent (commented):
let's call this function with an empty `extraTemplates` map instead of `renderTemplates` in https://github.com/cilium/cilium-cli/blob/main/connectivity/builder/builder.go#L313 to exercise this code path 🚀🙏 lgtm otherwise ✅

connectivity/builder/builder.go
michi-covalent: ... and also think about how to pass `extraTemplates` from the hooks. maybe SetupHooks needs to be extended somehow, or can we add `extraTemplates` from `SetupAndValidate` hook? 💭  https://github.com/cilium/cilium-cli/blob/e386af2b9f500e4c40436ac660cd6602da104fc7/connectivity/check/context.go#L268-L274

michi-covalent:
@ysksuzuki please rebase to pick up the fix for external workload / multicluster workflows. other than that this should be ready for review right?

PR #2759, opened 2024-08-15
connectivity: Run echo-ingress-l7-via-hostport-with-encryption for IPsec

jschwinger233:
Since https://github.com/cilium/cilium/issues/32897 has been fixed, it's time to cover it.


connectivity/builder/echo_ingress_l7.go
```diff
@@ -48,16 +48,27 @@ func (t echoIngressL7) build(ct *check.ConnectivityTest, _ map[string]string) {
 		WithCondition(func() bool {
 			return versioncheck.MustCompile(">=1.16.0")(ct.CiliumVersion)
```
julianwiedmann, line 49: We'll probably need to update this a bit then, to account for the fix backport.

Issue #2755, opened 2024-08-14
publish release images


michi-covalent:
publish cilium-cli release images. make it multi arch with arm64 and amd64. probably makes sense to create a new repo quay.io/cilium/cilium-cli instead of using https://quay.io/repository/cilium/cilium-cli-ci.

context: https://cilium.slack.com/archives/C1MATJ5U5/p1723450193483599

PR #2770, opened 2024-08-20
publish images: amd64/arm64 images containing cli only

leppeK:
build desired target with a buildargument
    
This way we are not diverging in multiple dockerfiles and we can choose between two targets:
- cilium-cli-ci The original target based on ubuntu and the default
- cilium-cli Only contains the cilium cli based on scratch
- 
Fixes: #2755


leppeK:
As there are changes to github ci i think these steps need to run from this branch to check if it builds correctly

Dockerfile-cli-only
```diff
@@ -0,0 +1,16 @@
+# syntax=docker/dockerfile:1.9@sha256:fe40cf4e92cd0c467be2cfc30657a680ae2398318afd50b0c80585784c604f28
+
+# Copyright Authors of Cilium
+# SPDX-License-Identifier: Apache-2.0
+
+FROM docker.io/library/golang:1.22.6-alpine3.19@sha256:1bad39361dd21f2f881ce10ff810e40e5be3eba89a0b61e762e05ec42f9bbaf2 AS builder
```
asauber, line 6: Can we avoid allowing this to diverge over time from the CI builder?

leppeK: I think we can by building the CI image of this one.

So to make it clear, Dockerfile should build only cilium-cli and we should have another Dockerfile building cilium-cli-ci of that one so no diverging happens

leppeK: https://github.com/cilium/cilium-cli/pull/2770/commits/bddd9844b646277013e4b67f828e996209d4f30f

I think this resolves everything with a single dockerfile

.github/workflows/images.yaml
```diff
@@ -25,8 +25,12 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: cilium-cli
+          - name: cilium-cli-ci
             dockerfile: ./Dockerfile
+            platforms: linux/amd64
+          - name: cilium-cli
+            dockerfile: ./Dockerfile-cli-only
```
asauber, line 32: Looks like this should now be `./Dockerfile`

asauber (changes requested):
Thanks for the refactor, looks good to me except for one change.

leppeK:
> Thanks for the refactor, looks good to me except for one change.

Ah yes, I'll change that back

.github/workflows/images.yaml
leppeK: Fixed in last commit, had to rebase from main as well

leppeK:
> Thanks for the refactor, looks good to me except for one change.
@asauber I believe it is in order now, can you sign off the requested changes?

asauber (approved):
LGTM. Thanks for the updates.

viktor-kurchenko (approved):
Thanks!

leppeK:
@michi-covalent Anything blocking this merge?

michi-covalent:
hi @leppeK thanks for the PR 👋 . a couple of things:

- please resolve the conflict in Dockerfile.
- this wouldn't work as is because there is no quay.io/cilium/cilium-cli repository right now, and i'm somewhat hesitant to introduce a new image repository for this purpose. could we use a different tag suffix (maybe something like `-scratch`) instead? that's what we do with https://quay.io/repository/cilium/cilium-ci?tab=tags. we push 2 additional images `-race` and `-unstripped`. the workflow is defined in https://github.com/cilium/cilium/blob/main/.github/workflows/build-images-ci.yaml.

cc @tklauser what do you think? 💭

tklauser (changes requested):
If I understand this change correctly, we would be publishing these cli-only images on every PR and push to `main`. I think we should only publish these for releases. Or do you have a specific use case in mind where you'd need to publish these images on each PR/push?

leppeK:
> If I understand this change correctly, we would be publishing these cli-only images on every PR and push to `main`. I think we should only publish these for releases. Or do you have a specific use case in mind where you'd need to publish these images on each PR/push?

I think it is okay to only do releases for these images so we'd only have stable/latest and release tags.

.github/workflows/images.yaml
```diff
@@ -63,18 +67,20 @@ jobs:
...
           push: true
-          platforms: linux/amd64
+          platforms: ${{ matrix.platforms }}
+          build-args: |
+            FINAL_CONTAINER=${{ matrix.name }}
           tags: |
-            quay.io/${{ github.repository_owner }}/${{ matrix.name }}-ci:latest
-            quay.io/${{ github.repository_owner }}/${{ matrix.name }}-ci:${{ steps.tag.outputs.tag }}
+            quay.io/${{ github.repository_owner }}/${{ matrix.name }}:latest
+            quay.io/${{ github.repository_owner }}/${{ matrix.name }}:${{ steps.tag.outputs.tag }}
```
joestringer, line 75: Reviewing on behalf of @cilium/security: We keep the CI workflows, builds, environment and secrets entirely separate from the production versions to reduce the risk to supply chain security.

I think it probably makes sense to split this overall effort into two steps:
1. Update the CI workflow to build arm64 builds and make any potential changes to the dockerfiles for development builds. Do not change the target repositories.
2. Create another separate workflow for publishing production builds.

@michi-covalent / @tklauser then the other aspect is how we're validating these workflow changes. I believe that deliberately these changes to the workflows would *not* be tested from the fork, because that would mean that arbitrary contributors could hijack the workflows to do what they want. So we probably need to collaborate a little more closely to validate any changes like this separately. That might look like enabling this on a third party github repo and validating there, or at your discretion merging the commits and keeping an eye on them for any potential breakage in the tree.

joestringer: > this wouldn't work as is because there is no quay.io/cilium/cilium-cli repository right now, and i'm somewhat hesitant to introduce a new image repository for this purpose. could we use a different tag suffix (maybe something like -scratch) instead? that's what we do with https://quay.io/repository/cilium/cilium-ci?tab=tags. we push 2 additional images -race and -unstripped. the workflow is defined in https://github.com/cilium/cilium/blob/main/.github/workflows/build-images-ci.yaml.

@michi-covalent if we only need to publish untrusted CI versions of the workflows, something like this works. However, if we intend to publish trusted production builds that we can validate are definitely based only on input from vetted code merged into the tree, then I think it's essential to keep those in a dedicated different repository. With a different repository we can enforce stricter constraints around how code flows into that repository, access to secrets, maintainer approval of individual builds.

.github/workflows/images.yaml
joestringer, line 1: As for the SBOM question from the community meeting: I think that it's not so important for CI builds (see also https://github.com/cilium/cilium/pull/35116), though having SBOM on CI builds can help us to find out if the SBOM steps in the workflow are working or not before we go to do a release.

For releases for CLI, I think it's valuable and should be relatively straightforward (something like https://github.com/cilium/cilium/blob/475310f713fb25897b1b939668c73db7da359d08/.github/workflows/build-images-releases.yaml#L121-L134), but let's get the repos up and running first and worry about SBOM after.

.github/workflows/images.yaml
leppeK: @joestringer 


> 1. Update the CI workflow to build arm64 builds and make any potential changes to the dockerfiles for development builds. Do not change the target repositories.

https://github.com/cilium/cilium-cli/pull/2770/files#diff-d0a3d6684c78a148cbf0725d5fe8b5aab6431da05b698a82c9e015516f3020baR27-R33 Specifically this is also changed so we do not change the target repository for the cli-ci builds

Also adding the new -ci suffixless version, which could be a seperate PR however the Dockerfile changes are already merged

leppeK: @joestringer 
Specifically your second point, there is already such a specific step in this workflow:
https://github.com/cilium/cilium-cli/blob/8aec2847dc037d0925fcda34ddc087b32b06a0be/.github/workflows/images.yaml#L61-L75

which only builds latest and the tagged release when main/tag is pushed

joestringer: The `${{ secrets.QUAY_CI_TOKEN }}` secrets do not have access to publish to quay.io/cilium/cilium-cli, only quay.io/cilium/cilium-cli-ci. The Cilium supply chain threat model considers these repositories to be in different threat categories due to the way that credentials are used in different organization workflows.

Typically the way we handle this in cilium/cilium is that there is an entirely different set of secrets for official builds vs. CI, then there is a dedicated GitHub environment for the official builds that applies strict limits on how builds can be created (for example, requiring maintainer approval), and finally there are different workflows for the different builds with different triggers.

joestringer (changes requested):
Marking as "request changes" until we've resolved the ongoing thread about supply chain / secrets.

PR #2762, opened 2024-08-16
Remove the repo lock workflow

michi-covalent:
https://github.com/cilium/cilium/pull/34178 got merged. vendor cilium-cli from cilium/cilium repo.

4 commits:

- pick up the latest cilium code from main branch: https://github.com/cilium/cilium/commit/e528f991156301a04be66d69e379e699b803d27d
- update main.go to import the CLI from cilium/cilium repo.
- delete unused code.
- unlock the repo and update CONTRIBUTING.md.

PR #2773, opened 2024-08-20
build: fix CLIVersion ldflag ref

chenrui333:
seeing some build failure on the homebrew side, https://github.com/Homebrew/homebrew-core/pull/181762, and realized the package re-org per https://github.com/cilium/cilium-cli/pull/2762

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ STRIP_DEBUG=-w -s
 ifdef DEBUG
 	STRIP_DEBUG=
 endif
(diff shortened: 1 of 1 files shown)


asauber (commented):
Did the build failure on the Homebrew side catch that debugging symbols had not been stripped?

chenrui333:
> Did the build failure on the Homebrew side catch that debugging symbols had not been stripped?

we stripped the debugging symbols for the release build

asauber (approved):
Confirmed locally that this change is needed:

```
cilium-cli $ build/cilium version
cilium-cli: v0.16.16-3-g083a070e compiled with go1.23.0 on darwin/arm64
```

Without this change:
```
cilium-cli $ build/cilium version
cilium-cli:  compiled with go1.23.0 on darwin/arm64
```

PR #2783, opened 2024-09-03
make: set set latest stable version based on stable.txt

tklauser:
Follow-up change for https://github.com/cilium/cilium/pull/34666. We need to set the latest stable Cilium version based on stable.txt from the main Cilium repo as it isn't vendored.

> curl it, man.
> 
>   -- @michi-covalent


michi-covalent (approved):
🥌

PR #2797, opened 2024-09-17
action: Run version command with --client flag

michi-covalent:
Specify --client flag so that the command succeeds even if .kube/config does not exist.

Ref: https://github.com/cilium/cilium/pull/34914


michi-covalent:
✅ tested the change with aks workflow: https://github.com/cilium/cilium-cli/actions/runs/10895501771/job/30234132660

asauber (approved):

PR #2817, opened 2024-09-30
Prepare for #2755 - Dockerfile only

leppeK:
Only the Dockerfile for now as discussed with @michi-covalent because it will diverge everytime

taken from the original commit:

--
This way we are not diverging in multiple dockerfiles and we can choose between two targets:
- cilium-cli-ci The original target based on ubuntu and the default
- cilium-cli Only contains the cilium cli based on scratch 
- --

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,9 +16,11 @@ RUN make
 
 FROM scratch AS cilium-cli
 COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
+ENTRYPOINT ["cilium"]
 
 FROM ubuntu:24.04@sha256:dfc10878be8d8fc9c61cbff33166cb1d1fe44391539243703c72766894fa834a AS cilium-cli-ci
 COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
+ENTRYPOINT []
 
 # Install cloud CLIs. Based on these instructions:
 # - https://cloud.google.com/sdk/docs/install#deb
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
✅

Dockerfile
```diff
@@ -3,32 +3,41 @@
...
 RUN make
 
-FROM ubuntu:24.04@sha256:dfc10878be8d8fc9c61cbff33166cb1d1fe44391539243703c72766894fa834a
-LABEL maintainer="maintainer@cilium.io"
-WORKDIR /root/app
+FROM scratch AS cilium-cli
+COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
+
+FROM ubuntu:24.04@sha256:dfc10878be8d8fc9c61cbff33166cb1d1fe44391539243703c72766894fa834a AS cilium-cli-ci
 COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
```
michi-covalent, line 21: i guess we need to move this line to the final container? 💭

leppeK: In either scenario we get a resulting image layer, one built on top of ubuntu + tools + cilium
the other from scratch with just cilium

leppeK: That is because the ci pipelines define cilium as an argument instead of as the entrypoint
so i made it configurable for now so we can build with a correct entrypoint for the scratch container

michi-covalent:
regarding publishing cilium-cli release images, let's get some input from @cilium/security team to find out whether we need to sign images, generate SBOM, etc.. if you can join weekly community meeting it might be quicker to get feedback from them: https://docs.cilium.io/en/latest/community/community/#weekly-community-meeting otherwise please open a slack thread in #development channel 🚀🙏

PR #2782, opened 2024-09-02
Optimize image layers

marcofranssen:
By moving the parts least likely to change first there will be less
layers that have to be rebuild/pushed/pulled.

PR #2842, opened 2024-10-29
Improve cacheability of Docker layers and add ca-certs to scratch image

marcofranssen:


.github/workflows/images.yaml
```diff
@@ -25,8 +25,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: cilium-cli
+          - name: cilium-cli-ci
```
marcofranssen, line 28: This way we can leverage the `target` in docker build. as well don't need to append `-ci` in various places.

Dockerfile
```diff
@@ -3,24 +3,24 @@
 # Copyright Authors of Cilium
 # SPDX-License-Identifier: Apache-2.0
 
-# FINAL_CONTAINER specifies the source for the output
-# cilium-cli-ci (default) is based on ubuntu with cloud CLIs
-# cilium-cli is from scratch only including cilium binaries
-ARG FINAL_CONTAINER="cilium-cli-ci"
```
marcofranssen, line 9: Replaced by the docker native feature

See https://github.com/cilium/cilium-cli/pull/2842/files#diff-d0a3d6684c78a148cbf0725d5fe8b5aab6431da05b698a82c9e015516f3020baR90

Dockerfile
```diff
@@ -3,24 +3,24 @@
...
 WORKDIR /go/src/github.com/cilium/cilium-cli
 RUN apk add --no-cache curl git make ca-certificates
 COPY . .
 RUN make
 
+# cilium-cli is from scratch only including cilium binaries
 FROM scratch AS cilium-cli
-COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
 ENTRYPOINT ["cilium"]
+LABEL maintainer="maintainer@cilium.io"
```
marcofranssen, line 15: Doing the label before copying the binaries improves cacheability of the layer as it will only require the last 2 layers to be replaced on the final image.

.github/workflows/images.yaml
```diff
@@ -25,8 +25,10 @@ jobs:
...
             dockerfile: ./Dockerfile
+            platforms: linux/amd64
```
marcofranssen, line 30: Makes it more transparant and prepares for #2782

.github/workflows/images.yaml
```diff
@@ -84,17 +87,18 @@ jobs:
         with:
           context: .
           file: ${{ matrix.dockerfile }}
+          target: ${{ matrix.name }}
+          platforms: ${{ matrix.platforms }}
```
marcofranssen, line 91: See here how we now use the matrix variable.

Dockerfile
```diff
@@ -38,7 +40,3 @@ RUN apt-get update -y \
   && ./aws/install \
   && rm -r ./aws awscliv2.zip \
   && curl -sL https://aka.ms/InstallAzureCLIDeb | bash
-
-FROM ${FINAL_CONTAINER} 
-LABEL maintainer="maintainer@cilium.io"
-WORKDIR /root/app
```
marcofranssen, line 44: Moving this into the 2 target images cilium-cli and cilium-cli-ci makes it possible to cache these layers more efficiently meaning those layers will never have to be redownloaded as it is only the layers with binaries at the end that change.

Dockerfile
```diff
@@ -3,24 +3,26 @@
...
 COPY . .
 RUN make
 
+# cilium-cli is from scratch only including cilium binaries
 FROM scratch AS cilium-cli
-COPY --from=builder /go/src/github.com/cilium/cilium-cli/cilium /usr/local/bin/cilium
 ENTRYPOINT ["cilium"]
+LABEL maintainer="maintainer@cilium.io"
+WORKDIR /root/app
+COPY --from=builder --chown=root:root --chmod=755 /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```
marcofranssen, line 17: Ensuring the scratch image also has ca-certificates to prevent issues when calling https endpoints.

Dockerfile
joestringer: nit: This seems like a noop to me. I would be highly surprised if the label/entrypoint commands took any significant amount of resources to execute. But :shrug: I don't see harm in shuffling these around. It's a little more tidy.

joestringer (approved):
LGTM thanks!

Dockerfile
marcofranssen: #2817 IMHO that was a poor mans solution to select a specific target in the Dockerfile, maybe the author wasn't aware of the --target build option.

marcofranssen:
> @marcofranssen Nice work Marco! I This sounds great! It would be nice to add here some benchmarks in terms of before/after build times and resulting image sizes.

Don't really have time to do that. But if you understand docker image layering you will understand there are fewer layers to rebuild as the ones that are most likely to change are the last layers, therefore allowing to reuse the layers before that are most likely not to change.

brlbil (approved):
Thanks LGTM!

michi-covalent (approved):
testing the workflow file change before merging 🚀🙏  https://github.com/cilium/cilium-cli/pull/2842/commits/c29ed79dadd49d5b075eceedd86fdf58fd99c6bb

Dockerfile
michi-covalent: ok let's do it

PR #2844, opened 2024-10-31
action: allow to configure image repo for CI installs

rgo3:


rgo3:
@gandro Was anything else necessary to be able to install the CLI from a third-party image repo?

gandro (approved):
I think this should be sufficient

tklauser (approved):

PR #2850, opened 2024-11-06
.github: use netcat to check hubble port-forward success

devodev:
Use more robust method for checking if hubble port-forward is successful using netcat.

Do note that the workflows use `on.pull_request_target` therefore the changes will not be reflected in this PR checks, but we've tested these on the isovalent CI with success.

CC @michi-covalent 

Will fix upcoming CI failure caused by: https://github.com/cilium/cilium/pull/35483

### DIFF
--- a/.github/workflows/eks-tunnel.yaml
+++ b/.github/workflows/eks-tunnel.yaml
@@ -158,7 +158,7 @@ jobs:
           # Port forward Relay
           cilium hubble port-forward&
           sleep 10s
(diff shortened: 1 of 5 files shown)


michi-covalent (approved):

aanm (approved):

christarazi (approved):

devodev:
I think the AKS test failure is a flaky one, if someone with appropriate permissions could re-run it, that would be great.

Thank you!

PR #2884, opened 2024-12-09
Makefile: Remove CILIUM_VERSION variable

michi-covalent:
Now Cilium CLI sets the default Cilium version without relying on the compile-time flag [^1].

[^1]: https://github.com/cilium/cilium/pull/36344


tklauser (approved):
🧹

PR #2893, opened 2024-12-17
Dockerfile: drop syntax comment

tklauser:
This is being updated by renovate for not much use. We don't use any fancy or deprecated Dockerfile syntax features so it should be fine to leave this unspecified.


michi-covalent (approved):
ok 👍

PR #2915, opened 2025-01-22
renovate config migration remake

bimmlerd:
This is mostly the same as https://github.com/cilium/cilium-cli/pull/2913 but done by hand and split into understandable commits.

I've also taken the liberty of dropping a few unused things from the config, that's the first couple of commits.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,7 +1,7 @@
 {
   $schema: "https://docs.renovatebot.com/renovate-schema.json",
   extends: [
-    "config:base",
+    "config:recommended",
     ":gitSignOff",
     "helpers:pinGitHubActionDigests",
   ],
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -1,235 +1,159 @@
...
-        "digest",
-        "patch",
-        "pin",
-        "pinDigest"
-      ]
-    },
+  stopUpdatingLabel: 'renovate/stop-updating',
+  packageRules: [
     {
-      "groupName": "all go dependencies main",
```
tklauser, line 58: If possible, we want to keep updating the `github.com/cilium/cilium` dependency in `go.mod` automatically so we probably need to keep this hunk around.

bimmlerd: good point, readded.

tklauser (approved):
Thanks for fixing this up and splitting it up into nicely documented individual commits!

PR #2951, opened 2025-02-25
Always update github.com/cilium/cilium from main

tklauser:
Currently, renovate will try to update to the latest patch release, i.e. 1.17.1 at the moment but the stable branches on github.com/cilium/cilium don't contain the cilium-cli/ directory, thus breaking the build on update PRs, see e.g. https://github.com/cilium/cilium-cli/pull/2948

Instead, we should always be vendoring the latest version from main. Drop the renovate configuration accordingly and instead update the github.com/cilium/cilium dependency upon release of a new cilium-cli version.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -58,32 +58,6 @@
       ],
     },
     {
-      groupName: 'all go dependencies main',
-      groupSlug: 'all-go-deps-main',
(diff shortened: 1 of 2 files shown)


qmonnet (approved):
Looks sensible

PR #2965, opened 2025-03-05
gh: add option to specify kubeconfig

marseel:
Example use-case: https://github.com/cilium/cilium/pull/37970


michi-covalent (approved):
😇

PR #2980, opened 2025-03-20
Simplify Dockerfile and clean up docker image workflow

michi-covalent:
- Make the same change as https://github.com/cilium/cilium/pull/37970 to remove cloud provider CLIs. This way we can use the same Dockerfile for CI and releases.
- Get rid of the step to install kubectl. Let's just use the one that comes in the actions runner.
- Clean up images.yaml workflow a bit. Remove the steps for generating image digest files since we don't currently use them.
- Handle the release environment so that images get pushed to the release repo quay.io/cilium/cilium-cli on tag pushes.

### DIFF
--- a/.github/workflows/aks-byocni.yaml
+++ b/.github/workflows/aks-byocni.yaml
@@ -40,7 +40,6 @@ env:
   cost_reduction: --node-vm-size Standard_B2s --node-osdisk-size 30
   # renovate: datasource=github-releases depName=cilium/cilium
   cilium_version: v1.16.8
-  kubectl_version: v1.23.6
 
 jobs:
   installation-and-connectivity:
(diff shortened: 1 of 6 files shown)


michi-covalent:
☑️  todo: remove the temporary commit before merging

.github/workflows/aks-byocni.yaml
```diff
@@ -14,7 +14,7 @@ on:
   #   will disappear from the PR checks: please provide a direct link to the
   #   successful workflow run (can be found from Actions tab) in a comment.
   #
-  # pull_request: {}
+  pull_request: {}
```
viktor-kurchenko, line 17: I think if you want to test `pull_request` then it's better to comment out `pull_request_target`.
Otherwise the result might be unpredictable.

michi-covalent: looks like commenting out `pull_request_target` made it worse, now it's running with `pull_request_target` trigger 😞 

anyways please review the pr, all the workflows were passing with `pull_request` target before 🚀 🙏

viktor-kurchenko (approved):
LGTM

PR #2983, opened 2025-03-21
github: Remove multicluster.yaml

michi-covalent:
Remove GKE-based multi cluster workflow. We already test clustermesh commands in Kind workflow [^1].

[^1]: https://github.com/cilium/cilium-cli/blob/main/.github/workflows/kind.yaml


michi-covalent:
cc @cilium/sig-clustermesh just an fyi

michi-covalent:
... i kind of want to get rid of all the workflows that run cilium connectivity test. it doesn't really make sense to run these in this repo now that the code is in cilium/cilium repo, and this repo is only used to build and publish releases. @tklauser what do you think? 💭

giorio94 (approved):
Sounds reasonable :broom:

viktor-kurchenko (approved):

tklauser:
> ... i kind of want to get rid of all the workflows that run cilium connectivity test. it doesn't really make sense to run these in this repo now that the code is in cilium/cilium repo, and this repo is only used to build and publish releases. @tklauser what do you think? 💭

Yeah, sounds good. Maybe keep one kind-based workflow as a sanity check before release (i.e. only triggered on tag push or something).

tklauser (approved):

michi-covalent:
> Yeah, sounds good. Maybe keep one kind-based workflow as a sanity check before release (i.e. only triggered on tag push or something).

let's start with this to simplify dockerfile https://github.com/cilium/cilium-cli/pull/2980 we can sleep on this a bit before deciding which workflows to keep 😴

PR #3026, opened 2025-05-06
Dockerfile: do not specify syntax

tklauser:
We no longer use the syntax directive in Cilium upstream Dockerfile, they were removed by commit cilium/cilium@005f900c4060 ("docker: Do not specify syntax"). Drop it in the cilium-cli Dockerfile as well and avoid noisy renovatebot PRs updating the synatx version.


mhofstetter (approved):
Thanks!

PR #3174, opened 2026-01-28
ci: Harden the image build process

ferozsalam:
- Add a comment to clearly indicate the untrusted code checkout
- Split the release and dev CI steps into two separate workflows
- Explicitly specify the permissions of the GitHub token in the workflows, to minimize the impact of token compromise.
- Split the individual workflows into separate jobs, such that the build happens in a different job to the push, which will allow the separation of the untrusted code build process from the authenticated push of the built artifact.

Thanks to @cedricvanrompay for reporting this.


.github/workflows/images-release.yaml
```diff
@@ -0,0 +1,89 @@
...
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.after }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    if: ${{ github.repository == 'cilium/cilium-cli' }}
+    runs-on: ubuntu-24.04
```
joestringer: This workflow omits `permissions`, is that intentional?
```suggestion
    runs-on: ubuntu-24.04
    permissions:
      contents: read
```

joestringer (approved):
Minor nit but LGTM. I reviewed mainly just by doing `diff -u` between the original workflow and each new workflow, and between each new workflow here to identify any discrepancies.

tklauser (approved):
> Test of the CI build here, used `pull_request` as the trigger to test: https://github.com/cilium/cilium-cli/actions/runs/21437055364/job/61730258277?pr=3174
> 
> Not certain of the best way to test the equivalent release build.

So far we've been testing changes to release workflows only when cutting a release anyway, making adjustments as we hit issues. This approach is not optimal but I can't think of a good way to test this without too much complication. So IMO it is fine to merge this untested and follow-up with fixes in case we hit issues on the next release.

PR #3233, opened 2026-05-13
ci: add write permissions to release workflow

tklauser:
This allows the release workflow to create releases. Otherwise it will fail with:

    Creating new GitHub release for tag v0.19.3...
    GitHub release failed with status: 403
    {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/releases/releases#generate-release-notes-content-for-a-release","status":"403"}

as seen on https://github.com/cilium/cilium-cli/actions/runs/25797138861/job/75777197322

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,11 +5,14 @@ on:
     tags:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
 
+permissions:
+  contents: read
+
 name: Release
 
 jobs:
   build:
-    name: Create Release ${{ github.event_name == 'pull_request' && '(dry-run)' }}
+    name: Build Release Artifacts
     if: github.repository == 'cilium/cilium-cli'
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):

.github/workflows/release.yaml
```diff
@@ -5,6 +5,9 @@ on:
     tags:
       - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
 
+permissions:
+  contents: write
```
joestringer: This workflow runs `make release`, so we'll run whatever logic a contributor puts into the Makefile directly on the runner with `contents:write` permission. Do we want that?

I'd feel slightly more confident if these permissions were scoped in jobs and the write privilege was only granted for the release creation step, perhaps by doing the build + artifact push in one job (which could push arbitrary unverified content) then separately pull and create the release only for `push` trigger from a separate job.

Why do we have the `pull_request: {}` trigger at all for this by the way? Another option might just be to drop this.

joestringer: I guess `pull_request: {}` is for "dry-run", probably as a CI validation check for the underlying mechanisms. But for that probably no permissions are needed.

tklauser: Thanks @joestringer! Indeed, granting `contents: write` permissions to the overall workflow is not what we want in this case. I adjusted the PR as suggested to split the workflow into to jobs. We npw still run the build step (i.e. `make release`) with `contents: read` permissions and only grant `contents: write` permissions to the release job.

> I guess pull_request: {} is for "dry-run", probably as a CI validation check for the underlying mechanisms. But for that probably no permissions are needed.

Exactly, we want to ensure that the release artifacts remain buildable on every PR rather than only finding out when we cut a release.

sekhar-isovalent (approved):

